PR #55, opened 2015-10-28
Added checking whether some obm setting is valid for node

iceiilin:
Check whether current node support a specific obm setting.
Also add retries in Panduit obm settings to cover vPDU's slow response.

https://hwjiraprd01.corp.emc.com/browse/MAG-222

related PRs
https://github.com/RackHD/on-http/pull/28
https://github.com/RackHD/on-core/pull/22

### DIFF
--- a/lib/services/panduit-obm-service.js
+++ b/lib/services/panduit-obm-service.js
@@ -68,14 +68,21 @@ function panduitObmServiceFactory(
                                      '.' + outletNum,
                                      's', cyclePwd],
                                      host,
-                                     community)
+                                     community,
+                                     'snmpset'
+                                    )
                 .then(function () {
+                    // vPDU doesn't accept control command right after setting password
+                    // Retry three times to ensure the controlling succeed
                     return self._runInternal([self.mibPowerCommand + 
                                              '.' + pduNum + 
                                              '.' + outletNum,
                                              'i', controlCmd[control]],
                                              host,
-                                             community);
+                                             community,
+                                             'snmpset',
+                                             3
+                                            );
                 });
         });
     };
@@ -143,14 +150,15 @@ function panduitObmServiceFactory(
         });
     };
 
-    PanduitObmService.prototype._runInternal = function (command, host, community, file) {
+    PanduitObmService.prototype._runInternal = function (command, host, community, file, retries) {
         return this.run({
-            command: file || 'snmpset',
+            command: file,
             args: [
                 '-v2c',
                 '-c', community,
                 host
-            ].concat(command)
+            ].concat(command),
+            retries: retries
         });
     };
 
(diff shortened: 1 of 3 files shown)


iceiilin:
@benbp

benbp:
Hey saw the updates. Sorry for the delay I'll be able to review this tomorrow.

spec/lib/services/obm-service-spec.js
```diff
@@ -89,4 +90,147 @@ describe('OBM Service', function() {
             });
         });
     });
+
+    describe('Check invalid obm service', function() {
+        var waterline;
+
+        before('check invalid obm service before', function() {
+            waterline = helper.injector.get('Services.Waterline');
+            waterline.skus.findOne = sinon.stub();
```
benbp: Does this not fail without doing `sinon.stub().resolves` ? I would expect calling `.then` on it to not work otherwise.

benbp:
:+1:

lib/services/panduit-obm-service.js
```diff
@@ -143,14 +150,15 @@ function panduitObmServiceFactory(
         });
     };
 
-    PanduitObmService.prototype._runInternal = function (command, host, community, file) {
+    PanduitObmService.prototype._runInternal = function (command, host, community, file, retries) {
         return this.run({
-            command: file || 'snmpset',
```
jlongever: how come the default 'snmpset' was removed?

lib/services/panduit-obm-service.js
```diff
@@ -143,14 +150,15 @@ function panduitObmServiceFactory(
...
+            command: file,
             args: [
                 '-v2c',
                 '-c', community,
                 host
-            ].concat(command)
+            ].concat(command),
+            retries: retries
```
jlongever: should there be a default value when retries is not set?

spec/lib/services/obm-service-spec.js
```diff
@@ -89,4 +90,147 @@ describe('OBM Service', function() {
...
+            .then(function() {
+                done(new Error('Expected invalid service'));
+            })
+            .catch(function(e) {
+                try {
+                    expect(e).to.have.property('name').that.equals('BadRequestError');
+                    expect(e).to.have.property('message').that.equals(
+                        'Service panduit-obm-service is not supported in current node'
+                    );
+                    done();
```
yyscamper: if the promise is returned, we don't need to call done()

iceiilin: done() is used to specify the case has finished.

lib/services/panduit-obm-service.js
iceiilin: this.run() command in base-obm-service always expect "retries" as its parameter. If it is not specified, run() in child-process.js of on-core has assigned a 0 as its default value.

lib/services/panduit-obm-service.js
iceiilin: 'snmpset' in "file" parameter was the last parameter and it can be set as a default one. Now I add "retires", so I don't view "file" as a default value, and added "snmpset" in this file when calling this _runInternal function. It works when this function is called in only one or two places. Thank you for reminding me that there are some other ways to add a new parameter with less code changes and confusion.

anhou:
:+1:

yyscamper:
:+1:

PR #58, opened 2015-10-30
made switch catalogs use numeric output 

VulpesArtificem:


benbp:
@VulpesArtificem was the resolution that we don't need this?

VulpesArtificem:
yes; unnecessary.

PR #60, opened 2015-11-04
VEY-204: Create local catalog job

jlongever:
Command based job that can run on the local management system.
Added local lldp catalog
Added local dmidecode catalog
Fixed local bmc catalog to include new "|| true" parsing format

Dependent PR: https://github.com/RackHD/on-taskgraph/pull/20

@RackHD/corecommitters @RackHD/rackhd_dev


jlongever:
@jfrey @benbp  Is that travis CI failure a legit failure?

zyoung51:
Great stuff :+1:

benbp:
@jlongever no the travis failures aren't real yet.

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,176 @@
...
+                .then(function(ret) {
+                    cmdResp.push(ret);
+                });
+            });
+        })
+        .then(function() {
+            return cmdResp;
+        })
+        .then(self.handleResponse.bind(self))
+        .then(function() {
```
benbp: Since you're doing bind trickery a line above, you can do it here too if you want:

```
.then(self._done.bind(self))
.catch(self._done.bind(self))
```

benbp: Also, fun fact, when we transition to node 4, we can use ES6 fat arrows to avoid having to call bind:

```
.then(() => self._done())
.catch(e => self._done(e))
```

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,176 @@
...
+                                   logger,
+                                   options, 
+                                   context, 
+                                   taskId);
+
+        assert.string(this.context.target);
+        assert.arrayOfString(this.options.commands);
+
+        this.nodeId = this.context.target;
+        this.commands = options.commands;
```
benbp: This line is unnecessary?

jlongever: Yep, i'll remove it.

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,176 @@
...
+        this.commands = _.isArray(options.commands) ? options.commands : [options.commands];
+        this.acceptedResponseCodes = options.acceptedResponseCodes[0] || 1;
```
benbp: If acceptedResponseCodes is undefined, this will throw an error.

lib/jobs/local-catalog.js
jlongever: Got unit test failure: TypeError: Cannot call method 'bind' of undefined

Will leave it as-is for now.

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,176 @@
...
+     */
+    LocalCatalogJob.prototype._run = function() {
+        var self = this;
+        var cmdResp = [];
+
+        waterline.nodes.findByIdentifier(self.nodeId)
+        .then(function (node) {
+            assert.ok(node, 'No node for local catalog');
+            
+            return Promise.each(self.commands, function(cmd) {
```
benbp: You can avoid using an outside variable here with Promise.map:

```javascript
var Promise = require('bluebird');
var commands = [1,2,3];

Promise.resolve()
.then(function() {
    return Promise.map(commands, function(cmd) {
        return Promise.resolve(cmd + 1);
    });
})
.then(function(args) {
    console.log(args);
});
```

```bash
$ node test.js
[ 2, 3, 4 ]
```

lib/jobs/local-catalog.js
jlongever: this can probably be removed,  only acceptable response should be 0.

lib/jobs/local-catalog.js
benbp: Huh, that's weird. Ok.

benbp:
So, my main worry here is this enables a pretty easy way for anyone with API access to run arbitrary privilege-escalated commands on the server. This is why with the local IPMI catalog job we coded only arguments to the support binary ipmitool. I think it might make sense here as well, instead of supporting any command, only support dmi, lldp, etc. and specify that as options.command, then the options.commands becomes options.arguments

jlongever:
@benbp If someone obtains API access it's probably already too late? ;-) 
Originally I made this command specific but wanted to avoid code duplication so I implemented it as a common local command job.  How about checking for the task parser match upfront prior to running the command?

benbp:
Ah ok, so I whitelist of supported commands? That will work.

I think if someone obtains API access it may be too late currently, but in the future may not indicate that, so don't want to set ourselves up with any backdoors that will be forgotten :smiley_cat:

jlongever:
Sounds good.  Will make the change.

spec/lib/jobs/local-catalog-spec.js
```diff
@@ -0,0 +1,179 @@
...
+
+        beforeEach('Local catalog job input validation', function(){
+            var options = {
+                commands: ['test_command'],
+                acceptedResponseCodes: [1]
+            };
+            job = new LocalCatalogJob(options, { target: 'bc7dab7e8fb7d6abf8e7d6ac' }, uuid.v4());
+            mockWaterline.nodes.findByIdentifier = function(){};
+
+            this.sandbox = sinon.sandbox.create();
```
anhou: Just as @benbp  mentioned in another PR "This was an error that I put in a bunch of tests a while back. This statement can actually go in the before block instead of the beforeEach block, since we only need to create the sandbox once (and will save us test execution time)."

jlongever: The helper.before() should probably be used across unit-tests, however that introduces rabbitmq dependencies.

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,175 @@
...
+        });
+    };
+
+    /**
+     * @memberOf LocalCatalogJob
+     */
+    LocalCatalogJob.prototype.handleResponse = function(result) {
+        var self = this;
+
+        return parser.parseTasks(result)
```
yyscamper: Just think of another not so good design: the command string need be exactly same to leverage previous catalog parser, "cmd -f -b" and "cmd -b -f" should generate same result, but now we cannot map it to same catalog parser.

lib/jobs/local-catalog.js
jlongever: @benbp actually,  it's the unit test not returning these correctly.

lib/jobs/local-catalog.js
jlongever: IMO, the command string should be explicitly matched,  comparing sets could also potentially pass if "-f -b cmd" were used.  It also protects commands that have a required argument order (success when "cmd -f -b" but errors or behaves differently when "cmd -b -f".

FWIW,  this doesn't change the original parser design;  see lib/utils/job-utils/command-parser.js.  It has always matched command strings.

lib/jobs/local-catalog.js
jlongever: Tricky tricky.  Thanks.

lib/jobs/local-catalog.js
benbp: We also support regexes to match against commands, in cases where we don't want to/can't be explicit (like ipmi user N).

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,180 @@
...
+                return parser.validateParser(cmd)
+                .then(function() {
+                    return self.runCommand(cmd);
+                })
+                .catch(function(err) {
+                    logger.error("Run error", { command: cmd, error: err });
+                });
+            });
+        })
+        .then(function(cmdResp) {
```
benbp: You can just get rid of this promise block now since you updated to .map.

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,180 @@
...
+        waterline.nodes.findByIdentifier(self.nodeId)
+        .then(function (node) {
+            assert.ok(node, 'No node for local catalog');
+            return Promise.map(self.commands, function(cmd) {
+                return parser.validateParser(cmd)
+                .then(function() {
+                    return self.runCommand(cmd);
+                })
+                .catch(function(err) {
+                    logger.error("Run error", { command: cmd, error: err });
```
benbp: Do we want to throw here and fail the job if a command isn't in the whitelist, rather than just dropping it?

lib/jobs/local-catalog.js
```diff
@@ -0,0 +1,180 @@
...
+            program = args[1];
+            args.shift();
+        }
+        args.shift();
+        return Promise.resolve()
+        .then(function() {
+            var env = {
+                'nodeid': self.nodeId
+            };
+            var childProcess = new ChildProcess(program, args, env);
```
benbp: One thing that would be good to support in this job is adding a _cleanup() function to the job that kills this ChildProcess, that way if someone cancels the workflow we don't have a bunch of background child processes still running. It's probably not critical since the catalog commands shouldn't change state, but it's still nice to know we're not leaking processes for any amount of time.

benbp: I think @zyoung51 has an example of this in his ansible job.

lib/utils/job-utils/command-parser.js
```diff
@@ -1028,6 +1028,18 @@ function commandParserFactory(Logger, Promise, _) {
...
+    
+    CommandParser.prototype.validateParser = function validateParser(cmd) {
+        var self = this;
+        if (self[cmd] !== undefined) {
+            return Promise.resolve(cmd);
+        }
+        var error = new Error("Command parser does not exist");
+        return Promise.reject({
+            command: cmd, 
+            error: error
```
benbp: Does this work? I thought reject took an error argument.

lib/jobs/local-catalog.js
jlongever: Throwing here would prevent the next command in the list to run?

lib/utils/job-utils/command-parser.js
jlongever: This is the handled catch on an invalid command:

```E 2015-11-05T19:13:07.005Z [on-taskgraph] [Job.Local.Catalog] [Server] Run error
 -> /node_modules/on-tasks/lib/jobs/local-catalog.js:75
command: sudo /usr/sbin/lldpcliXXX show neighbor -f keyvalue
error:
  command: sudo /usr/sbin/lldpcliXXX show neighbor -f keyvalue
  error:
    stack:
      - Error: Command parser does not exist
      -     at CommandParser.validateParser (/var/renasar/on-taskgraph/node_modules/on-tasks/lib/utils/job-utils/command-parser.js:1037:21)
      -     at /var/renasar/on-taskgraph/node_modules/on-tasks/lib/jobs/local-catalog.js:70:31
      -     at tryCatcher (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/bluebird/js/main/util.js:26:23)
      -     at MappingPromiseArray._promiseFulfilled (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/bluebird/js/main/map.js:56:38)
      -     at MappingPromiseArray.init [as _init$] (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/bluebird/js/main/promise_array.js:92:18)
      -     at MappingPromiseArray.init (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/bluebird/js/main/map.js:29:23)
      -     at Async._drainQueue (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/bluebird/js/main/async.js:128:12)
      -     at Async._drainQueues (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/bluebird/js/main/async.js:133:10)
      -     at Async.drainQueues (/var/renasar/on-taskgraph/node_modules/on-core/node_modules/bluebird/js/main/async.js:15:14)
      -     at process._tickDomainCallback (node.js:459:13)
    message: Command parser does not exist```

Comment (jlongever) [inline lib/utils/job-utils/command-parser.js]: Is that the same as return Promise.reject(new Error("message"));  ?

Comment (benbp) [inline lib/utils/job-utils/command-parser.js]: Huh, guess it does work! The command info seems to be redundant anyways though, since it's right there in Error message?

File: lib/jobs/local-catalog.js (hunk 8 above)
Comment (benbp) [inline lib/jobs/local-catalog.js]: Yeah, I guess that's the question I'm asking: do we want to just run whatever we can, or only succeed if we can run everything specified.

File: lib/utils/job-utils/command-parser.js (hunk 10 above)
Comment (jlongever) [inline lib/utils/job-utils/command-parser.js]: Actually it worked but not the way I thought, it's nested.   It should just be an error object.

File: lib/jobs/local-catalog.js (hunk 8 above)
Comment (jlongever) [inline lib/jobs/local-catalog.js]: IMO it should go to the next command;  as long as the failed command is logged.  But I'm not 100% sure.

**Comment by jlongever** [comment]
@benbp

File: lib/jobs/local-catalog.js (hunk 8 above)
Comment (benbp) [inline lib/jobs/local-catalog.js]: Sounds good, yeah I'm not sure either.

**Comment by benbp** [comment]
:+1:

**Comment by benbp** [comment]
Let's wait until https://github.com/RackHD/on-tasks/pull/61 is merged so we can take advantage of the travis tests to start the trend? @heckj is fixing that build, so it should be merged shortly.

**Comment by benbp** [comment]
@jlongever will you rebase this against the current master and push again? That way we can kick off a Travis build (yay).

**Comment by jlongever** [comment]
@benbp will do.

File: lib/jobs/local-catalog.js [hunk 11]
```diff
@@ -0,0 +1,189 @@
...
+                'nodeid': self.nodeId
+            };
+            self.childProcess = new ChildProcess(program, args, env);
+
+            logger.debug("Sending command to node.", {
+                command: cmd,
+                nodeId: self.nodeId
+            });
+
+            return self.childProcess.run({ retries: 0, delay: 0 });
```
iceiilin: Can we add a timeout value for executing child process in case that the thread hangs and blocks this job?

jlongever: what timeout value would you suggest?

jlongever: @iceiilin  execFile() is encapsulated in on-core/lib/common/child-process.js and doesn't set the timeout option.   We can revisit this in another PR when timeout is supported there.

iceiilin: It partially depends on the actual time of executing commands. Besides of this, I think several minutes might help from the aspect of user experience.

PR #62, opened 2015-11-05
merge ESXi bootstrap and add analyze os repo job

yyscamper:
This PR is to sync up the bootstrap implementation of 1.0.0 branch, it can be thought of the combination of following PRs :
 - https://github.com/RackHD/on-tasks/pull/47   (1.0.0 branch)
 - https://github.com/RackHD/on-tasks/pull/45   (1.0.0 branch)
 - https://github.com/RackHD/on-tasks/pull/43   (master, but closed)

@benbp  had some comment in this PR (https://github.com/RackHD/on-tasks/pull/45) about the code architecture design, this PR also contains the proposal code arch.

Now the separated job "analyze-os-repo" is designed to dynamic fetch some options from user's repository, this can be benefit in following area:
(1) The ESXi 5.5 and ESXi 6.0 workflow can be merged into one.
(2) Can be used as a pre-validation for the external repository, if repository is not correct, the bootstrap will skip early and save 5-10 min. 

@RackHD/corecommitters @zyoung51 @iceiilin @WangWinson @pengz1 

Following PRs need be merged together to ensure the bootstrap function works well for master branch:
1. https://github.com/RackHD/on-http/pull/38
2. https://github.com/RackHD/on-taskgraph/pull/22
3. https://github.com/RackHD/on-tasks/pull/62
4. https://github.com/RackHD/onserve/pull/64

### DIFF
--- a/lib/jobs/analyze-os-repo-job.js
+++ b/lib/jobs/analyze-os-repo-job.js
@@ -1,0 +1,151 @@
+// Copyright 2015, EMC, Inc.
+/* jshint: node:true */
+
+'use strict';
+
+var di = require('di');
+
+module.exports = analyzeOsRepoJobFactory;
+di.annotate(analyzeOsRepoJobFactory, new di.Provide('Job.Os.Analyze.Repo'));
+    di.annotate(analyzeOsRepoJobFactory,
+    new di.Inject(
+        'Job.Base',
+        'Logger',
+        'Assert',
+        'Util',
+        '_',
+        'Promise',
+        'JobUtils.OsRepoTool'
+    )
+);
+
+function analyzeOsRepoJobFactory(
+    BaseJob,
+    Logger,
+    assert,
+    util,
+    _,
+    Promise,
+    repoTool
+) {
+    var logger = Logger.initialize(analyzeOsRepoJobFactory);
+
+    /**
+     * This job will analyze the external OS repository, fetch some options from some repository
+     * files and pass these options to shared context:
+     * context.repoOptions  = { xxx: xxx }
+     *
+     * @param {Object} options
+     * @param {Object} context
+     * @param {String} taskId
+     * @constructor
+     */
+    function AnalyzeOsRepoJob(options, context, taskId) {
+        var self = this;
+        AnalyzeOsRepoJob.super_.call(self, logger, options, context, taskId);
+
+        self.nodeId = self.context.target;
+
+        assert.string(self.options.osName);
+        assert.string(self.options.repo);
+
+        // Both http://xxx/repo and http://xxx/repo/ should be valid and point to same repository,
+        // but our code prefer the previous one
+        if (self.options.repo) {
+            self.options.repo =  self.options.repo.trim();
+            if (_.last(self.options.repo) === '/') {
+                self.options.repo =  self.options.repo.substring(0, self.options.repo.length-1);
+            }
+        }
+    }
+
+    util.inherits(AnalyzeOsRepoJob, BaseJob);
+
+    /**
+     * @memberOf AnalyzeOsRepoJob
+     */
+    AnalyzeOsRepoJob.prototype._run = function() {
+        var self = this;
+
+        return Promise.resolve().then(function() {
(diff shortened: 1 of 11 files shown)


yyscamper:
test this please

package.json
```diff
@@ -41,6 +41,7 @@
     "sinon-as-promised": "^2.0.3",
     "sinon-chai": "^2.7.0",
     "supertest": "^0.15.0",
-    "xunit-file": "0.0.6"
+    "xunit-file": "0.0.6",
+    "nock": "^2.17.0"
   }
```
yyscamper: nock (https://github.com/pgte/nock) is a famous HTTP mocking and expectations library, I plan to add it to the spec/helper.js, but by different PR, how do you think? @RackHD/corecommitters

lib/jobs/analyze-os-repo-job.js
```diff
@@ -0,0 +1,151 @@
...
+     * @memberof AnalyzeOsRepoJob
+     * @param {String} repo - the external repository address.
+     * @return {Promise}
+     */
+    AnalyzeOsRepoJob.prototype._esxHandle = function (repo) {
+        //first try the lower case because the installation has some problem when the repository
+        //is in upper case, but anyway we will try the upper case as a retry, in future we (maybe
+        //Vmware?) may have solution to fix the upper case problem.
+        return repoTool.downloadViaHttp(repo + '/boot.cfg').catch(function() {
+            return repoTool.downloadViaHttp(repo + '/BOOT.CFG');
```
benbp: Wow, that's weird, oh well.

package.json
benbp: Yeah, seems generally useful, though I'm also fine with leaving it in the os-repo-tool-spec unit test since it's not used anywhere else.

benbp:
:+1:

benbp:
So, just to make sure I'm understanding correctly: Both the install-os and analyze-os tasks get the same options, but the analyze-os just validates them first and fails if they are invalid? Otherwise, we proceed to install-os with the same options?

jlongever:
@yyscamper This seems specific to ESXi repositories,  analyze-os sounds general.  Any plans to support other OS repos?

benbp:
Yeah, some re-naming probably makes sense.

yyscamper:
analyze repo is only accept option "repo" and "version", other options such as vlan, domain, root password and so on will still be validated in install os. Now only ESXi uses analyze repo job, I don't want analyze repo to take all validation work.

Analyze repo only pass the new options that from repo and converted repo address to install os, nothing more (at least now).


Felix

yyscamper:
@jlongever @benbp : I did think about the naming. I choose the common one because it want this job can be used to handle all os repo analysis work, though now we only do this for ESXi.

You could see in this line (https://github.com/RackHD/on-tasks/pull/62/files#diff-0968fbbf157945a0d7f47f29f9ab72f3R71), I write a function to dynamic find the handler for each OS, so did think about we will need extra analysis for other OS.

Analyze-os-repo job will accept an options "osName" to load the correct handle, just like we did for obm-control, use the "action" to specify which operation need to done.

Some of works that we may need to do in future for other OS, and suitable for analyze-os-repo job:
(1) Analyze whether the input repo is really existed.
(2) Analyze the integrity of external repo, whether it is valid, any file missing.

package.json
yyscamper: OK, we can move it until we find another repo still need http mock.

iceiilin:
+1 
I like the detailed error log and the code structure here. 
I am thinking about adding analyze-os-repo job to every install OS workflow to make it seems common, though we may passthrough it inside the job itself if it isn't needed actually. Just some minor thoughts, and you can ignore it if you think it doesn't make sense.

zyoung51:
@yyscamper Are there stories created to add analyze routines for other OS (RHEL seems like the next priority here)?
:+1:

yyscamper:
@zyoung51 : currently, there is no need to analyze other OS, though I proposed we can use the analyze-os-repo to do some wors, but it is not urgent now.  Do you have some idea on this?

PR #61, opened 2015-11-05
adding travis.yml

heckj:

PR #63, opened 2015-11-05
Remove helper.before from snmp-pdu-power-status-spec

benbp:
The test passes without this, and helper.before introduces mongo+rabbit dependencies, so remove it.

@VulpesArtificem @heckj


heckj:
lgtm, i'll hold on resetting and retrying https://github.com/RackHD/on-tasks/pull/61 until this is merged if that works for you two

VulpesArtificem:
:+1:

PR #71, opened 2015-11-17
Add drive partition control job and task

yyscamper:
This PR is to implement a job to do drive partition control, such as clear all partitions for target drives.

It is a sub-feature of ORFS-97 [Extend base OS and workflow to configure RAID array prior to bootstrap](https://hardware.corp.emc.com/display/CI/Extend+base+OS+and+workflow+to+configure+RAID+array+prior+to+bootstrap).

This task will be eventually embedded in OS bootstrap workflow, as we found some error when install ESXi and Redhat alternately, if the drive partition is corrupted, the ESXi installation will fail, so we have to manually reset the drive config. So if this task is inserted, we can avoid such manual work.

The drive partition operation will depend on a tool named "parted", @iceiilin opened another PR (https://github.com/RackHD/on-imagebuilder/pull/7) to include this tool in our micorkernel.

The dependent PR: https://github.com/RackHD/on-imagebuilder/pull/7

### DIFF
--- a/lib/jobs/drive-partition-control.js
+++ b/lib/jobs/drive-partition-control.js
@@ -1,0 +1,222 @@
+// Copyright 2015, EMC, Inc.
+/* jshint: node:true */
+
+'use strict';
+
+var di = require('di');
+
+module.exports = drivePartitionJobFactory;
+di.annotate(drivePartitionJobFactory, new di.Provide('Job.Drive.Partition.Control'));
+    di.annotate(drivePartitionJobFactory,
+    new di.Inject(
+        'Job.Base',
+        'Logger',
+        'Promise',
+        'Assert',
+        'Util',
+        '_'
+    )
+);
+
+function drivePartitionJobFactory(
+    BaseJob,
+    Logger,
+    Promise,
+    assert,
+    util,
+    _)
+{
+    var logger = Logger.initialize(drivePartitionJobFactory);
+
+    /**
+     * @param {Object} options
+     * @param {Object} context
+     * @param {String} taskId
+     * @constructor
+     */
+    function DrivePartitionJob(options, context, taskId) {
+        DrivePartitionJob.super_.call(this, logger, options, context, taskId);
+
+        assert.string(this.context.target);
+        assert.string(this.options.action);
+        this.nodeId = this.context.target;
+    }
+    util.inherits(DrivePartitionJob, BaseJob);
+
+    /**
+     * Find the action handle function for specified action
+     * @param {String} actionName - The action
+     * @return {Function} The action handle function
+     */
+    DrivePartitionJob.prototype._findActionHandle = function(actionName) {
+        var funcName = actionName + 'Handle';
+        var handleFunc = this[funcName];
+        if (!handleFunc) {
+            return function() {};
+        }
+        if (!_.isFunction(handleFunc)) {
+            throw(new Error('The handling for partition ' + actionName + ' is not callable.'));
+        }
+        return handleFunc;
+    };
+
+    /**
+     * Build the commands for disk partition control
+     * @return {Array} The array contains the full format command of partition control using
+     * the tool 'parted'
+     */
+    DrivePartitionJob.prototype._buildCommands = function() {
+        var self = this;
+        if (self.options.action === 'clear') {
+            var type = self.options.partitionType || 'gpt';
+            return self.options.driveIds.map(function(id) {
+                return {
+                    cmd: 'sudo parted -s ' + id + ' mklabel ' + type
+                };
+            });
+        }
+    };
+
+    /**
+     * The function the parse the output of tool 'parted', to judge whether the parted does job
+     * successfully or failed.
+     *
+     * For the 'parted' tool, if command execution success, the command output is empty. So if the
+     * command output is emtpy, it means the command runs successfully, otherwise failed.
+     *
+     * @param {String} commandOutput - The stdout of parted command execution.
+     * @return {Object} The returned object contains two properties 'error', 'data'; the 'error'
+     * indicates whether the command execution has failure (by judging the command output), the
+     * 'data' is optional, it contains the parsed or re-formated data.
+     */
+    DrivePartitionJob.prototype._outputParser = function(commandOutput) {
+       var failed = (commandOutput !== '');
+        return {
+            error: failed
+        };
+    };
+
+    /**
+     * The function to handle partition clearing.
+     * @return {Promise}
+     */
+    DrivePartitionJob.prototype.clearHandle = function() {
+        return this.runRemoteCommand(this._buildCommands(), this._outputParser);
+    };
+
+    /**
+     * Handle the response data from node
+     * @param {Array} data - The response data from node for each commands.
+     * @param {Function} parser - The function to parse the each command output.
+     * @return {Object}
+     */
+    DrivePartitionJob.prototype.handleResponse = function(data, parser) {
+        var self = this;
+        var parsedResult = [];
+        var failed = _.some(data.tasks, function(task) {
+            if (task.error) {
+                logger.error('Fail to run command on node.', {
+                    nodeId: self.nodeId,
+                    responseObject: task
+                });
+                return true;
+            }
+            if (parser) {
+                var result = parser.call(null, task.stdout);
+                if (result.error) {
+                    logger.error('The parser tells command execution has failure on node.', {
+                        nodeId: self.nodeId,
+                        responseObject: task
+                    });
+                    return true;
+                }
+                parsedResult.push(result);
+            }
+            else {
+                parsedResult.push(task.stdout);
+            }
+            return false;
+        });
+
+        if (failed) {
+            return { error: true };
+        }
+        else {
+            return { error: false, data: parsedResult };
+        }
+    };
+
+    /**
+     * The common function to remotely run commands in microkernel and handle its response,
+     * This function is wrapped as Promise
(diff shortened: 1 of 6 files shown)


yyscamper:
@zyoung51 : this will eventually a feature of bootstrap.

lib/jobs/drive-partition-control.js
```diff
@@ -0,0 +1,222 @@
...
+        // Called if this job is used to render a script template
+        self._subscribeRequestProperties(function() {
+            return self.options;
+        });
+
+        //if not disks are specified, then do nothing
+        if (!self.options.driveIds || self.options.driveIds.length === 0) {
+            logger.debug('no driveIds is specified, so ignore drive partition operation.');
+            self._done();
+            return Promise.resolve();
```
jlongever: this resolve() isn't needed, _done() will handle it.

jlongever:
Is there a way to get the current partitions and driveId's and catalog them?  Is there still manual steps to get them prior to running this workflow?

lib/jobs/drive-partition-control.js
benbp: I think @yyscamper's intent here is so the _run function can be tested more easily, however I still think it's better to do `return job.deferred.promise` in test because it's more flexible for _run functions.

benbp:
Also, would it be simpler to just utilize the existing linux commands job, so it's a little more transparent that we're just really doing a couple bash one-liners, and we don't have to add a bunch more code? Or do you think this will limit us? I'm generally in favor of re-using the same job and having the differentiation in the task definition, rather than making something like a Generic Partition Control Job.

For example, I think we could do all this with a simple task definition (correct me if I'm wrong):
```
module.exports = {
    friendlyName: "Clear Drives Partition",
    injectableName: "Task.Drive.Clear.Partition",
    implementsTask: "Task.Base.Drive.Partition.Control",
    options: {
        driveIds: "/dev/disk/by-id/wwn-0x6001636001940aa01dd55961a20fb847 "\
                  "/dev/sda " \
                  "/dev/disk/by-id/ata-SATADOM-SV_3SE_20150522AA999091007C "
        partitionType: "gpt",
        command: "for drive in {{ options.driveIds }}; " \
                 "do sudo parted -s $drive mklabel {{ options.partitionType || gpt }}; done"
    },
    properties: {}
};
```

benbp:
@wangwinson was just talking to me about improving task rendering to enumerate template tags/hashes which would make my above example cleaner and remove the need for the for loop as well.

lib/jobs/drive-partition-control.js
yyscamper: Yea, you got it. In  the unit-test, I directly use the promise chain of job._run(). As you said, If change it to job._deferred.promise in unit-test, this resolve() can be removed.

yyscamper:
@jlongever: sure we will have catalog for driveId, @pengz1  is working on it in southbound, @iceiilin is working on northbound API to report driveId to user.

But we don't plan to do drive partition catalog at present. Firstly now we only use it to clear drive partition before bootstrap, if someday we need to create drive partition, then I think we need the drive partition catalog. Secondly, the OS installer will automatically create partitions (now we configure it as auto-part), so the drive partition catalog will be out-of-date. If we force re-catalog after bootstrap, this will increase about 5 min for the overall bootstrap time.

yyscamper:
@benbp : thanks for your suggestion about using Linux Command job, I didn't realize this design.

For command Linux command job, the short-coming that I can think of right now is it cannot be used to do some customized input pre-checking before command execution. May we can improve it by following the command parser, implement a centralized pre-checking util functions pool. 

Anyway, the pre-checking is not so important for clearing partition control (at least now), so I will change it as your suggestion, it saves so many codes!!

yyscamper:
@benbp , when I change the task to use the Linux Command Job, I find some problem:

- {{options.drivIds}} will be rendered to something like "/dev/sda, /dev/sdb" (pay attention to the comma!), this comma will cause the shell command fails.

- Even I didn't care the comma problem, the command rendering will still fail due to the Hogan error, I copied the error message at end of this comment.  I tried the same input in this demo page (http://mustache.github.io/#demo), it will success. So this should be a problem of our code. I will try to find the root cause laster.

- I change to use shell command like "cmd1; cmd2; cmd3;", so this requires the template should support iteration. But unfortunately, it still fail.  @WangWinson put a fix for this issue in this PR (https://github.com/RackHD/on-tasks/pull/76)  

- Another question is, we use Mustache template engine for task, but use ejs for on-http template file, I think we'd better use the same template engine.

'''
error: 
  stack: 
    - TypeError: Object /dev/disk/by-id/wwn-0x6001636001940aa01dd55983a41675b0,/dev/sda,/dev/sdb has no method 'charAt'
    -     at tagChange (/home/onrack/src/on-core/node_modules/hogan.js/lib/compiler.js:177:14)
    -     at Object.scan (/home/onrack/src/on-core/node_modules/hogan.js/lib/compiler.js:112:13)
    -     at Task.parse (/home/onrack/src/on-tasks/lib/task.js:183:34)
    -     at Task.renderString (/home/onrack/src/on-tasks/lib/task.js:168:41)
    -     at /home/onrack/src/on-tasks/lib/task.js:210:37
    -     at Function.map (/home/onrack/src/on-core/node_modules/lodash/dist/lodash.js:3509:27)
    -     at Task.render (/home/onrack/src/on-tasks/lib/task.js:202:18)
    -     at Task.renderOptions (/home/onrack/src/on-tasks/lib/task.js:225:25)
    -     at /home/onrack/src/on-tasks/lib/task.js:228:31
    -     at /home/onrack/src/on-core/node_modules/lodash/dist/lodash.js:2886:18
    -     at forEach (/home/onrack/src/on-core/node_modules/lodash/dist/lodash.js:3298:15)
    -     at Function.transform (/home/onrack/src/on-core/node_modules/lodash/dist/lodash.js:2885:35)
    -     at Task.renderOptions (/home/onrack/src/on-tasks/lib/task.js:227:22)
    -     at /home/onrack/src/on-tasks/lib/task.js:228:31
    -     at /home/onrack/src/on-core/node_modules/lodash/dist/lodash.js:2886:18
    -     at forOwn (/home/onrack/src/on-core/node_modules/lodash/dist/lodash.js:2106:15)
    -     at Function.transform (/home/onrack/src/on-core/node_modules/lodash/dist/lodash.js:2885:35)
    -     at Task.renderOptions (/home/onrack/src/on-tasks/lib/task.js:227:22)
    -     at Task.renderOwnOptions (/home/onrack/src/on-tasks/lib/task.js:234:29)
    -     at new Task (/home/onrack/src/on-tasks/lib/task.js:126:14)
    -     at Function.create (/home/onrack/src/on-tasks/lib/task.js:398:20)
    -     at create (/home/onrack/src/on-tasks/lib/task.js:407:29)
  arguments: 
    - charAt
    - 
      - /dev/disk/by-id/wwn-0x6001636001940aa01dd55983a41675b0
      - /dev/sda
      - /dev/sdb
  type:      undefined_method
  message:   Object /dev/disk/by-id/wwn-0x6001636001940aa01dd55983a41675b0,/dev/sda,/dev/sdb has no method 'charAt'

benbp:
Yeah, I'd be curious if you can get it working with Winson's changes in #76 as you mentioned.

Agreed, what we do lose is the flexibility for input pre-checking as you said, but I think your suggestion that we should update the command parser logic to handle this is a good one. It will give us something like a flexible command whitelist/validator, but still let people rapidly iterate with custom workflows using shell commands.

In regards to ejs vs. mustache, yeah if we determine we can do the same things with mustache that we are doing with ejs templates right now, it would probably make sense to switch (another reason why having template unit tests will be helpful).

zyoung51:
@benbp Is this the direction we want to shift the templateing to?  The pattern established with the linux command tasks is to expand the template retrieved by downloadUrl and baking the logic/parameterization into the templates in on-http/templates.  Is that methodology obsolete?

Should we not use EJS to iterate through the driveIds when the downloaded template is rendered, ie: 

module.exports = {
    friendlyName: "Clear Drives Partition",
    injectableName: "Task.Drive.Clear.Partition",
    implementsTask: "Task.Base.Linux.Commands",
    options: {
        driveIds: "/dev/disk/by-id/wwn-0x6001636001940aa01dd55961a20fb847 "\
                  "/dev/sda " \
                  "/dev/disk/by-id/ata-SATADOM-SV_3SE_20150522AA999091007C "
        partitionType: "gpt",
                command: 'sudo bash clear_part.sh',
                downloadUrl: '/api/1.1/templates/clear_part.sh'
    },
    properties: {}
};

benbp:
@zyoung51 I think it's nice to be able to support both patterns for different sized templates. In cases like this, where it's just a one-liner that we want repeated a couple of times with different options, I think it's nicer to be able to embed that logic in the template rather than spreading it out across multiple files. In cases like kickstart files, we definitely don't want to embed that within a task definition given the size of the template. So I don't think that methodology is obsolete at all, but I think there should be best practices, based on template size/complexity, for using these different approaches. I think of this part of the system somewhat like how Ansible is setup, where there are many different ways to do the same thing, all with their advantages and disadvantages, that people can choose to use based on preference and best practice. Thoughts?

I know @yyscamper wants to converge on using a single template engine if mustache/hogan can support the same kinds of loops and conditionals that EJS can, but I don't think that should stop us in the near term from submitting more templates to get stuff done.

yyscamper:
Temporary to close this PR, base on the new finding and new design of ORFS-97, we should move the clear partition work to kickstart file, now a brand new job.

If clear partition is a standalone feature in future, I can open this PR again.

PR #73, opened 2015-11-17
testing emptyfunction against build server - dpm

davequick:


benbp:
:+1:

benbp:
Was a test case that reproed the Jenkins issue. Actual fix will be different.

PR #76, opened 2015-11-18
MAG-286 [ORFS-97] Implement RAID creation/deletion in Monorail

WangWinson:
1. Add original Hogan render to support complex template in task render
2. Add Create, Delete RAID tasks

This PR is sub story of ORFS-97

### DIFF
--- a/spec/lib/task-spec.js
+++ b/spec/lib/task-spec.js
@@ -1,6 +1,5 @@
 // Copyright 2015, EMC, Inc.
 /* jshint node:true */
-
 'use strict';
 
 describe("Task", function () {
@@ -12,6 +11,7 @@ describe("Task", function () {
     var noopDefinition;
     var Promise;
     var Constants;
+    var _;
 
     function literalCompare(objA, objB) {
         _.forEach(objA, function(v, k) {
@@ -39,6 +39,7 @@ describe("Task", function () {
         var Logger = helper.injector.get('Logger');
         Logger.prototype.log = sinon.spy();
         Task = helper.injector.get('Task.Task');
+        _ = helper.injector.get('_');
         taskData = taskModule.taskData;
 
         _.forEach(taskData, function(definition) {
@@ -240,6 +241,39 @@ describe("Task", function () {
             expect(task.options.nested1).to.equal(definition.options.sourceValue);
             expect(task.options.nested2).to.equal(definition.options.sourceValue);
             expect(task.options.nested3).to.equal(definition.options.sourceValue);
+        });
+
+        it("should render iteration templates", function() {
+            definition.options = {
+                testList: [
+                    {
+                        name: 'item 1'
+                    },
+                    {
+                        name: 'item 2'
+                    },
+                    {
+                        name: 'item 3'
+                    }
+                ],
+                testVal: '{{#options.testList}}{{ name }}.{{/options.testList}}'
+            };
+            var task = Task.create(definition, {}, {});
+            var tempList = _.transform(definition.options.testList, function (result, n) {
+                result.push(n.name);
+            });
+            expect(task.options.testVal).to.equal(tempList.join('.') + '.');
+        });
+
+        it("should render condition templates", function() {
+            definition.options = {
+                testSrc1: 'Test source 1 exist',
+                testVal1: '{{#options.testSrc1}}{{ options.testSrc1 }}{{/options.testSrc1}}',
+                testVal2: '{{#options.testSrc2}}{{ options.testSrc2 }}{{/options.testSrc2}}'
+            };
+            var task = Task.create(definition, {}, {});
+            expect(task.options.testVal1).to.equal(definition.options.testSrc1);
+            expect(task.options.testVal2).to.equal('');
         });
 
         describe('deferred rendering', function() {
--- a/lib/task-data/tasks/create-megaraid.js
+++ b/lib/task-data/tasks/create-megaraid.js
@@ -1,0 +1,57 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Create RAID via Storcli',
+    injectableName: 'Task.Raid.Create.MegaRAID',
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        createDefault: true,
+        controller: 0,
+        raidList: [],
+        /*[
+            {
+                'enclosure': 255,
+                'type': 'raid1',
+                'drives': [1, 4],
+                'name': 'VD1'
+            },
+            {
+                'enclosure': 255,
+                'type': 'raid5',
+                'drives': [2, 5, 3],
+                'name': 'VD2'
+            },
+            {
+                'enclosure': 255,
+                'type': 'raid10',
+                'drives': [6, 7, 8, 9],
+                'name': 'VD3',
+                'size': '200G',
+                'drivePerArray': 2
+            },
+            {
+                'enclosure': 255,
+                'type': 'raid10',
+                'drives': [6, 7, 8, 9],
+                'name': 'VD4',
+                'size': '800G',
+                'drivePerArray': 2
+            }
+        ],*/
+        path: '/opt/MegaRAID/storcli/storcli64',
+        commands: [
+            '{{#options.createDefault}}' +
+                'sudo {{options.path}} /c{{options.controller}} add vd each type=raid0' +
+            '{{/options.createDefault}}' +
+            '{{^options.createDefault}}{{#options.raidList}}' +
+                'sudo {{options.path}} /c{{options.controller}} add vd type={{type}}' +
+                    '{{#name}} name={{name}}{{/name}}{{#size}} size={{size}}{{/size}} ' +
+                    'drives={{enclosure}}:{{drives}}' +
+                    '{{#drivePerArray}} PDPerArray={{drivePerArray}}{{/drivePerArray}};' +
+            '{{/options.raidList}}{{/options.createDefault}}'
+        ]
+    },
+    properties: {}
+};
--- a/lib/task-data/tasks/delete-megaraid.js
+++ b/lib/task-data/tasks/delete-megaraid.js
@@ -1,0 +1,24 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Delete RAID via Storcli',
+    injectableName: 'Task.Raid.Delete.MegaRAID',
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        deleteAll: true,
+        controller: 0,
+        raidIds: [], //[0,1,2]
+        path: '/opt/MegaRAID/storcli/storcli64',
+        commands: [
+            '{{#options.deleteAll}}' +
+                'sudo {{options.path}} /c{{options.controller}}/vall del force' +
+            '{{/options.deleteAll}}' +
+            '{{^options.deleteAll}}{{#options.raidIds}}' +
+                'sudo {{options.path}} /c{{options.controller}}/v{{.}} del force;' +
+            '{{/options.raidIds}}{{/options.deleteAll}}'
+        ]
+    },
+    properties: {}
+};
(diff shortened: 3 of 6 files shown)


WangWinson:
Need to add require('hogan.js') in on-core/spec/helper.js to make the build pass

spec/lib/task-spec.js
```diff
@@ -242,6 +242,39 @@ describe("Task", function () {
...
+                        name: 'item 2'
+                    },
+                    {
+                        name: 'item 3'
+                    }
+                ],
+                testVal: '{{#options.testList}}{{ name }}.{{/options.testList}}'
+            };
+            var task = Task.create(definition, {}, {});
+            var tempList = _.transform(definition.options.testList, function (result, n) {
```
WangWinson: Should be hound issue. This could pass locally

spec/lib/task-data/tasks/delete-raid-spec.js
```diff
@@ -0,0 +1,36 @@
...
+/* jshint node:true */
+
+'use strict';
+
+describe(require('path').basename(__filename), function () {
+    var base = require('./base-tasks-spec');
+    var hogan = require('hogan.js');
+
+    base.before(function (context) {
+        context.taskdefinition = helper.require('/lib/task-data/tasks/delete-raid.js');
```
WangWinson: Should be hound issue. This could pass locally

spec/lib/task-data/tasks/create-raid-spec.js
```diff
@@ -0,0 +1,36 @@
...
+/* jshint node:true */
+
+'use strict';
+
+describe(require('path').basename(__filename), function () {
+    var base = require('./base-tasks-spec');
+    var hogan = require('hogan.js');
+
+    base.before(function (context) {
+        context.taskdefinition = helper.require('/lib/task-data/tasks/create-raid.js');
```
WangWinson: Should be hound issue. This could pass locally

lib/task-data/tasks/delete-raid.js
```diff
@@ -0,0 +1,31 @@
...
+            }
+        ],*/
+        path: '/opt/MegaRAID/storcli/storcli64',
+        commands: [
+            '{{#options.deleteAll}}\
+				sudo {{options.path}} /c{{options.controller}}/vall del force\
+			{{/options.deleteAll}}\
+			{{^options.deleteAll}}\
+			{{#options.raidIds}}\
+				sudo {{options.path}} /c{{options.controller}}/v{{id}} del force;\
```
yyscamper: @WangWinson , we can use the {{.}} to represent the current array element, so we don't have to change the raidIds  to something like [ {id:0}, {id:1}], we can use the simper one [0, 1]

lib/task-data/tasks/create-raid.js
```diff
@@ -0,0 +1,57 @@
+// Copyright 2015, EMC, Inc.
+/* jshint multistr:true */
+
+module.exports = {
+    friendlyName: 'Create RAID via Storcli',
+    injectableName: 'Task.Raid.Create',
```
benbp: I suggest a name change, since this is specific to MegaRAID commands rather than a generic RAID creation task. Perhaps `Task.MegaRAID.Create` or `Task.Raid.Create.MegaRAID`?

lib/task-data/tasks/create-raid.js
```diff
@@ -0,0 +1,57 @@
...
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        createDefault: true,
+        controller: 0,
+        raidList: [], /*[
+			{
+				'enclosure': 255,
+				'type': 'raid1',
+				'drives': [1, 4],
```
benbp: These are all tab-indented but they should be space-indented.

lib/task-data/tasks/delete-raid.js
```diff
@@ -0,0 +1,32 @@
+// Copyright 2015, EMC, Inc.
+/* jshint multistr:true */
+
+module.exports = {
+    friendlyName: 'Delete RAID via Storcli',
+    injectableName: 'Task.Raid.Delete',
```
benbp: Same naming comment as above.

spec/lib/task-data/tasks/create-raid-spec.js
```diff
@@ -0,0 +1,36 @@
...
+/* jshint node:true */
+
+'use strict';
+
+describe(require('path').basename(__filename), function () {
+    var base = require('./base-tasks-spec');
+    //var hogan = require('hogan.js');
+
+    base.before(function (context) {
+        context.taskdefinition = helper.require('/lib/task-data/tasks/create-raid.js');
```
benbp: Same naming comment, but for the filename, perhaps `create-megaraid.js` or something?

spec/lib/task-data/tasks/create-raid-spec.js
```diff
@@ -0,0 +1,36 @@
...
+            expect(this.taskdefinition).to.have.property('options');
+            expect(this.taskdefinition.options).to.be.an('Object');
+        });
+
+        it('should have commands option', function() {
+            expect(this.taskdefinition.options).to.have.property('commands');
+            expect(this.taskdefinition.options.commands).to.be.an('Array').with.length(1);
+        });
+
+        /*it('command value could be parsed', function() {
```
benbp: Do we want this test? If not, let's remove it.

spec/lib/task-data/tasks/delete-raid-spec.js
```diff
@@ -0,0 +1,36 @@
...
+/* jshint node:true */
+
+'use strict';
+
+describe(require('path').basename(__filename), function () {
+    var base = require('./base-tasks-spec');
+    //var hogan = require('hogan.js');
+
+    base.before(function (context) {
+        context.taskdefinition = helper.require('/lib/task-data/tasks/delete-raid.js');
```
benbp: Same naming comment as above.

spec/lib/task-data/tasks/delete-raid-spec.js
```diff
@@ -0,0 +1,36 @@
...
+            expect(this.taskdefinition).to.have.property('options');
+            expect(this.taskdefinition.options).to.be.an('Object');
+        });
+
+        it('should have commands option', function() {
+            expect(this.taskdefinition.options).to.have.property('commands');
+            expect(this.taskdefinition.options.commands).to.be.an('Array').with.length(1);
+        });
+
+        /*it('command value could be parsed', function() {
```
benbp: Do we want this commented out test?

spec/lib/task-spec.js
benbp: Yeah we make it available globally in the test helper (https://github.com/RackHD/on-core/blob/master/spec/helper.js#L15), but the .jshintrc doesn't mark global._ as a lint exception. That said, I'd like to try moving away from using too many test globals, so if you don't mind switching to `var _ = helper.injector.get('_')` that would be nice.

spec/lib/task-spec.js
```diff
@@ -242,6 +242,39 @@ describe("Task", function () {
...
+
+        it("should render condition templates", function() {
+            definition.options = {
+                testSrc1: 'Test source 1 exist',
+                testVal1: '{{#options.testSrc1}}{{ options.testSrc1 }}{{/options.testSrc1}}',
+                testVal2: '{{#options.testSrc2}}{{ options.testSrc2 }}{{/options.testSrc2}}'
+            };
+            var task = Task.create(definition, {}, {});
+            expect(task.options.testVal1).to.equal(definition.options.testSrc1);
+            expect(task.options.testVal2).to.equal('');
```
benbp: Hm, okay, so general question to you here: Currently in the task rendering code we have decided to fail task validation if a render string references a value that does not exist, whereas with normal hogan rendering, it looks like it will just bypass that. Do you think we should care about enforcing that behavior for complex templates, or just decide to render certain values as empty silently as this is doing?

benbp: On the other hand, I guess if we really want to enforce certain values, we can add them to the requiredOptions object.

lib/task-data/tasks/delete-raid.js
```diff
@@ -0,0 +1,32 @@
...
+        controller: 0,
+        raidIds: [], /*[
+            {
+                id: 0
+            },
+            {
+                id: 1
+            }
+        ],*/
+        path: '/opt/MegaRAID/storcli/storcli64',
```
benbp: Seems like it would be useful for us to add `chdir` support or something to the command object so you don't have to do this.

spec/lib/task-data/tasks/create-raid-spec.js
```diff
@@ -0,0 +1,36 @@
...
+        it('should have an options', function() {
+            expect(this.taskdefinition).to.have.property('options');
+            expect(this.taskdefinition.options).to.be.an('Object');
+        });
+
+        it('should have commands option', function() {
+            expect(this.taskdefinition.options).to.have.property('commands');
+            expect(this.taskdefinition.options.commands).to.be.an('Array').with.length(1);
+        });
+
```
benbp: I'd like to see tests for the create/delete raid tasks actually define the tasks and render them, then compare their commands output against the expected, that way we know the rendering results in the command we want, and it also provides some informal documentation to future developers about what we expect the template to look like after being rendered. Also, we protect ourselves in case anyone makes changes to the command string in the future.

benbp:
Also, I should mention, this is super cool. I'm really excited to have this feature and start defining tasks like this!

spec/lib/task-spec.js
WangWinson: OK, sounds reasonable. will update

lib/task-data/tasks/delete-raid.js
WangWinson: Good catch. Will update accordingly.

lib/task-data/tasks/create-raid.js
WangWinson: ```Task.Raid.Create.MegaRAID```  would be better. Will update

benbp:
Adding a ticket for follow-on documentation: https://hwjiraprd01.corp.emc.com/browse/MON-669

spec/lib/task-spec.js
```diff
@@ -39,6 +39,7 @@ describe("Task", function () {
         var Logger = helper.injector.get('Logger');
         Logger.prototype.log = sinon.spy();
         Task = helper.injector.get('Task.Task');
+        _ = helper.injector.get('_');
```
WangWinson: Looks like .jshintrc doesn't mark global.helper as a lint exception. 
This could pass locally

spec/lib/task-spec.js
WangWinson: I think we can keep the way which normal hogan rendering. Since the certain value was put in the condtional section {{#condition}}  {{/condtion}}. It's reasonable to skip anything inside the condition if the condition doesn't meet. 
While, if we just put an not existing value {{value}} to template, hogan would also report error during render time.

spec/lib/task-data/tasks/create-raid-spec.js
WangWinson: The test case needed for verify command template could be parsed. So I would uncomment this and make it work

spec/lib/task-data/tasks/create-raid-spec.js
WangWinson: Good point. Should add the real case

WangWinson:
@benbp commit again to address you comments.
1.rename task file and injectable name
2.add more test case to test command template in task
3.fix some style issue.

spec/lib/task-spec.js
benbp: Ah okay, I didn't quite make the connection that this was just a normal conditional. Yeah, what you are saying sounds reasonable to me as well. Nice to have this feature.

spec/lib/task-spec.js
benbp: Yep, working on getting hound to ignore this.

benbp:
:+1: thanks!

jlongever:
+1

PR #78, opened 2015-11-20
task cleanup

heckj:
removing unused and un-needed options

### DIFF
--- a/lib/task-data/tasks/install-coreos.js
+++ b/lib/task-data/tasks/install-coreos.js
@@ -7,10 +7,6 @@ module.exports = {
     injectableName: 'Task.Os.Install.CoreOS',
     implementsTask: 'Task.Base.Os.Install',
     options: {
-        // NOTE: user/pass aren't used by the coreos installer at the moment,
-        // but they are required values
-        username: 'root',
-        password: 'root',
         profile: 'install-coreos.ipxe',
         comport: 'ttyS0',
         hostname: 'coreos-node',


lib/task-data/tasks/install-coreos.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
     implementsTask: 'Task.Base.Os.Install',
     options: {
         // NOTE: user/pass aren't used by the coreos installer at the moment,
-        // but they are required values
+        // but they are required values for Task.Base.Os.Install
```
yyscamper: From the definition of Task.Base.Os.Install, the user/password are not required (at least now).
Meanwhile, by the design of "Parameterized OS Bootstrap", no option for "username", the name for the password for root account should be "rootPassword"

benbp: Yeah, this is right. 

This was actually the original reasoning behind why we had base task definitions be separate from jobs, so that you could have different schemas/required options for the same job, but in practice, it seemed to be mostly redundant rather than useful. But I wonder whether it's an idea worth holding onto when we refactor base task definitions into the job classes.

heckj: The reason for the comment was that I found the pass-through of values in the declarative JSON blobs to be fairly confusion - not sure what's defined where, what's required, and what's not. The CoreOS installer, in fact, doesn't have any capacity to specify this - the most you can tweak there is the SSH key that allows you to log into the base account for their system. An outlier in the OS install space, but worth noting.

benbp: Does it work if you remove the options?

heckj: Good question, I'll remove and see - I really don't know.

heckj: so... apparently it does!, huh - will update the PR to just remove that stuff since it doesn't use it OR require it

benbp:
Oh good, glad it worked.

DavidjohnBlodgett:
:+1:

PR #80, opened 2015-11-25
Modified lsscsi command parser to combine phrases correctly

iceiilin:
Original lsscsi parser combines two phrases in one item, and doesn't count in the case where there are three phrases, making array index wrong after processing.
For example,
input: '[10:0:0:0]   disk    ATA      32GB SATA Flash  SFDE  /dev/sda   32.0GB'
Current:
output: ['[10:0:0:0]','disk','ATA',**'32GB SATA','Flash'**,'SFDE','/dev/sda','32.0GB']
This PR:
output: ['[10:0:0:0]','disk','ATA',**'32GB SATA Flash'**,'SFDE','/dev/sda','32.0GB']

### DIFF
--- a/lib/utils/job-utils/command-parser.js
+++ b/lib/utils/job-utils/command-parser.js
@@ -615,9 +615,14 @@ function commandParserFactory(Logger, Promise, _) {
                     // Grab lsscsi information
 
                     split = _.compact(line.split(' '));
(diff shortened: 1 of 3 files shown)


benbp:
Ah, good fix. :+1:

heckj:
+1

PR #85, opened 2015-12-03
support specify install disk for install-os

yyscamper:
- add a new option 'installDisk' in install-os job.
 - installDisk support both drive name (string) or drive index (number).
 - if use specify drive index, install-job will lookup the 'driveId' catalog to find the corresponding drive wwid.

Below is the dependent PR:
https://github.com/RackHD/on-http/pull/58

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -1,5 +1,5 @@
-// Copyright 2015, EMC, Inc.
 
+// Copyright 2015, EMC, Inc.
 'use strict';
 
 var di = require('di');
@@ -13,7 +13,10 @@ di.annotate(installOsJobFactory, new di.Provide('Job.Os.Install'));
         'Assert',
         'Util',
         '_',
-        'Services.Encryption'
+        'Services.Encryption',
+        'Promise',
+        'JobUtils.CatalogSearchHelpers',
+        'Services.Waterline'
     )
 );
 
@@ -23,7 +26,10 @@ function installOsJobFactory(
     assert,
     util,
     _,
-    encrypt
+    encrypt,
+    Promise,
+    catalogSearch,
+    waterline
 ) {
     var logger = Logger.initialize(installOsJobFactory);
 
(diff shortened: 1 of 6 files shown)


yyscamper:
@zyoung51

spec/lib/jobs/install-os-job-spec.js
```diff
@@ -87,21 +93,138 @@ describe('Install OS Job', function () {
...
+                return job._convertInstallDisk();
+            }).then(function() {
+                expect(job.options.installDisk).to.equal('naa.rstuvw');
+            });
+        });
+
+        it('should set correct installDisk linux wwid', function() {
+            job.options.osType = 'linux';
+            return Promise.resolve().then(function() {
+                return job._convertInstallDisk();
```
benbp: FYI, a nice way you can also do this is like:

`return expect(job._convertInstallDisk()).to.become('/dev/test1');`

lib/jobs/install-os.js
```diff
@@ -148,24 +154,76 @@ function installOsJobFactory(
...
+        var disk = self.options.installDisk;
+
+        //If disk is empty then do nothing.
+        //If disk is string, it means user directly input the drive wwid, so don't need conversion.
+        if ((!disk && disk !== 0) || _.isString(disk)) {
+            return;
+        }
+
+        if (!_.isNumber(disk)) {
+            throw new Error('The installDisk format is not correct');
```
benbp: Just a little nitpick: I generally like to normalize what comes out of promise returning functions, so doing a `Promise.reject(new Error(...))`.

benbp:
Just some minor style comments, up to you if you care to address them.

:+1: regardless.

spec/lib/jobs/install-os-job-spec.js
yyscamper: @benbp , Your suggestion is good and much simpler than my design, However what I really want to verify is the job.options.installDisk is correctly converted after called function _convertInstallDisk, not the promise resolved value. 

Change _convertInstallDisk to return the job.options.installDisk will make the ".to.become" works, but I still need some logic to ensure the "options.installDisk" is also changed to the resolved value.

anhou:
+1

PR #87, opened 2015-12-04
GRIP-22: Add support for booting LiveCD

zyoung51:
@RackHD/corecommitters

### DIFF
--- a/lib/task-data/tasks/boot-livecd.js
+++ b/lib/task-data/tasks/boot-livecd.js
@@ -1,0 +1,22 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Boot LiveCD',
+    injectableName: 'Task.Os.Boot.LiveCD',
+    implementsTask: 'Task.Base.Os.Install',
+    options: {
+        profile: 'boot-livecd.ipxe',
+        completionUri: 'renasar-ansible.pub',
+        version: 'livecd',
+        repo: '{{api.server}}/LiveCD/{{options.version}}'
+    },
+    properties: {
+        os: {
+            linux: {
+                distribution: 'livecd'
+            }
+        }
+    }
+};


lib/task-data/tasks/boot-livecd.js
```diff
@@ -0,0 +1,22 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Boot LiveCD',
+    injectableName: 'Task.Os.Boot.LiveCD',
+    implementsTask: 'Task.Base.Os.Install',
+    options: {
+        profile: 'boot-livecd.ipxe',
+        completionUri: 'renasar-ansible.pub',
```
yyscamper: I didn't see you reference the completionUri in your corresponding on-http PR (https://github.com/RackHD/on-http/pull/61) , so how we notify the ORA the installation success?

lib/task-data/tasks/boot-livecd.js
```diff
@@ -0,0 +1,22 @@
...
+        version: 'livecd',
```
yyscamper: Regularly the version should be number or release nick name, whether we can change it to another reasonable name?

lib/task-data/tasks/boot-livecd.js
zyoung51: Not sure what you are talking about... The booted OS queries the completionUri to notify the boot has completed.  Why would IPXE reference it?

lib/task-data/tasks/boot-livecd.js
zyoung51: The ISO image is user generated and can be any platform/version.  As there are no fixed/expected names to start with, 'livecd' seems like a good guess at a name that someone might use when creating their own livecd.

jlongever:
+1

lib/task-data/tasks/boot-livecd.js
yyscamper: My meaning is how this Boot LiveCD task notify the its has success? Where do you defined to query the compltionUri?

zyoung51: it is implementation specific on the LiveCD.  In the LiveCD I created, I issue a wget to retrieve a 'completionUrl' of 'init-commands.sh' which is a template of more commands to run.

yyscamper: Do you have any script to help create the LiveCD so that ensure the completionUrl query is added into it? Or any document?

Previously, the completionUrl query is in our kickstart script, we are in the control of it. But if we define it in LiveCD, we pass the control to user. If user forgets this customization work, our workflow cannot detect this and the workflow will eventually hang until timeout.

Do you have every study whether we can always inject some command into the LiveCD booting, so we can take this to notify the boot LiveCD success.

brianparry:
:+1:

PR #89, opened 2015-12-07
Avoid getting default install disk when user assigns one in os instal…

iceiilin:
Provide more flexibility for user's configuration. In case the default satadom os installation failed for some reason, users can assign a disk by their own.

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -191,7 +191,12 @@ function installOsJobFactory(
         var self = this;
         return Promise.resolve()
         .then(function () {
-            return self._getInstallDisk();
+            if (self.options.hasOwnProperty('installDisk') &&
+                self.options.installDisk !== '') {
+                return Promise.resolve();
+            } else {
+                return self._getInstallDisk();
+            }
         })
         .then(function () {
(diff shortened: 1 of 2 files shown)


heckj:
I don't mean to make this harder under a deadline, but the business logic you slipped in there (with good reason) would be perfectly suited to a unit test to verify the inputs and outputs of that job all functioned exactly as expected. Would you please consider adding a unit test to verify that the disk is set to an empty string when appropriate based on your logic?

lib/jobs/install-os.js
```diff
@@ -191,7 +191,13 @@ function installOsJobFactory(
         var self = this;
         return Promise.resolve()
         .then(function () {
-            return self._getInstallDisk();
+            if (self.options.hasOwnProperty('installDisk') &&
+                self.options.installDisk !== '') {
+                return Promise.resolve();
+            }
+            else {
```
anhou: :+1: just style "} else {" is better

anhou:
:+1:  after as joe suggest refine the unit test

yyscamper:
:+1:

iceiilin:
Unit test added.

PR #90, opened 2015-12-09
adding placeholder option to optionally add KVM packages

heckj:
coordinates with https://github.com/RackHD/on-http/pull/67 to enable adding KVM to CentOS/RHEL installs

### DIFF
--- a/spec/lib/jobs/ansible-job-spec.js
+++ b/spec/lib/jobs/ansible-job-spec.js
@@ -3,8 +3,7 @@
 
 'use strict';
 
-var uuid = require('node-uuid'),
-    events = require('events');
+var uuid = require('node-uuid');
 
 describe(require('path').basename(__filename), function () {
     var base = require('./base-spec');
@@ -33,8 +32,6 @@ describe(require('path').basename(__filename), function () {
     });
 
     describe("ansible-job", function() {
-        var Snmptool;
-        var Constants;
         before(function() {
         });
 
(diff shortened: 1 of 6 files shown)


benbp:
:+1:

spec/lib/jobs/message-cache-job-spec.js
```diff
@@ -108,7 +108,8 @@ describe("Message Cache Job", function () {
             this.sandbox.stub(job, 'createSetIpmiCommandResultCallback').returns('fn return stub');
             job._run();
             expect(job._subscribeIpmiCommandResult.callCount).to.equal(5);
-            _.forEach(['sdr', 'selInformation', 'sel', 'chassis', 'driveHealth'], function(command) {
+            _.forEach(['sdr', 'selInformation', 'sel', 'chassis', 'driveHealth'],
```
heckj: safely ignorable - loaded in spec helper as a global lodash instance for tests

yyscamper:
:+1:

spec/lib/jobs/ansible-job-spec.js
```diff
@@ -3,8 +3,7 @@
 
 'use strict';
 
-var uuid = require('node-uuid'),
-    events = require('events');
+var uuid = require('node-uuid');
```
yyscamper: Seems this file change doesn't belong to this commit as not relation with kvm. How about move it to another commit?

heckj: I think you just wanted to see if I could rebase a single commit back out into multiple commit. Never one to stand down from a dare... Viola!

heckj:
For future reference, the 4GB Installation ISO that I used to provide my testing repository doesn't include kvm or python-virtinst, but otherwise I've verified these combined patches work to provide an optional KVM installation, ready to merge!

PR #94, opened 2015-12-17
Add IPMI over LAN lanplus support

anhou:
IPMI 2.0 support both lanplus and lan, adding lanplus support, it's tested on Dell servers, EMC internal servers. also remove the old comment out lines. 
@RackHD/rackhd_dev @RackHD/corecommitters


heckj:
+1

brianparry:
:+1:

PR #97, opened 2015-12-24
Quick Fix for issue on install CentOS error w/o rootPassword

cgx027:
Issue link: https://github.com/RackHD/RackHD/issues/36
Replace the 'null' default rootPassword.

The long term solution discuss is still undergoing at github issue
page. This is rather a short term fix to make it workable immediatly


lib/task-data/tasks/install-centos.js
```diff
@@ -12,12 +12,12 @@ module.exports = {
         profile: 'install-centos.ipxe',
         hostname: 'localhost',
         comport: 'ttyS0',
-        domain: 'rackhd.github.com',
+        domain: 'rackhd',
```
heckj: Was reducing this to just RackHD intentional?

heckj:
Question about one of the changes, but :+1: for the quick fix solution while we sort out longer term defaults

lib/task-data/tasks/install-centos.js
cgx027: This was suggested by @yyscamper. This is also intended to keep consistent with the install-ESXi job where domain is set to 'rackhd'. It is suggested that having github.com in the domain makes rackhd less visible. I am open to this change. Let me konw if you are preferring to have github.com in the domain.

cgx027:
Also copy @RackHD/corecommitters  and folks in Shanghai, @WangWinson , @yyscamper , @iceiilin , @anhou .

lib/task-data/tasks/install-centos.js
heckj: It's fine, just wanted to make sure there was intention behind it. A single word as a domain is uncommon, but as a placeholder/reference example, it's perfectly acceptable

yyscamper:
:+1:

PR #86, opened 2015-12-04
Install os in satadom

iceiilin:
It is the fix of ODR-338, https://hwjiraprd01.corp.emc.com/browse/ODR-338
Previous kickstart file assigned default disk for installation (sda for RHEL and firstdisk for ESXi). It will cause RHEL failed to boot if "sda" and "firstdisk" is not the same when switching between these two OS.

The fix includes:
1. added driveId catalog to get wwid - drive mapping table (on-tasks, on-http and on-taskgraph)
2. Choosed SATADOM as installed disk based on driveId catalog (on-tasks)
3. Modified kickstart file to render installDisk info got from job (on-http)

Related PRs:
https://github.com/RackHD/on-taskgraph/pull/36
https://github.com/RackHD/on-http/pull/60

PR #100, opened 2015-12-29
Update install os job to handle more input case of installDisk param

WangWinson:
- set default value ```firstdisk``` or ```sda``` when driveId catalog does not exist
- throw error if installDisk is number but could not found in driveId catalog
- when installDisk is empty, use wwid of SATADOM, if not exist use the first device in driveId catalog
- set installDisk default to null in install-esx and install-centos task definition

@RackHD/corecommitters @iceiilin @sunnyqianzhang @pengz1 @cgx027 
This change is similar to previous PR https://github.com/RackHD/on-tasks/pull/86 on 1.0 branch. Also, it covers https://github.com/RackHD/RackHD/issues/53

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -171,7 +171,7 @@ function installOsJobFactory(
         if (_.isString(disk) && !_.isEmpty(disk)) {
             return Promise.resolve(disk);
         }
-
+        //If disk is not number and not empty value, it should be reject for incorrect format
         if (!_.isNumber(disk) && !_.isEmpty(disk)) {
             return Promise.reject(new Error('The installDisk format is not correct'));
         }
@@ -248,7 +248,7 @@ function installOsJobFactory(
 
      /**
      * Search the driveid catalog and lookup the corresponding drive WWID of SATADOM
-     * @param {Object} catalog - the catalog data of drive id
+     * @param {Array} catalog - the catalog data of drive id
      * @param {Boolean} isEsx - True to return the ESXi formated wwid,
      *                          otherwise linux format wwid.
      * @return {String} The WWID of SATADOM. If failed, return null
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -167,32 +167,45 @@ function installOsJobFactory(
         var self = this;
         var disk = self.options.installDisk;
 
-        //If disk is empty then do nothing.
         //If disk is string, it means user directly input the drive wwid, so don't need conversion.
-        if ((!disk && disk !== 0) || _.isString(disk)) {
+        if (_.isString(disk) && !_.isEmpty(disk)) {
             return Promise.resolve(disk);
         }
 
-        if (!_.isNumber(disk)) {
+        if (!_.isNumber(disk) && !_.isEmpty(disk)) {
             return Promise.reject(new Error('The installDisk format is not correct'));
         }
 
-        //if it is integer, we think it is drive index, so we need to lookup the database to map
-        //the drive index to drive WWID
         return waterline.catalogs.findMostRecent({
             node: self.nodeId,
             source: 'driveId'
         }).then(function(catalog) {
-            var wwid = catalogSearch.findDriveWwidByIndex(
-                catalog.data,
-                (self.options.osType.toLowerCase() === 'esx'),
-                disk
-            );
-
-            if (!wwid) {
-                return Promise.reject(new Error('Fail to find the WWID for installDisk'));
+            var isEsx = /^esx$/i.test(self.options.osType);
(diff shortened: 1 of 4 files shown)


spec/lib/jobs/install-os-job-spec.js
```diff
@@ -255,5 +260,32 @@ describe('Install OS Job', function () {
...
+                return job._convertInstallDisk();
+            }).then(function() {
+                expect(job.options.installDisk).to.equal('sda');
+            });
+        });
+
+        it('should set firstdisk if installDisk is null and catalog is empty (ESXi)', function() {
+            job.options.osType = 'esx';
+            job.options.installDisk = null;
+            waterline.catalogs.findMostRecent = sinon.stub().resolves({});
```
WangWinson: sinon is global , could be ignore

iceiilin:
+1

lib/jobs/install-os.js
```diff
@@ -167,32 +167,45 @@ function installOsJobFactory(
...
         }).then(function(catalog) {
-            var wwid = catalogSearch.findDriveWwidByIndex(
-                catalog.data,
-                (self.options.osType.toLowerCase() === 'esx'),
-                disk
-            );
-
-            if (!wwid) {
-                return Promise.reject(new Error('Fail to find the WWID for installDisk'));
+            var isEsx = /^esx$/i.test(self.options.osType);
```
sunnyqianzhang: Does this mean a variable starting with 'e' and ending with 'x'?

lib/jobs/install-os.js
```diff
@@ -233,5 +246,23 @@ function installOsJobFactory(
         });
     };
 
+     /**
+     * Search the driveid catalog and lookup the corresponding drive WWID of SATADOM
+     * @param {Object} catalog - the catalog data of drive id
```
yyscamper: catalog is Array

spec/lib/jobs/install-os-job-spec.js
```diff
@@ -231,9 +232,13 @@ describe('Install OS Job', function () {
             return expect(job._convertInstallDisk()).to.be.rejectedWith(Error);
         });
 
-        it('should do nothing if installDisk is not specified', function() {
+        it('should set SATADOM wwid as default if installDisk is not specified', function() {
             job.options.installDisk = null;
-            return expect(job._convertInstallDisk()).to.not.be.rejected;
+            return Promise.resolve().then(function() {
```
yyscamper: Since now _convertInstallDisk always return promise, we can remove this line in point of simplicity.

yyscamper:
- the task definition of install centos and install exs should be updated to default set installDisk as null or undefined, otherwise the feature of finding default SATADom will not work well.

- it's better to change the commit message, MAG-xx and ODR-xxx is used internally, it's not used in open source, meanwhile, only write the issue number doesn't give clear message what you have done in this PR

lib/jobs/install-os.js
```diff
@@ -167,32 +167,45 @@ function installOsJobFactory(
...
 
-        //If disk is empty then do nothing.
         //If disk is string, it means user directly input the drive wwid, so don't need conversion.
-        if ((!disk && disk !== 0) || _.isString(disk)) {
+        if (_.isString(disk) && !_.isEmpty(disk)) {
             return Promise.resolve(disk);
         }
 
-        if (!_.isNumber(disk)) {
+        if (!_.isNumber(disk) && !_.isEmpty(disk)) {
```
yyscamper: it's better to have a comment here why the disk can be empty, if people is not familiar with our design, he may feel a little confused

lib/jobs/install-os.js
WangWinson: This means the ```osType``` should equals ```'esx'``` in lower case

lib/jobs/install-os.js
WangWinson: Good catch. Copy paste issue

spec/lib/jobs/install-os-job-spec.js
WangWinson: updated unit test to remove the ```Promise.resolve()```

lib/jobs/install-os.js
WangWinson: Comment added

WangWinson:
@yyscamper , as to the commit message, I think the story/ODR number is useful for issue tracking purpose. So that people who review this PR could directly find the related story/ODR item.

I agree that it will be confused to external user what's the story/ODR number means. I think the better way is to put a related github issue number. But currently , we still using the ODR system and do not have one to one matched github issue. I bring up this topic on today's daily meeting, hopefully we could found out a better way thru discussion.

yyscamper:
+1

PR #102, opened 2016-01-06
modify generate enclosure job to not override other relation type

iceiilin:
Update generate enclosure job.

In previous version, code works as there is only one type of relation (enclosedBy - encloses). Every time when this relation type needs to be updated, the whole relation field will be updated to this relation type.

This PR covers cases where there will be other relation type existing in this node, thus "generate enclosure job" will only update (enclosedBy - encloses) relation type without overriding other types.

### DIFF
--- a/lib/jobs/generate-enclosure.js
+++ b/lib/jobs/generate-enclosure.js
@@ -219,7 +219,8 @@ function generateEnclosureJobFactory(
     }
 
     /**
-     * Get target nodes of the encloses or enclosedBy relation
+     * Get target nodes of the encloses or enclosedBy relation, and
+     * remove this relation type from node's relations
      *
      * @param {object} node
      * @return {array} target nodes
--- a/lib/jobs/generate-enclosure.js
+++ b/lib/jobs/generate-enclosure.js
@@ -82,14 +82,14 @@ function generateEnclosureJobFactory(
                 var enclData = {
                     name: 'Enclosure Node ' + matchInfo.sn,
                     type: TYPE_ENCL,
-                    relations: [{
-                        relationType: RELATION_ENCL,
-                        targets: []
-                    }]
+                    relations: []
                 };
 
                 return waterline.nodes.create(enclData)
                 .then(function(node) {
+                    if (!node) {
+                        return Promise.reject('Could not create enclosure node');
+                    }
                     enclNode = node;
                     logger.debug('No matched enclosure, create a new one', {
                         id: self.nodeId,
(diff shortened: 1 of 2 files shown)


spec/lib/jobs/generate-enclosure-spec.js
```diff
@@ -318,6 +329,115 @@ describe(require('path').basename(__filename), function () {
...
+            };
+
+            this.sandbox.stub(mockWaterline.nodes,'create');
+            this.sandbox.stub(mockWaterline.nodes,'find').resolves(encl);
+            this.sandbox.stub(mockWaterline.nodes,'findByIdentifier').resolves(node);
+            return job.run()
+            .then(function() {
+                expect(mockWaterline.nodes.create).to.not.have.been.called;
+                expect(mockWaterline.nodes.updateByIdentifier)
+                .to.have.been.calledWith(enclId, enclRelation);
```
yyscamper: Increase the indent for this line should be better.

spec/lib/jobs/generate-enclosure-spec.js
```diff
@@ -318,6 +329,115 @@ describe(require('path').basename(__filename), function () {
...
+            this.sandbox.stub(mockWaterline.nodes,'create');
+            this.sandbox.stub(mockWaterline.nodes,'find').resolves(encl);
+            this.sandbox.stub(mockWaterline.nodes,'findByIdentifier').resolves(node);
+            return job.run()
+            .then(function() {
+                expect(mockWaterline.nodes.create).to.not.have.been.called;
+                expect(mockWaterline.nodes.updateByIdentifier)
+                .to.have.been.calledWith(enclId, {relations: enclRelation});
+                expect(mockWaterline.nodes.updateByIdentifier)
+                .to.have.been.calledWith(nodeId, {relations: nodeRelation});
```
yyscamper: Increase the indent for this line should be better. And line#407

yyscamper:
:+1:

spec/lib/jobs/generate-enclosure-spec.js
iceiilin: Updated

brianparry:
:+1:

lib/jobs/generate-enclosure.js
```diff
@@ -203,5 +218,32 @@ function generateEnclosureJobFactory(
         });
     }
 
+    /**
+     * Get target nodes of the encloses or enclosedBy relation
```
heckj: could you improve this to reflect what the method is doing? It's not just getting the node, but removing the relation as well. "pop" to me isn't clear as to the intent

spec/lib/jobs/generate-enclosure-spec.js
```diff
@@ -318,6 +329,115 @@ describe(require('path').basename(__filename), function () {
...
+            this.sandbox.stub(mockWaterline.nodes,'create');
+            this.sandbox.stub(mockWaterline.nodes,'find').resolves(encl);
+            this.sandbox.stub(mockWaterline.nodes,'findByIdentifier').resolves(node);
+            return job.run()
+            .then(function() {
+                expect(mockWaterline.nodes.create).to.not.have.been.called;
+                expect(mockWaterline.nodes.updateByIdentifier)
+                    .to.have.been.calledWith(enclId, {relations: enclRelation});
+                expect(mockWaterline.nodes.updateByIdentifier)
+                    .to.have.been.calledWith(nodeId, {relations: nodeRelation});
```
heckj: could you also verify the expected state of the node after this sequence? The test verifies that methods were called or not, but doesn't appear to be verifying that the nodes (and content of those objects) remain unchanged

spec/lib/jobs/generate-enclosure-spec.js
```diff
@@ -318,6 +329,115 @@ describe(require('path').basename(__filename), function () {
...
+            };
+
+            this.sandbox.stub(mockWaterline.nodes,'create');
+            this.sandbox.stub(mockWaterline.nodes,'find').resolves(encl);
+            this.sandbox.stub(mockWaterline.nodes,'findByIdentifier').resolves(node);
+            return job.run()
+            .then(function() {
+                expect(mockWaterline.nodes.create).to.not.have.been.called;
+                expect(mockWaterline.nodes.updateByIdentifier)
+                    .to.have.been.calledWith(enclId, enclRelation);
```
heckj: could you also verify the expected state of the node after this sequence? The test verifies that methods were called or not, but doesn't appear to be verifying that the nodes (and content of those objects) remain unchanged

heckj:
Looks good - I made a request for a slight update in the comments of a method function and I think adding explicit node state validation would improve the unit tests significantly in value.

spec/lib/jobs/generate-enclosure-spec.js
iceiilin: Since the final node content is determined by external promise "updateByIdentifier", there is no convenient way to verify its output. The node's content is changed by this function, whose input has been verified by these two statements above to make sure only the (encloses-enclosedBy) relation will be modified, and it is reliable to pass this test case. Any thoughts?

lib/jobs/generate-enclosure.js
iceiilin: updated

iceiilin:
Made the change accordingly and feedback to the comments inline.

PR #104, opened 2016-01-08
modify generate enclosure job to not override other relation type

iceiilin:
PR to 1.1.0 branch, the same as https://github.com/RackHD/on-tasks/pull/102 that has been merged into master.

Without this PR, the enclosure node that doesn't have "encloses" relationType previously will not be updated to include information about compute nodes. This problem is only exposed with the fix of ODR-314 (https://github.com/RackHD/on-http/pull/88, https://github.com/RackHD/on-core/pull/62) about adding the deleting and updating enclosure/compute node.


heckj:
:+1:

jlongever:
Can one of the admins verify this patch?

anhou:
:+1:

PR #105, opened 2016-01-13
Modify local ipmi command

iceiilin:
Fix ODR-339 (https://hwjiraprd01.corp.emc.com/browse/ODR-399) reporting no poller for management server.

The missing of poller is caused by missing of "bmc" source in catalogs, which is the response from "ipmitool lan print" command. In the catalog-mgmt-bmc task, the command is "ipmitool lan print || true",  which is sent to a child process as parameters to ipmitool. the "|| true" is invalid parameters and caused the stderr in the response.

The solution is to remove "|| true" from command, and store it as an ignoreFailure flag to determine whether other commands will continue to process.


lib/jobs/ipmi-catalog.js
```diff
@@ -110,11 +110,26 @@ function ipmiCatalogJobFactory(
         assert.string(command);
         assert.object(obmSettings);
 
-        return [ 
+        // Some commands end with " || true"
+        // This tail needs to be deleted from the command args
+        // Otherwise the tail passed to ipmitool child process will cause error response
+        // this tail is stored and used to determine whether continue sending other commands
+        var ignore = false;
+        var tailIdx = command.indexOf(' ||');
```
yyscamper: Why not only change the command to remove the "|| true" in the task definition of [catalog-mgmt-bmc.js](https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/catalog-mgmt-bmc.js). I think this kind of "smart' conversion is not necessary and waste.

heckj: agreed - we control what's being requested there, no reason to put in logic to remove it here when we can directly update the workflow.

iceiilin: Joel added "|| true" before in compute and mgmt node discovery, as well as command parser (https://github.com/RackHD/on-tasks/pull/35, https://github.com/RackHD/on-tasks/pull/60). The aim is to not block commands after the failure one and collect other catalog information. Although the failure case is rare when using child process in this job, I still would like to cover it and use this flag to catch error and continue processing (https://github.com/iceiilin/on-tasks/blob/mgmt_srv/lib/jobs/ipmi-catalog.js#L160). Also I don't plan to change command parser that is used commonly at this time (since the same PR will be open to 1.1.0 branch too)

yyscamper: Didn't realize remove the "|| true" will cause not able to find the command parser, so I think this is not a good design for looking up command parser, we need improvement for it.

If "|| true" is a kind of flag to ignore failure, I would prefer to explicitly define a flag for each command to indicate whether ignore error.

Considering the urgency of  1.1.x release, I agree to merge into 1.1.x branch, but we can improve this in master.

brianparry: @iceiilin `var tailIdx = command.indexOf(' ||');` is sensitive to whitespace.  I'd suggest using a regular expression for this.  Perhaps something like: `command = command.split(/\s*||/)[0];`

iceiilin: Thanks @brianparry . After investigation, it is found that "|| true" is not the best solution here, since tasks already uses acceptedResponseCodes as a flag. Will close this PR and open another for a better structure.

PR #107, opened 2016-01-15
Modify local ipmi command to remove "ignoreFailure" flag

iceiilin:
Quick fix ODR-339 (https://hwjiraprd01.corp.emc.com/browse/ODR-399) reporting no poller for management server.

The missing of poller is caused by missing of "bmc" source in catalogs, which is the response from "ipmitool lan print" command. In the catalog-mgmt-bmc task, the command is "ipmitool lan print || true", which is sent to a child process as parameters to ipmitool. the "|| true" is invalid parameters and caused the stderr in the response.

The solution is to remove "|| true" from command, and store it as an ignoreFailure flag to determine whether other commands will continue to process.

It might not be the best solution here, just a quick fix of 1.1.0 branch without modifying other common codes.
A more clear way about 1) adding ignore flag in workflow and 2) modify command parser will be added to master branch.

### DIFF
--- a/lib/jobs/ipmi-catalog.js
+++ b/lib/jobs/ipmi-catalog.js
@@ -21,13 +21,13 @@ di.annotate(ipmiCatalogJobFactory, new di.Provide('Job.LocalIpmi.Catalog'));
     )
 );
 function ipmiCatalogJobFactory(
-    BaseJob, 
-    parser, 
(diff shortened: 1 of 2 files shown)


anhou:
:+1:

yyscamper:
:+1:

PR #111, opened 2016-01-27
Updated so that the dhcp lease poller parsed data does not include expired leases

tldavies:


tldavies:
@DavidjohnBlodgett  @RackHD/corecommitters

lib/jobs/isc-dhcp-lease-poller.js
```diff
@@ -144,6 +144,10 @@ function iscDhcpLeasePollerFactory(
         */
         var split = data.split('\n');
         var lease;
+        var todaysDate = new Date();
```
benbp: Can we call this currentTime or something? `todaysDate` just makes me think of a date at the day level, not one that's precise to the millisecond.

lib/jobs/isc-dhcp-lease-poller.js
```diff
@@ -144,6 +144,10 @@ function iscDhcpLeasePollerFactory(
...
+        var notExpired = false;
+
+       //look at cat /var/lib/dhcp/dhcpd.leases to find leases
```
benbp: I would get rid of the `cat` in this sentence.

lib/jobs/isc-dhcp-lease-poller.js
```diff
@@ -157,6 +161,18 @@ function iscDhcpLeasePollerFactory(
                 return ip;
             }
 
+            if(line.startsWith('ends')) {
+                // slice off the to just get the date and time
+                var endTime = line.slice(7, -1);
+
+                var leaseDate = new Date(endTime);
```
benbp: Another naming comment: perhaps `expirationDate`?

lib/jobs/isc-dhcp-lease-poller.js
```diff
@@ -144,6 +144,10 @@ function iscDhcpLeasePollerFactory(
         */
         var split = data.split('\n');
         var lease;
+        var todaysDate = new Date();
+        var notExpired = false;
```
benbp: This double negative is also kind of confusing, can we just do: `var expired = true`?

lib/jobs/isc-dhcp-lease-poller.js
```diff
@@ -157,6 +161,18 @@ function iscDhcpLeasePollerFactory(
...
 
+            if(line.startsWith('ends')) {
+                // slice off the to just get the date and time
+                var endTime = line.slice(7, -1);
+
+                var leaseDate = new Date(endTime);
+
+                //Check to see if it is not expired.
+                if(leaseDate > todaysDate) {
+                    notExpired = true;
```
benbp: Actually, is this variable even necessary? Couldn't you just `return ip` here if the conditional is `leaseDate < todaysDate`?

spec/lib/jobs/isc-dhcp-poller-job-spec.js
```diff
@@ -75,12 +126,35 @@ describe('ISC DHCP Poller Job', function () {
...
         });
+
+        it('should not parse an expired lease', function() {
+            var job = new this.Jobclass({}, {}, uuid.v4());
+            var parsed = job.parseLeaseData(expiredLeaseData.toString());
+
+            expect(parsed).to.be.undefined;
+        });
+
+        it('should only parse none expired leases', function() {
```
benbp: "none expired leases"?

lib/jobs/isc-dhcp-lease-poller.js
tldavies: Sure

lib/jobs/isc-dhcp-lease-poller.js
tldavies: have about expiredtime as you didnt seem to like date before?

lib/jobs/isc-dhcp-lease-poller.js
tldavies: if you return the ip here it will still go to the next line. Which will add the lease.

benbp:
:+1:

DavidjohnBlodgett:
:+1:

PR #112, opened 2016-01-28
ON-2719 Update obm control constructor to support sourcing a node

jimturnquist:
ON-2719 Update obm control constructor to support sourcing a node via its options instead of the graph context.

@benbp

### DIFF
--- a/lib/jobs/obm-control.js
+++ b/lib/jobs/obm-control.js
@@ -38,7 +38,7 @@ function obmControlJobFactory(
      * @param {String} taskId
      * @constructor
      */
-   var ObmControlJob = function ObmControlJob(options, context, taskId) {
+    var ObmControlJob = function ObmControlJob(options, context, taskId) {
         ObmControlJob.super_.call(this, logger, options, context, taskId);
 
         assert.string(this.options.action, 'action');
--- a/lib/jobs/obm-control.js
+++ b/lib/jobs/obm-control.js
@@ -38,17 +38,17 @@ function obmControlJobFactory(
      * @param {String} taskId
      * @constructor
      */
-    var ObmControlJob = function ObmControlJob(options, context, taskId) {
+   var ObmControlJob = function ObmControlJob(options, context, taskId) {
         ObmControlJob.super_.call(this, logger, options, context, taskId);
 
-        assert.string(this.context.target);
-        assert.string(this.options.action);
+        assert.string(this.options.action, 'action');
         assert.ok(_.contains(_.methods(ObmService.prototype), this.options.action),
                 'OBM action is a known action');
         // Defaults for this should have been set by the definition
-        assert.ok(this.options.obmServiceName);
+        assert.ok(this.options.obmServiceName, 'obm service name');
 
-        this.nodeId = this.context.target;
+        this.nodeId = this.context.target || this.options.nodeId;
+        assert.string(this.nodeId, 'nodeId');
         this.obmServiceName = this.options.obmServiceName;
     };
     util.inherits(ObmControlJob, BaseJob);
--- a/spec/lib/jobs/obm-control-spec.js
+++ b/spec/lib/jobs/obm-control-spec.js
@@ -183,7 +183,42 @@ describe("Job.Obm.Node", function () {
             });
         });
 
-        it('should fail to run OBM command if no target or nodeId was specified', function() {
+        it('should create new Job with node selected from target (when node specified in both target and options)', function() {
+            // local options with nodeId set
+            var options = {
+                action: 'reboot',
+                obmServiceName: 'ipmi-obm-service',
+                nodeId: 'not_this_one'
+            };
+
+            var newJob = new Job(options, { target: 'pick_me' }, uuid.v4());
+            expect(newJob.nodeId).to.equal('pick_me');
+        });
+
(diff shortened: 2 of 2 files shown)


lib/jobs/obm-control.js
```diff
@@ -38,17 +38,17 @@ function obmControlJobFactory(
      * @param {String} taskId
      * @constructor
      */
-    var ObmControlJob = function ObmControlJob(options, context, taskId) {
+   var ObmControlJob = function ObmControlJob(options, context, taskId) {
```
heckj: total nit, but slight indent error here. Not that javascript cares, just looks funny

lib/jobs/obm-control.js
```diff
@@ -38,17 +38,17 @@ function obmControlJobFactory(
...
-        assert.string(this.options.action);
+        assert.string(this.options.action, 'action');
         assert.ok(_.contains(_.methods(ObmService.prototype), this.options.action),
                 'OBM action is a known action');
         // Defaults for this should have been set by the definition
-        assert.ok(this.options.obmServiceName);
+        assert.ok(this.options.obmServiceName, 'obm service name');
 
-        this.nodeId = this.context.target;
+        this.nodeId = this.context.target || this.options.nodeId;
```
heckj: could we add a quick unit test validating this logic? As I'm reading it, the Job will use target by default, and override specifically with `options.nodeId` if it's there. Looks like it would be a quick add to `spec/lib/jobs/obm-control-spec.js`

lib/jobs/obm-control.js
jimturnquist: I will tweak.  Thanks.

lib/jobs/obm-control.js
jimturnquist: Thanks Joe.  Looking now...

benbp: Once again, hit me up on slack if you need unit test pointers!

benbp:
Ignore the hound errors here, we still have to copy the linter fixes made to on-core into the other repos.

benbp:
:+1:

jimturnquist:
@heckj Tests have been added for the cases you mentioned.

heckj:
Sorry, thanks for the bump :+1:

spec/lib/jobs/obm-control-spec.js
```diff
@@ -151,6 +182,19 @@ describe("Job.Obm.Node", function () {
...
+
+            // local options with no nodeId field
+            var options = {
+                action: 'reboot',
+                obmServiceName: 'ipmi-obm-service'
+            };
+
+            expect(function() {
+                return new Job(options, {}, uuid.v4());
+            }).to.throw('nodeId (string) is required');
```
yyscamper: It would be better to expect an `AssertionError` not the specified error message, like `expect(xxx).to.throw(AssertionError)`, so this leave the flexibility to change error message without modifying the unit-test.

spec/lib/jobs/obm-control-spec.js
```diff
@@ -127,6 +128,36 @@ describe("Job.Obm.Node", function () {
...
+                            "user": "admin",
+                            "password": "admin",
+                            "host": "10.0.0.254"
+                        }
+                    }
+                ]
+            };
+            mockWaterline.nodes.findByIdentifier.resolves(node);
+
+            options.nodeId = job.context.target;
```
yyscamper: Since you pass the `context.target` to `options.nodeId`, so you cannot make sure the calledWith argument `nodeId` is from options or context.

I would expect there will be 4 test cases here:
- options.nodeId is empty and context.target is valid, so context.target will be used.
- options.nodeId is valid and the context.target is empty, so options.nodeId will be used.
- both options.nodeId and context.target are valid, options.nodeId will take precedence.
- both options.nodeId and context.target are empty, there will be Error threw out.

yyscamper:
Since houdci false alarm fix should be added into mater branch, so the we could still see those false houndci alarm in if the PR is for 1.1.0 branch

spec/lib/jobs/obm-control-spec.js
jimturnquist: Thanks.  I'm not sure that looking for any AssertionError is the right approach.  Modifying the unit test is not necessarily a bad thing if the behavior has actually changed.  Expecting AssertionError is too generic and would not actually verify that the specific condition that we are looking for.

spec/lib/jobs/obm-control-spec.js
jimturnquist: Good suggestions.  Doing that now...

spec/lib/jobs/obm-control-spec.js
benbp: Luckily, we can have the best of both worlds. When I want to preserve error message flexibility I usually just do a regex to make sure it's at least *probably* the right error message:

```
expect(xxx).to.throw(AssertionError, /nodeId/)
```

benbp: Seem okay to you guys @yyscamper @jimturnquist ?

spec/lib/jobs/obm-control-spec.js
jimturnquist: @yyscamper Just added a change that creates your suggested tests.

PR #114, opened 2016-01-29
Use acceptedResponseCodes instead of "|| true" in bmc catalog

iceiilin:
Fix ODR-339 (https://hwjiraprd01.corp.emc.com/browse/ODR-399) reporting no poller for management server.

The missing of poller is caused by missing of "bmc" source in catalogs, which is the response from "ipmitool lan print" command. In the catalog-mgmt-bmc task, the command is "ipmitool lan print || true", which is sent to a child process as parameters to ipmitool. the "|| true" is invalid parameters and caused the stderr in the response.

The solution is to remove "|| true" from command, and use acceptedResponseCodes to check whether it is an acceptable return code. This change applies to bmc catalog tasks in both mgmt and compute node.

The functionality depends on https://github.com/RackHD/on-core/pull/70 about the modification for acceptable return code in child process.

### DIFF
--- a/lib/task-data/tasks/catalog-bmc.js
+++ b/lib/task-data/tasks/catalog-bmc.js
@@ -8,57 +8,191 @@ module.exports = {
     implementsTask: 'Task.Base.Linux.Catalog',
     options: {
         commands: [
-            'sudo ipmitool lan print || true',
+            {
+                command: 'sudo ipmitool lan print',
+                acceptedResponseCodes: [1]
+            },
             'sudo ipmitool sel',
             'sudo ipmitool sel list -c',
             'sudo ipmitool mc info',
-            'sudo ipmitool user summary 1 || true',
-            'sudo ipmitool -c user list 1 || true',
+            {
+                command: 'sudo ipmitool user summary 1',
+                acceptedResponseCodes: [1]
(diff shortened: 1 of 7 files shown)


lib/task-data/tasks/catalog-bmc.js
```diff
@@ -8,57 +8,191 @@ module.exports = {
     implementsTask: 'Task.Base.Linux.Catalog',
     options: {
         commands: [
-            'sudo ipmitool lan print || true',
+            {
```
benbp: Bummer, I guess we can't take advantage of mustache rendering loops because we have to represent these as objects, huh?

benbp:
:+1:

yyscamper:
:+1:

lib/task-data/tasks/catalog-bmc.js
iceiilin: Interesting suggestion. I will try.

iceiilin:
@benbp @yyscamper I read through mustache specs (https://github.com/mustache/spec/tree/master/specs), and it doesn't support object rendering:(. Please help to merge this PR along with https://github.com/RackHD/on-core/pull/70

PR #109, opened 2016-01-18
Revert "Update kernel version of ubuntu microkernel to 3.16.0-25-generic."

heckj:
Reverts RackHD/on-tasks#108

Per @benbp and @stuart-stanley reverting this PR - to be re-introduced later so as to avoid some internal EMC logistical issues.

PR #115, opened 2016-01-29
Revert "Revert "Update kernel version of ubuntu microkernel to 3.16.0-25-generic.""

benbp:
Reverts RackHD/on-tasks#109

Looks like we're ready to have this back! (Context: we are bumping the kernel version to support UEFI booting, the appropriate images are already built and hosted in bintray, this is the remaining piece to utilize them during discovery).

@stuart-stanley @heckj @amymullins @davequick @RackHD/corecommitters


benbp:
For posterity: We can't go straight to the 3.19 kernel yet, because we make use of the overlayfs driver, which has been changed between 3.16 and 3.19 to be called the "overlay" driver instead, with some different mount requirements/arguments as well.

stuart-stanley:
+1 for me. What could possibly go wrong?  **diving for cover**

jlongever:
+1

PR #117, opened 2016-02-03
Add valid ip to lookups when discovery regardless of its source

iceiilin:
Currently there are two cases in RackHD that the BMC MAC-IP relations will be put into lookups:
1.	BMC IP is DHCP and in the subnet of eth1
2.	BMC IP is static

There is also the case where BMC is DHCP in subnet of ORA eth0 with a valid IP address (the IP address is assigned by external DHCP server in eth0).

This PR covers the above case, removing check of "DHCP" source in bmc catalogs when adding lookup entry. It will not cause duplicate in lookups, since lookup.js in on-core has covered this situation.

### DIFF
--- a/lib/utils/job-utils/command-parser.js
+++ b/lib/utils/job-utils/command-parser.js
@@ -219,10 +219,10 @@ function commandParserFactory(Logger, Promise, _) {
                 var store = (parsed['MAC Address'] !== '00:00:00:00:00:00');
 
                 // If the BMC IP is statically configured at time of discovery,
+                // or assigned by external dhcp-server,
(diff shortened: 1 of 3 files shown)


benbp:
:+1: though I was under the impression that you couldn't get IP information out of the lan catalog if the BMC was set to DHCP, and that it would always be '0.0.0.0'?

iceiilin:
@benbp as we discussed in the email, BMC is in the same subnet as eth0 of ORA, and an external dhcp server assigned an IP to it.

yyscamper:
:+1:

iceiilin:
test this please

PR #119, opened 2016-02-04
Modify the regex about finding serial number when generating enclosure

iceiilin:
Modify the regex about finding serial number when generating enclosure.

This PR covers the case where serial number in catalogs of mgmt server is from ESXi and is not valid to generate an enclosure accordingly.

Serial number is from two sources: ipmi-fru and dmi. When ipmitool is faulty, dmi from ESXi would be the only candidate with serial number like "VMware-42 2a 27 06 e1 68 c5 44-9b 2c aa 7d 3a 53 99 **64**". Previous code would get the serial number as "**64**" (words from the end of the string).
In this PR, only a string, all of which are words, can be viewed as a valid serial number.

### DIFF
--- a/lib/jobs/generate-enclosure.js
+++ b/lib/jobs/generate-enclosure.js
@@ -173,7 +173,7 @@ function generateEnclosureJobFactory(
             source: enclPath.src
         }).then(function (catalog) {
             var nodeSn = catalogSearch.getPath(catalog.data, enclPath.entry);
-            var regex = /[\w]+$/;
+            var regex = /^[\w]+$/;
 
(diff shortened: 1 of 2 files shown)


benbp:
Ah thank you. I had noticed that discovery of VirtualBox VMs created enclosure nodes too, perhaps this is why.  :+1:

yyscamper:
:+1:

iceiilin:
Please help to merge. @benbp @yyscamper

heckj:
nice @iceiilin, thank you!

:+1:

PR #120, opened 2016-02-08
Fix lshw catalog parsing and lookup generation with multiple nics

benbp:
In certain node catalogs with multiple nics present, the pci child objects have an id value of "network:N" instead of "network". This is a fix to our parsing to capture these values. The consequences of failing to parse these values is that we weren't generating lookup records for multiple NIC mac addresses for a node, and other workflows that depend on those records being there because of inconsistent boot ordering, like the CentOS installer, were failing.

### DIFF
--- a/lib/utils/job-utils/command-parser.js
+++ b/lib/utils/job-utils/command-parser.js
@@ -548,12 +548,12 @@ function commandParserFactory(Logger, Promise, _) {
                 return child.id.startsWith('pci');
             });
             var macs = _.compact(_.map(pci.children, function(child) {
-                if (child.id === 'network') {
+                if (child.id && child.id.startsWith('network')) {
                     return child.serial;
                 }
-                if (child.id.startsWith('pci')) {
+                if (child.id && child.id.startsWith('pci')) {
                     return _.compact(_.map(child.children, function(_child) {
-                        if (_child.class === 'network') {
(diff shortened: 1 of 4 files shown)


benbp:
@RackHD/corecommitters @johren @stuart-stanley @pscharla @heckj

stuart-stanley:
+1 with the caveat being that I'm not steeped in nodejs enough yet to be a good sanity checker there. I can see how what is being fix would cause.... interesting.... intermittent issues though.

benbp:
@stuart-stanley in the case of the bug that spawned this root cause investigation, the CentOS installer job listens for an AMQP event from our HTTP server that says "hey this node downloaded the renasar-ansible.pub" file." However, since we failed to generate a lookup record correlating one of the mac addresses to the node ID during discovery, the HTTP server publishes the AMQP event with a routing key that _does not_ include the node ID, hence, the workflow never completes because it is listening for the message from HTTP on a routing key that _does_ include the node ID.

jlongever:
+1

benbp:
@johren @pscharla looks like this is ready to go. I know there are a lot of smoke test failures so I won't merge, but any of us CCs know when it's ready and we'll do it.

johren:
@RackHD/corecommitters Please go ahead and merge.

PR #121, opened 2016-02-08
Fix lshw catalog parsing and lookup generation with multiple nics

benbp:
In certain node catalogs with multiple nics present, the pci child objects have an id value of "network:N" instead of "network". This is a fix to our parsing to capture these values. The consequences of failing to parse these values is that we weren't generating lookup records for multiple NIC mac addresses for a node, and other workflows that depend on those records being there because of inconsistent boot ordering, like the CentOS installer, were failing.

### DIFF
--- a/spec/lib/utils/job-utils/command-parser-spec.js
+++ b/spec/lib/utils/job-utils/command-parser-spec.js
@@ -296,8 +296,35 @@ describe("Task Parser", function () {
                         { mac: '00:1e:67:69:4c:b8' }
                     ]
                 );
-                expect(_.isEqual(result.data,
-                    JSON.parse(stdoutMocks.lshwOutput))).to.be.true;
+                expect(JSON.parse(stdoutMocks.lshwOutput)).to.deep.equal(result.data);
+                expect(result.source).to.equal('lshw');
+            });
+        });
+
+        it("should parse lshw -json when there are multiple nics ", function () {
+            // Test cases where pci network string resembles "network:0", "network:1", etc.
+            // rather than just "network".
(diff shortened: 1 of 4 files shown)


benbp:
@RackHD/corecommitters @johren @stuart-stanley @pscharla @heckj

jlongever:
+1

VulpesArtificem:
+1

spec/lib/utils/job-utils/command-parser-spec.js
```diff
@@ -302,6 +302,35 @@ describe("Task Parser", function () {
...
+                expect(result.error).to.be.undefined;
+                expect(result.store).to.be.true;
+                expect(result.lookups).to.deep.equal(
+                    [
+                        { mac: 'f8:bc:12:0b:0b:40' },
+                        { mac: 'f8:bc:12:0b:0b:41' }
+                    ]
+                );
+                expect(_.isEqual(result.data,
+                    JSON.parse(stdoutMocks.lshwOutputMultiNic))).to.be.true;
```
yyscamper: Isn't this style better?
```
expect(JSON.parse(xxx)).to.deep.equal(result.data)
```

yyscamper:
:+1: though a minor style suggestion.

spec/lib/utils/job-utils/command-parser-spec.js
benbp: Yep, I'll fix it. I copy/pasted the above test, which was written before we knew the Chai framework very well, haha :)

benbp:
Updated the style @yyscamper

PR #122, opened 2016-02-15
ODR-547 use findOrCreate instead of create when creating default pollers

jimturnquist:
ODR-547 use findOrCreate instead of create when creating default pollers

Updated spec tests to account for switch from create to findOrCreate.

### DIFF
--- a/lib/jobs/create-default-pollers.js
+++ b/lib/jobs/create-default-pollers.js
@@ -71,7 +71,8 @@ function createDefaultPollersJobFactory(
             }).then(function (catalog) {
                 if (catalog) {
                     poller.node = self.nodeId;
-                    return waterline.workitems.create(poller);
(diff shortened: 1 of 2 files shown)


heckj:
hound errors re: 'expect and sinon' are ignorable...

heckj:
:+1:

heckj:
Note for other reviewers: this effectively makes the createDefaultPollers task idempotent so that it can safely be called on a "rediscovery" workflows to refresh the catalogs for a node.

yyscamper:
:+1:

johren:
@RackHD/corecommitters

jlongever:
Can ignore ^^ jenkins error... SDR format changed in master,  no smoke tests for 1.1.0 right now.
+1

PR #124, opened 2016-02-17
Add tasks to run the node tagging job.

zyoung51:
@RackHD/corecommitters @uppalk1 @keedya @tannoa2 

Depends on https://github.com/RackHD/on-core/pull/88

### DIFF
--- a/lib/jobs/generate-tag.js
+++ b/lib/jobs/generate-tag.js
@@ -1,0 +1,119 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = generateTagJobFactory;
+di.annotate(generateTagJobFactory, new di.Provide('Job.Catalog.GenerateTag'));
+di.annotate(generateTagJobFactory, new di.Inject(
+    'Job.Base',
+    'Services.Waterline',
+    'Protocol.Events',
+    'JobUtils.CatalogSearchHelpers',
+    'anchor',
+    'Logger',
+    'Util',
+    'Promise',
+    'Assert',
+    '_'
+));
+function generateTagJobFactory(
+    BaseJob,
+    waterline,
+    eventsProtocol,
+    catalogSearch,
+    anchor,
+    Logger,
+    util,
+    Promise,
+    assert,
+    _
+) {
+
+    var logger = Logger.initialize(generateTagJobFactory);
+
+    /**
+     *
+     * @param {Object} [options]
+     * @constructor
+     */
+    function GenerateTagJob(options, context, taskId) {
+        GenerateTagJob.super_.call(this, logger, options, context, taskId);
+
+        this.nodeId = context.target || options.nodeId;
+        assert.isMongoId(this.nodeId, 'context.target || options.nodeId');
+    }
+
+    util.inherits(GenerateTagJob, BaseJob);
+
+    var RULE_DELIMITER = /\./g;
+
+    /**
+     * @memberOf GenerateTagJob
+     * @returns {Promise}
+     */
+    GenerateTagJob.prototype._run = function run() {
+        var self = this;
+
+        waterline.tags.find({}).then(function (tags) {
+            var catalogTypes = _(tags)
+                .pluck('rules')
+                .flattenDeep()
+                .pluck('path')
+                .map(function (path) {
+                    return _.compact(path.split(RULE_DELIMITER))[0];
+                })
+                .uniq()
+                .value();
+            return [ Promise.map(catalogTypes, function(type) {
+                return waterline.catalogs.findMostRecent({
+                    node: self.nodeId,
+                    source: type
+                });
+            }), tags ];
+        }).spread(function(catalogs, tags) {
+            catalogs = _(catalogs)
+                .flattenDeep()
+                .compact()
+                .transform(function (catalogs, catalog) {
+                    catalogs[catalog.source] = catalog.data;
+                }, {})
+                .value();
+
+            var matches = _.filter( matchTags(catalogs, tags), function(match) {
+                return match.tag.rules.length && !match.errors.length;
+            });
+
+            var tagNames = _(matches).pluck('tag.name').value();
+            if( !_.isEmpty(tagNames)) {
+                return waterline.nodes.addTags(self.nodeId, tagNames);
+            }
+        }).then(function() {
+            self._done();
+        }).catch(function(err) {
+            self._done(err);
+        });
+    };
+
+    function matchTags(catalogs, tags) {
+        return _.map(tags, function (tag) {
+            var result = _.reduce(tag.rules, function (result, rule) {
+                var path = _.compact(rule.path.split(RULE_DELIMITER));
+                var depth = path.length;
+                var value = catalogSearch.getPath(catalogs, path.join('.'));
+                return {
+                    maxDepth: Math.max(depth, result.maxDepth),
+                    errors: result.errors.concat(anchor(value).to(_.omit(rule, 'path')) || []),
+                };
+            }, {
+                maxDepth: 0,
+                errors: []
+            });
+            result.tag = tag;
+            return result;
+        });
+    }
+
+    return GenerateTagJob;
+}
--- a/lib/task-data/base-tasks/generate-tag.js
+++ b/lib/task-data/base-tasks/generate-tag.js
@@ -1,0 +1,15 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Generate Tag',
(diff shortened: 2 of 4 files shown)


lib/jobs/generate-tag.js
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @memberOf GenerateTagJob
+     * @returns {Promise}
+     */
+    GenerateTagJob.prototype._run = function run() {
+        var self = this;
+
+        waterline.tags.find({}).then(function (tags) {
+            var catalogTypes = _(tags)
```
benbp: I just discovered this usage pattern of lodash a couple weeks ago, glad you are using it.

zyoung51: Heavily leveraged: https://github.com/RackHD/on-tasks/blob/master/lib/jobs/generate-sku.js#L59

benbp: Ah, well, hat tip to @halfspiral then :)

lib/jobs/generate-tag.js
```diff
@@ -0,0 +1,119 @@
...
+                return waterline.nodes.addTags(tagNames);
+            }
+        }).then(function() {
+            self._done();
+        }).catch(function(err) {
+            self._done(err);
+        });
+    };
+
+    function matchTags(catalogs, tags) {
```
benbp: One minor problem I'm anticipating, not with your code but with the existing system, is that we have to sanitize catalog key values with '.' characters in them to be '_' in order to be mongo compatible JSON (the ipmi sdr catalog as an example of this). I'm not sure we're very judicious about deserializing this data to restore the '.' characters.

spec/lib/jobs/generate-tag-spec.js
```diff
@@ -0,0 +1,243 @@
...
+    var GenerateTag;
+    var uuid;
+
+    var catalog1 = {
+        id: 'abf9d5a4ebf61b6715dbaef6',
+        source: 'dmi',
+        data: {
+            dmi: {
+                system: {
+                    manufacturer: 'Renasar',
```
benbp: :P

spec/lib/jobs/generate-tag-spec.js
```diff
@@ -0,0 +1,243 @@
...
+        waterline.catalogs.findMostRecent.onCall(1).resolves(catalog2);
+
+        return job.run()
+        .then(function() {
+            expect(waterline.nodes.addTags).to.have.not.been.called;
+        });
+    });
+
+
+    it('assigns nothing when no catalogs present', function() {
```
benbp: Since you are doing this type of test, might be good to throw in an 'assigns nothing when no tags present' test as well?

spec/lib/jobs/generate-tag-spec.js
```diff
@@ -0,0 +1,243 @@
...
+    beforeEach(function () {
+        waterline.tags.find.reset();
+        waterline.catalogs.findMostRecent.reset();
+        waterline.nodes.addTags.reset();
+        taskProtocol.subscribeActiveTaskExists = sinon.stub().resolves({
+            dispose: sinon.stub()
+        });
+    });
+
+    it('assigns a matching tag', function() {
```
benbp: Humor me, but if you have the time would you add a simple test that does a regex match rather than an equals match? Just so anyone looking for tag definition examples can see one of that usage.

benbp:
+0, a couple of minor tests that I think might be useful in the long run, but if you spend more than 10 minutes on them I'll feel bad :)

lib/task-data/base-tasks/generate-tag.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Generate Tag',
+    injectableName: 'Task.Base.Catalog.GenerateTag',
+    runJob: 'Job.Catalog.GenerateTag',
+    requiredOptions: [],
```
heckj: should we assert `nodeId` as a required option here?

spec/lib/jobs/generate-tag-spec.js
```diff
@@ -0,0 +1,256 @@
...
+        waterline.catalogs.findMostRecent.resolves(catalog1);
+
+        return job.run()
+        .then(function() {
+            expect(waterline.nodes.addTags).to.have.been.calledOnce;
+            expect(waterline.nodes.addTags).to.have.been.calledWith(['Test Tag', 'Test Tag 2']);
+        });
+    });
+
+    it('assigns a matching tag against multiple catalogs', function() {
```
heckj: just verifying, this is effectively "assign the tag if both these rules match" - and *`AND`* logical setup? Is there an *`OR`*? mechanism

zyoung51: This is validating that when multiple tag rules match all of them are added to the node.  This is different from the SKU logic which will only applies the best matched sku to the node.

lib/jobs/generate-tag.js
```diff
@@ -0,0 +1,119 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: Still the copyright problem

lib/jobs/generate-tag.js
yyscamper: I like this chain style, seems very cool :)

lib/jobs/generate-tag.js
yyscamper: Image a float value or string like `1.2` in catalog, changing the `.` to `_` will lost its meaning. Of course, it should a rare case to use float string as key.

An alternative way is copying the `%` output in C language, we can use double `.` in rule path to explicitly tells it belongs to the key itself, don't consider it as delimiter.

yyscamper: Meanwhile, I prefer to use some standard way to retrieve catalog in future, such as [JSONPath](https://github.com/jayway/JsonPath) or [XPath](https://en.wikipedia.org/wiki/XPath) or [JPath](http://bluelinecity.com/software/jpath/)
Waterline or Mongodb has its own path syntax?

lib/task-data/base-tasks/generate-tag.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Generate Tag',
+    injectableName: 'Task.Base.Catalog.GenerateTag',
```
yyscamper: GenerateTag should not be a standard catalog job (For me, Standard Catalog Job means executing catalog command, parse the output and store it to database), so I am wondering whether it is suitable to set the injectableName to be nested under `Catalog`.

@RackHD/corecommitters may have other thought on this.

yyscamper:
:+1: after fix the copyright and the clarify the injectableName naming.

lib/task-data/base-tasks/generate-tag.js
zyoung51: @yyscamper: The SKU command, which is exactly the same, is called Task.Base.Catalog.GenerateSKU.  I believe they are named this way because they are initiated during discovery cataloging.

spec/lib/jobs/generate-tag-spec.js
zyoung51: np, I'll add one

spec/lib/jobs/generate-tag-spec.js
zyoung51: np, I'll add one.

lib/jobs/generate-tag.js
benbp: @yyscamper yeah I agree I think formalizing on something like JSONPath is a good idea to start work on. Mongodb doesn't support `.` characters within keys because the mongo query language allows for nested key searches, for example:

```
find({
    'key.subkey.subsubkey': <value>
})
```

would find a document like:

```
{
    "key": {
        "subkey": {
            "subsubkey": <value>
        }
    }
}
```

lib/task-data/base-tasks/generate-tag.js
benbp: @zyoung51 @yyscamper many of our injector names (especially older ones) have no strong reasoning behind them, and may not actually make sense anymore :). I think we initially nested it under `Catalog` because the job relied upon and searched through the catalogs.

benbp:
:+1:

lib/task-data/base-tasks/generate-tag.js
yyscamper: Thanks @benbp , OK for me.

yyscamper:
:+1: after fix the newly appeared  "Line is too long" warning.

jlongever:
+1

PR #127, opened 2016-02-20
Add task and job for setting bios with racadm tool

sunnyqianzhang:
Add task and job for setting bios with racadm tool


lib/jobs/dell-racadm-set-bios-job.js
```diff
@@ -0,0 +1,111 @@
...
+                return racadmTool.setBiosConfig(obmSetting.host, obmSetting.user,
+                    obmSetting.password,
+                    self.cifsInfo);
+            })
+            .then(function(results){
+                assert.equal(results.jobStatus.status,"Completed",results.jobStatus.status);
+                self._done();
+            })
+            .catch(function(err) {
+                logger.error('The Error Is Found Here!', {error: err});
```
yyscamper: This error message "The Error Is Found Here" is not good, it tells nothing to debug

lib/jobs/dell-racadm-set-bios-job.js
```diff
@@ -0,0 +1,111 @@
...
+            })
+            .then(self.revealSecrets)
+            .then(self.lookupHost)
+            .then(function(obmSetting) {
+                return racadmTool.setBiosConfig(obmSetting.host, obmSetting.user,
+                    obmSetting.password,
+                    self.cifsInfo);
+            })
+            .then(function(results){
+                assert.equal(results.jobStatus.status,"Completed",results.jobStatus.status);
```
yyscamper: If the jobStatus.status is not "completed", isn't the racadmTool.setBiosConfig will be rejected? I think this kind of judgement should be better wrapped in racadmTool.setBiosConfig.

lib/jobs/dell-racadm-set-bios-job.js
```diff
@@ -0,0 +1,111 @@
...
+     * @constructor
+     */
+    function RacadmSetBiosJob(options, context, taskId) {
+        RacadmSetBiosJob.super_.call(this, logger, options, context, taskId);
+
+        this.nodeId = this.context.target;
+        this.userName= options.username;
+        this.passWord = options.password;
+        this.filePath = options.filePath;
+        this.cifsInfo = {user: this.userName, password: this.passWord , filePath: this.filePath};
```
yyscamper: I see `this.userNmae`, `this.passWord` and `this.filePath` is only used in the assignment of `this.cifsInfo`, so it's better to remove `userName`/`password`/`filePath` from `this`, in order  to keep a clean this context.

lib/jobs/dell-racadm-set-bios-job.js
```diff
@@ -0,0 +1,111 @@
...
+     * @param {Object} context
+     * @param {String} taskId
+     * @constructor
+     */
+    function RacadmSetBiosJob(options, context, taskId) {
+        RacadmSetBiosJob.super_.call(this, logger, options, context, taskId);
+
+        this.nodeId = this.context.target;
+        this.userName= options.username;
+        this.passWord = options.password;
```
yyscamper: It's better to use `password` not `passWord`, because for security reason, the password will be redact so that the plain password will not be printed in log file. But `passWord` will not be redact by design.

lib/task-data/tasks/dell-racadm-set-bios.js
```diff
@@ -0,0 +1,15 @@
...
+    friendlyName: 'dell racadm set bios',
+    injectableName: 'Task.Dell.Racadm.SetBIOS',
+    implementsTask: 'Task.Base.Dell.Racadm.SetBIOS',
+    options: {
+        username: null,
+        password: null,
+        filePath: null
+    },
+    properties: {}
+};
```
yyscamper: it's better to add a newline symbol to avoid such warning.

spec/lib/jobs/dell-racadm-set-bios-job-spec.js
```diff
@@ -0,0 +1,194 @@
...
+        it('should fail if node does not exist', function(done) {
+            mockWaterline.nodes.findByIdentifier.resolves(null);
+            job.run()
+                .then(function() {
+                    done(new Error("Expected job to fail"));
+                })
+                .catch(function(e) {
+                    try {
+                        expect(e).to.have.property('name').that.equals('AssertionError');
+                        expect(e).to.have.property('message').that.equals('No node for dell' +
```
yyscamper: Since `e` is an instance of Error, so it's not necessary to check its property `name` and `message`, so to check its Error type and message is enough.
Meanwhile, to expect a promise to rejected, it can be as simple as following:
```
return expect(job.run()).to.be.rejectedWith(AssertionError, 'No node for dell racadm set bios');
```

spec/lib/jobs/dell-racadm-set-bios-job-spec.js
```diff
@@ -0,0 +1,194 @@
...
+            mockWaterline.nodes.findByIdentifier.resolves(node);
+            job.run()
+                .then(function() {
+                    done(new Error("Expected job to fail"));
+                })
+                .catch(function(e) {
+                    try {
+                        expect(e).to.have.property('name').that.equals('AssertionError');
+                        expect(e).to.have.property('message').that.equals('No ipmi obmSetting for' +
+                            ' dell racadm set bios');
```
yyscamper: same comment as above.

spec/lib/jobs/dell-racadm-set-bios-job-spec.js
```diff
@@ -0,0 +1,194 @@
...
+                '10.1.1.2'
+            );
+            var data = {
+                host: '7a:c0:7a:c0:be:ef'
+            };
+            return job.lookupHost(data)
+                .then(function(result){
+                    expect(result).to.deep.equal({
+                        host: '10.1.1.2'
+                    });
```
yyscamper: There is a simpler writing for this:
```javascript
return expect(job.lookupHost(data).to.become({ host: '10.1.1.2'});
```
or 
```javascript
return expect(job.lookupHost(data).to.be.eventually.deep.equal({ host: '10.1.1.2'});
```

spec/lib/task-data/base-tasks/dell-racadm-set-bios-base-spec.js
```diff
@@ -0,0 +1,18 @@
...
+/* jshint node:true */
+
+'use strict';
+
+describe(require('path').basename(__filename), function () {
+    var base = require('./base-task-data-spec');
+
+    base.before(function (context) {
+        context.taskdefinition = helper.require('/lib/task-data/base-tasks/dell-racadm-set-' +
+            'bios-base.js');
```
yyscamper: Just style suggestion.
This kind of line break is a little not so good for me :), if it is just because the line is too long, you can move the total string to next line, like:
```javascript
context.taskdefinition = helper.require(
    '/lib/task-data/base-tasks/dell-racadm-set-bios-base.js'
);
```

spec/lib/jobs/dell-racadm-set-bios-job-spec.js
```diff
@@ -0,0 +1,194 @@
...
+                {
+                    jobStatus:{
+                        status:"Completed"
+                    }
+                }
+            );
+            return job.run()
+                .then(function() {
+                    expect(setBiosStub).to.have.been.called;
+                    expect(setBiosStub).to.have.been.calledWith(
```
yyscamper: Since you have expected the stub function to be called with some argument, so it's not necessary to expected it is called.

anhou:
:+1:  after fix Felix's comments

lib/task-data/tasks/dell-racadm-set-bios.js
sunnyqianzhang: Fixed

sunnyqianzhang: Ok

lib/jobs/dell-racadm-set-bios-job.js
```diff
@@ -0,0 +1,109 @@
...
+            })
+            .then(self.revealSecrets)
+            .then(self.lookupHost)
+            .then(function(obmSetting) {
+                return racadmTool.setBiosConfig(obmSetting.host, obmSetting.user,
+                    obmSetting.password,
+                    self.cifsInfo);
+            })
+            .then(function(results){
+                logger.log(results.status);
```
iceiilin: logger doesn't have the property "log", causing travis CI and jenkins to fail. You can change it to logger.debug().

yyscamper:
:+1:

PR #128, opened 2016-02-21
Add redfish discovery workflow and OBM service

jlongever:
- Create unbound redfish discovery workflow
- Create redfish chassis/system nodes with redfish target relations
- Create redfish OBM service for redfish reset actions
- Added support to pass task parameters into OBM service constructor
- Add/Update unit-tests

Dependent: 
https://github.com/RackHD/on-taskgraph/pull/64
https://github.com/RackHD/on-http/pull/142

@RackHD/corecommitters @zyoung51 @tannoa2 @uppalk1 @keedya

### DIFF
--- a/lib/jobs/redfish-discovery.js
+++ b/lib/jobs/redfish-discovery.js
@@ -1,0 +1,212 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+var redfish = require('redfish-node');
+
+module.exports = RedfishDiscoveryJobFactory;
+di.annotate(RedfishDiscoveryJobFactory, new di.Provide('Job.Redfish.Discovery'));
+di.annotate(RedfishDiscoveryJobFactory,
+    new di.Inject(
+        'Job.Base',
+        'Logger',
+        'Promise',
+        'Assert',
+        'Util',
+        'Services.Waterline',
+        'Services.Encryption',
+        '_'
+    )
+);
+function RedfishDiscoveryJobFactory(
+    BaseJob,
+    Logger,
+    Promise,
+    assert,
+    util,
+    waterline,
+    encryption,
+    _
+) {
+    var logger = Logger.initialize(RedfishDiscoveryJobFactory);
+
+    /**
+     * @param {Object} options task options object
+     * @param {Object} context graph context object
+     * @param {String} taskId running task identifier
+     * @constructor
+     */
+    function RedfishDiscoveryJob(options, context, taskId) {
+        RedfishDiscoveryJob.super_.call(this,
+                                   logger,
+                                   options,
+                                   context,
+                                   taskId);
+
+        assert.object(this.options);
+        assert.string(this.options.uri);
+        this.settings = {
+            uri: this.options.uri,
+            username: this.options.username,
+            password: this.options.password
+        };
+        
+        var apiClient = new redfish.ApiClient();
+        apiClient.basePath = this.settings.uri.replace(/\/+$/, '');
+        
+        // setup basic authorization
+        if (!_.isUndefined(this.settings.username) && 
+            !_.isUndefined(this.settings.password)) {
+            var token = new Buffer(
+                this.settings.username + ':' + this.settings.password
+            ).toString('base64');
+            apiClient.defaultHeaders.Authorization = 'Basic ' + token;
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"; // (jl) need a ssl_verify option here.
+        }
+        this.redfishApi = Promise.promisifyAll(new redfish.RedfishvApi(apiClient));
+    }
+    util.inherits(RedfishDiscoveryJob, BaseJob);
+
+    /**
+     * @memberOf RedfishDiscoveryJob
+     */
+    RedfishDiscoveryJob.prototype._run = function() {
+        var self = this;
+        return Promise.resolve()
+        .then(function() {
+            return self.createChassis();
+        })
+        .then(function() {
+            return self.createSystems();
+        })
+        .catch(function(err) {
+            throw err;
+        });
+    };
+    
+    /**
+     * @function createChassis
+     * @description initiate redfish chassis discovery
+     */
+    RedfishDiscoveryJob.prototype.createChassis = function () {
+        var self = this;
+        return self.redfishApi.listChassisAsync()
+        .then(function(res) {
+            assert.object(res);
+            return res[1].body.Members;
+        })
+        .map(function(member) {
+            var id = member['@odata.id']
+                .split('Chassis/')[1].split(/\/+$/)[0]; // trim slash
+                
+            assert.string(id);
+            return self.redfishApi.getChassisAsync(id)
+            .catch(function(err) {
+                throw new Error(err.response.text);
+            });
+        })
+        .map(function(chassis) {
+            chassis = chassis[1].body;
+            var systems = _.get(chassis, 'Links.ComputerSystems') ||
+                          _.get(chassis, 'links.ComputerSystems');
+            
+            if (_.isUndefined(systems)) {
+                return Promise.reject(
+                    new Error('failed to find System members for Chassis')
+                );
+            }
+                          
+            return Promise.resolve({
+                chassis: chassis, 
+                systems: systems
+            });
+        })
+        .map(function(data) {
+            assert.object(data);
+            var targetList = [];
+            
+            _.forEach(data.systems, function(sys) {
+                var target = _.get(sys, '@odata.id') ||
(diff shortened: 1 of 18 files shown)


jlongever:
test this please

jlongever:
test this please

jlongever:
test this please

zyoung51:
+1

lib/jobs/redfish-discovery.js
```diff
@@ -0,0 +1,189 @@
...
+            uri: this.options.uri,
+            username: this.options.username,
+            password: this.options.password
+        };
+        
+        var apiClient = new redfish.ApiClient();
+        apiClient.basePath = this.settings.uri.replace(/\/+$/, '');
+        
+        // setup basic authorization
+        if (!_.isUndefined(this.settings.username) &&
```
benbp: What happens if one is undefined and not the other?

lib/jobs/redfish-discovery.js
```diff
@@ -0,0 +1,189 @@
...
+        var self = this;
+        return Promise.resolve()
+        .then(function() {
+            return self.createChassis();
+        })
+        .then(function() {
+            return self.createSystems();
+        })
+        .catch(function(err) {
+            throw err;
```
benbp: This job will never finish because you don't call `self._done()`

benbp: In the catch block you should do `self._done(err)` and in the final then block you should do `self._done()`.

lib/services/redfish-obm-service.js
```diff
@@ -0,0 +1,145 @@
...
+    )
+);
+function redfishObmServiceFactory(
+    BaseObmService, 
+    Promise, 
+    waterline,
+    assert,
+    _
+) {
+    function redfishObmService(options) {
```
benbp: This needs to be capitalized since it's a constructor.

benbp: That will make the hound strict violations go away.

lib/services/redfish-obm-service.js
```diff
@@ -0,0 +1,145 @@
...
+                'expected an array of reset types');
+            if (0 > _.indexOf(resetTypes, action)) {
+                throw new Error('unsupported reset type ' + action);
+            }
+        })
+        .then(function() {
+            return Promise.resolve();
+        })
+        .catch(function(err) {
+            return Promise.reject(err);
```
benbp: Not a problem with functionality per se, but this whole section of code from the second then block is unnecessary.

lib/services/redfish-obm-service.js
```diff
@@ -0,0 +1,125 @@
...
+        });
+    }
+    
+    redfishObmService.prototype._runInternal = function (action) {
+        var self = this;
+        return self._checkAction(action)
+        .then(function() {
+            return self.redfishApi.doResetAsync(
+                self.targetId, 
+                { reset_type: action }
```
benbp: You can throw a `// jshint ignore:line` if you want to silence these.

lib/services/redfish-obm-service.js
```diff
@@ -0,0 +1,145 @@
...
+            )
+            .catch(function(err) {
+                throw new Error(err.response.text);
+            });
+        })
+        .then(function(res) {
+            assert.ok(Array.isArray(res), 'expected an array response type');
+            var taskId = res[1].body['@odata.id'];
+            assert.string(taskId);
+            return Promise.resolve(taskId);
```
benbp: You can just `return taskId` since it will get promisified if returned from within a then block.

spec/lib/jobs/redfish-discovery-spec.js
```diff
@@ -0,0 +1,162 @@
...
+        ];
+    });
+    
+    describe('redfish discovery', function() {
+        it('should successfully run job', function() { 
+            redfishApi.listChassisAsync.resolves(listChassisData);
+            redfishApi.getChassisAsync.resolves(getChassisData);
+            redfishApi.listSystemsAsync.resolves(listSystemData);
+            redfishApi.getSystemAsync.resolves(getSystemData);
+            return redfishJob._run()
```
benbp: I know it's easier to do it this way, but just FYI the way we hook into the job in the actual code is more something like:

```
redfishJob._run();
return redfishJob._deferred
.then(<...stuff...>);
```

spec/lib/services/redfish-obm-service-spec.js
```diff
@@ -0,0 +1,122 @@
...
+    }); 
+    
+    describe('redfish service initialize client', function() {  
+        var redfishService, baseObm;
+        base.before('before', servicePath, function(self) {
+            redfishService = helper.injector.get('redfish-obm-service');
+            baseObm = helper.injector.get('OBM.base');
+        });  
+        
+        it ('should with no credentials', function() {
```
benbp: "should with no credentials" isn't very descriptive :P

spec/lib/services/redfish-obm-service-spec.js
```diff
@@ -0,0 +1,122 @@
...
+            delete testOptions.config.username;
+            delete testOptions.config.password;
+            redfishService.create(testOptions);
+            expect(redfishService).to.be.ok;
```
benbp: Do you also mean to check what is returned from `create` above?

lib/jobs/redfish-discovery.js
```diff
@@ -0,0 +1,189 @@
...
+            assert.string(id);
+            return self.redfishApi.getChassisAsync(id)
+            .catch(function(err) {
+                throw new Error(err.response.text);
+            });
+        })
+        .map(function(chassis) {
+            chassis = chassis[1].body;
+            if (_.has(chassis, 'Links.ComputerSystems')) {
+                return Promise.resolve({
```
benbp: Also don't need to wrap this with a Promise.

benbp:
A couple general notes, mainly redundancies and constructor capitalization.

-1 on the absence of a call to `self._done()` in the job, since it will never finish.

benbp:
In general though, looks great!

lib/jobs/redfish-discovery.js
jlongever: Both are required,  otherwise basic auth won't be set in the header.

lib/jobs/redfish-discovery.js
jlongever: Bah,  thanks.

lib/services/redfish-obm-service.js
jlongever: Yep, being explicit here.  Can remove.

lib/services/redfish-obm-service.js
jlongever: Being explicit,  will change.

spec/lib/services/redfish-obm-service-spec.js
jlongever: "redfish service Initialize client with no credentials"  isn't descriptive?

lib/services/redfish-obm-service.js
jlongever: Actually,  I added the return Promise.resolve() because the unit test was randomly timing out waiting for done().

jlongever: Is something like: 
```
return Promise.resolve()
        .then(function() {
            return this.redfishApi.listResetTypesAsync(this.targetId)
```
acceptable?

benbp: yeah, although you'll want to reference `self` since `this` will refer to promise scope.

lib/jobs/redfish-discovery.js
benbp: Is it worth adding an error if one is defined and not the other?

spec/lib/services/redfish-obm-service-spec.js
benbp: It is, but that's not what will be printed:

"redfish service initialize client should with no credentials" is what will show up.

lib/jobs/redfish-discovery.js
jlongever: Probably,  Will add it.

jlongever:
test this please

jlongever:
@benbp updated.

anhou:
@jlongever please fix all houndci comment issues, if it could be ignored, please comment it

jlongever:
@anhou ignoring the remaining houndci comments.    Just waiting for @benbp changes to go in to clear up Jenkins failure.

jlongever:
test this please

jlongever:
@RackHD/corecommitters this good to merge?

benbp:
:+1:

PR #129, opened 2016-02-22
Check for valid parsed lease output

jlongever:
@RackHD/corecommitters

### DIFF
--- a/lib/jobs/isc-dhcp-lease-poller.js
+++ b/lib/jobs/isc-dhcp-lease-poller.js
@@ -84,14 +84,16 @@ function iscDhcpLeasePollerFactory(
 
         Promise.try(function () {
             var lease = self.parseLeaseData(data.toString());
-
-            self.queue.enqueue(
-                lookupService.setIpAddress.bind(
-                    lookupService,
-                    lease.ip,
-                    lease.mac
-                )
-            );
+            
+            if (!_.isUndefined(lease)) {
+                self.queue.enqueue(
+                    lookupService.setIpAddress.bind(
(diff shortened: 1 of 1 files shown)


lib/jobs/isc-dhcp-lease-poller.js
```diff
@@ -84,14 +84,16 @@ function iscDhcpLeasePollerFactory(
...
+            
+            if (!_.isUndefined(lease)) {
+                self.queue.enqueue(
+                    lookupService.setIpAddress.bind(
+                        lookupService,
+                        lease.ip,
+                        lease.mac
+                    )
+                );
+            }
```
heckj: do we want to log a warning here, or pass it through silently when the lease value is undefined into this call? Don't know if it's appropriate, just asking the question...

jlongever: Good question.  I'm assuming there wasn't a new lease to look up and parseLeaseData returns undefined.

benbp:
Oh thank god.

lib/jobs/isc-dhcp-lease-poller.js
benbp: @heckj I think no, there are valid stanzas that we parse that don't have a mac address in them, and the way this code was written, it detects those by having an undefined lease. The bug here is more that we try to do things with an undefined lease instead of ignoring it. The unit tests should take care of all parsing cases we care about.

benbp:
:+1:

brianparry:
:+1:

PR #130, opened 2016-02-23
Add xml parser and get BIOS configuration functions

pengz1:
Major changes is to add two utils for RACADM tools:
1. xml to JSON parser, with this paser, we can covert xml file exported from Dell server into JSON object;
2. Get BIOS configure, this util will get bios configure and store it to specified path.

### DIFF
--- a/lib/utils/job-utils/racadm-parser.js
+++ b/lib/utils/job-utils/racadm-parser.js
@@ -11,14 +11,16 @@ di.annotate(parseRacadmDataFactory,
     new di.Inject(
         'Assert',
         '_',
-        'fs'
+        'fs',
+        'Promise'
     )
 );
 
 function parseRacadmDataFactory(
     assert,
     _,
-    fs
+    fs,
+    Promise
 ) {
 
     function RacadmCommandParser() {}
--- a/lib/utils/job-utils/racadm-parser.js
+++ b/lib/utils/job-utils/racadm-parser.js
@@ -2,20 +2,23 @@
 
 "use strict";
 
-var di = require('di');
+var di = require('di'),
+    xml2js = require('xml2js');
 
 module.exports = parseRacadmDataFactory;
 di.annotate(parseRacadmDataFactory, new di.Provide('JobUtils.RacadmCommandParser'));
 di.annotate(parseRacadmDataFactory,
     new di.Inject(
         'Assert',
-        '_'
+        '_',
+        'fs'
     )
 );
 
 function parseRacadmDataFactory(
     assert,
-    _
+    _,
+    fs
 ) {
 
     function RacadmCommandParser() {}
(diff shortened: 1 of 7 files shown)


lib/utils/job-utils/racadm-parser.js
```diff
@@ -206,5 +209,87 @@ function parseRacadmDataFactory(
...
+    RacadmCommandParser.prototype.xmlToJson = function(filePath) {
+        var xmlToJsonParser = new xml2js.Parser({attrkey: '$', charkey: '_'});
+        var xmlCatalog = {
+                systemInfo: {},
+                components: []
+            },
+            self = this;
+        return Promise.resolve(filePath)
+            .then(function(filePath){
+                return fs.readFileSync(filePath, 'utf8');
```
yyscamper: Change the readFile to the `async` version, async version is preferred.

lib/utils/job-utils/racadm-tool.js
```diff
@@ -208,13 +210,74 @@ function racadmFactory(
...
+        return Promise.resolve()
+            .then(function(){
+                return self.getSoftwareList(host, user, password);
+            })
+            .then(function(softwareList){
+                if (typeof softwareList.BIOS.FQDD === "undefined"){
+                    throw new Error ('Can not get BIOS FQDD');
+                }
+                fqdd = device || softwareList.BIOS.FQDD;
+                return Promise.resolve();
```
yyscamper: return Promise.resolve() is not necessary.

pengz1:
@RackHD/corecommitters @sunnyqianzhang @iceiilin @WangWinson @cgx027

lib/utils/job-utils/racadm-tool.js
```diff
@@ -53,7 +55,9 @@ function racadmFactory(
         var childProcess = new ChildProcess(
             '/opt/dell/srvadmin/sbin/racadm',
             formatedCmd,
-            {});
+            {},
+            [0, 48]
```
yyscamper: It's better to write some comment why the code `48` should be ignored.

anhou:
test this please

pengz1:
@anhou All utils are tested separated on physical machine. Sunny is working on integrating tests with workflows.

sunnyqianzhang:
@anhou This patch is verified with the workflow today. I'm working on the rest of the code review...

lib/utils/job-utils/racadm-parser.js
```diff
@@ -206,5 +211,85 @@ function parseRacadmDataFactory(
...
+        var componentsList = [], self = this;
+        _.forEach(components, function(component) {
+            var componentCatalog = {
+                FQDD: '',
+                attribute: [], //use Attribute for json2Xml
+                commentedAttribute: []
+            };
+            if (component.Component) {
+                componentCatalog.components = [];
+                componentCatalog.components = self._xmlToJson(component.Component);
```
sunnyqianzhang: Why two values for componentCatalog.components?

sunnyqianzhang:
Talked with Krein, and the code works for me. +1

lib/utils/job-utils/racadm-parser.js
yyscamper: Yes, the above line can be removed.

lib/utils/job-utils/racadm-tool.js
```diff
@@ -192,29 +220,136 @@ function racadmFactory(
...
+            throw new Error('Can not find file path required for iDRAC image update');
+        }
+        cifsUser = cifsConfig.user || '';
+        cifsPassword = cifsConfig.password || '';
 
         var fileInfo = parser.getPathFilename(cifsConfig.filePath);
-        if (!_.endsWith(fileInfo.name, '.d7')){
-            throw new Error('iDRAC image format is not supported');
+        if (!(_.endsWith(fileInfo.name, '.d7') || _.endsWith(fileInfo.name, '.exe') ||
+            _.endsWith(fileInfo.name, '.EXE'))){
```
yyscamper: As file extension is usually not case sensitive, so 'D7' or even 'eXe' should be OK as well. You can use a regex expression, so you don't have to list all possible cases.

lib/utils/job-utils/racadm-tool.js
```diff
@@ -192,29 +220,136 @@ function racadmFactory(
...
+            .then(function(softwareList){
+                if (typeof softwareList.BIOS.FQDD === "undefined"){
+                    throw new Error ('Can not get BIOS FQDD');
+                }
+                fqdd = device || softwareList.BIOS.FQDD;
+            })
+            .then(function(){
+                return parser.getPathFilename(fileStoredPath);
+            })
+            .then(function(fileInfo){
```
yyscamper: This design is OK, so following is just a suggestion, it may be more beautiful and most importantly we can avoid the the bridged variable`fqdd` outside the scope:
```javascript
.then( function()
xxx
var fqdd = device || softwareList.BIOS.FQDD;
var fileInfo parser.getPathFilename(fileStoredPath);
return [fqdd, fileInfo];
}).spread(fqdd, fileInfo) {
 xxx
}
```

lib/utils/job-utils/racadm-tool.js
```diff
@@ -192,29 +220,136 @@ function racadmFactory(
...
+            command =  'get -f ' + filePath + ' -t xml';
+        return Promise.resolve()
+            .then(function(){
+                return self.runCommand(host, user, password, command, 0, 1000);
+            })
+            .then(function(){
+                fs.exists(filePath, function(exist){
+                    if(!exist) {
+                        throw new Error('File /tmp/idra_configure.xml does not exist');
+                    }
```
yyscamper: as `fs.exists` is an async function, so the error will not be threw as your design. You need to warp this function to Promise if you want to use it in the promise chain.

lib/utils/job-utils/racadm-tool.js
yyscamper: just tested, this regex will work:
```javascript
var extReg = /\w*.(d7|exe)$/i;
if (!extReg.test(fileInfo.name)) {
    ...
}
```

lib/utils/job-utils/racadm-tool.js
```diff
@@ -192,29 +220,136 @@ function racadmFactory(
...
-        var command = '';
-        var cifsUser = cifsConfig.user,
-            cifsPassword = cifsConfig.password;
-        assert.string(cifsUser);
-        assert.string(cifsPassword);
+    RacadmTool.prototype.updateFirmware = function(host, user, password, cifsConfig) {
+        var command = '', cifsUser, cifsPassword, self = this;
+        cifsConfig = cifsConfig || {};
+        if (!cifsConfig.filePath){
+            throw new Error('Can not find file path required for iDRAC image update');
```
yyscamper: This error should be wrapped as a rejected Promise, so that the external caller can expect all error is returned from Promise and will avoid the `try/catch` to handle this specified error.

spec/lib/utils/job-utils/racadm-tool-spec.js
```diff
@@ -339,6 +358,12 @@ describe("racadm-tool", function() {
                 this.sandbox.restore();
             });
 
+            it('should throw Error if no pathFile is found', function(){
+                expect( function() {
+                    return instance.setBiosConfig('192.168.188.103', 'admin', 'admin');
+                }).throw(Error, 'Can not find file path required for set BIOS configuration');
```
yyscamper: it's better to expect the setBiosConfig return a rejected Promise not throw error

lib/utils/job-utils/racadm-parser.js
```diff
@@ -2,20 +2,25 @@
 
 "use strict";
 
-var di = require('di');
+var di = require('di'),
+    xml2js = require('xml2js');
```
WangWinson: Does the 'xml2js' library need to be included in `package.json`

lib/utils/job-utils/racadm-tool.js
```diff
@@ -89,6 +95,30 @@ function racadmFactory(
...
+     */
+    RacadmTool.prototype.getLatestJobId = function(host, user, password) {
+        return this.runCommand(host, user, password, "jobqueue view")
+            .then(function(out){
+                return out.slice(out.lastIndexOf("[Job ID"));
+            })
+            .then(function(lastJob){
+                var jobStatus = parser.getJobStatus(lastJob);
+                var jobId = jobStatus.jobId || '';
+                if (jobId.indexOf("JID")!==0){
```
WangWinson: style:space needed around `!==`

anhou:
+1

lib/utils/job-utils/racadm-parser.js
pengz1: xml2js library is not a newly introduced one, it is already included in on-tasks package.json.

spec/lib/utils/job-utils/racadm-tool-spec.js
pengz1: All throw errors are aligned with Promise.reject.

lib/utils/job-utils/racadm-tool.js
pengz1: fs.exists is packed in promise.

yyscamper:
:+1:

PR #139, opened 2016-03-06
Redfish Chassis telemetry pollers

jlongever:
Add Redfish Chassis power and thermal pollers
Add relative pub/sub methods for Redfish command data
Add unit tests

Requires dependency: https://github.com/RackHD/on-core/pull/96

@RackHD/corecommitters @tannoa2 @zyoung51 @uppalk1 @keedya

### DIFF
--- a/lib/jobs/redfish-job.js
+++ b/lib/jobs/redfish-job.js
@@ -85,10 +85,7 @@ function redfishJobFactory(
                 assert.object(data.config, 
                     'Redfish Poller Data Config');
 
-                return Promise.resolve()
-                .then(function() {
-                    return waterline.nodes.needByIdentifier(data.node);
-                })
+                return waterline.nodes.needByIdentifier(data.node)
                 .then(function(node) {
                     var obmSetting = _.find(node.obmSettings, { service: 'redfish-obm-service' });
                     if (obmSetting) {
--- a/lib/jobs/redfish-job.js
+++ b/lib/jobs/redfish-job.js
@@ -1,0 +1,236 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di'),
+    redfish = require('redfish-node');
+
+module.exports = redfishJobFactory;
+di.annotate(redfishJobFactory, new di.Provide('Job.Redfish'));
+di.annotate(redfishJobFactory, new di.Inject(
+    'Job.Base',
+    'Logger',
+    'Util',
+    'Assert',
+    'Promise',
+    '_',
+    'Services.Waterline',
+    'Protocol.Events',
+    'Services.Messenger',
+    'Constants'      
+));
+
+function redfishJobFactory(
+    BaseJob,
+    Logger,
+    util,
+    assert,
+    Promise,
+    _,
+    waterline,
+    protocolEvents,
+    messenger,
+    Constants
+) {
+    var logger = Logger.initialize(redfishJobFactory);
+
+    /**
+     *
+     * @param {Object} options
+     * @param {Object} context
+     * @param {String} taskId
+     * @constructor
+     */
+    function RedfishJob(options, context, taskId) {
+        RedfishJob.super_.call(this, logger, options, context, taskId);
+        
+        this.routingKey = this.context.graphId;
+        assert.uuid(this.routingKey) ;
+
+        this.concurrent = {};
+        this.maxConcurrent = 2;
+    }
+    util.inherits(RedfishJob, BaseJob);
+    
+    /**
+     * @function initClient
+     * @param settings takes OBM settings from node and initializes a new
+     *                 Redfish API client
+     * @description initiate redfish chassis discovery
+     * @return apiClient object
+     */
+    RedfishJob.prototype.initClient = function (settings) {
+        assert.object(settings,
+            'Missing required OBM settings');
+            
+        var apiClient = new redfish.ApiClient();
+        apiClient.basePath = settings.uri.replace(/\/+$/, '');
+        if (settings.username && settings.password) {
+            var token = new Buffer(
+                settings.username + ':' + settings.password
+            ).toString('base64');
+            apiClient.defaultHeaders.Authorization = 'Basic ' + token;
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
+        }
+        return Promise.promisifyAll(new redfish.RedfishvApi(apiClient));
+    };
+
+    /**
+     * @function _run
+     * @description the jobs internal run method
+     */
+    RedfishJob.prototype._run = function run() {
+        var self = this;
+        return waterline.workitems.update({name: "Pollers.Redfish"}, {failureCount: 0})
+        .then(function() {
+            self._subscribeRedfishCommand(self.routingKey, function(data) {
+                if (self.concurrentRequests(data.node, data.workItemId)) {
+                    return;
+                }
+                self.addConcurrentRequest(data.node, data.workItemId);
+                assert.object(data.config, 
+                    'Redfish Poller Data Config');
+
+                return Promise.resolve()
+                .then(function() {
+                    return waterline.nodes.needByIdentifier(data.node);
+                })
+                .then(function(node) {
+                    var obmSetting = _.find(node.obmSettings, { service: 'redfish-obm-service' });
+                    if (obmSetting) {
+                        return obmSetting.config;
+                    }
+                })
+                .then(self.initClient)
+                .then(function(client) {
+                    return self.collectChassisData(
+                        client, 
+                        data.config.command
+                    );
+                })
+                .then(function(result) {
+                    assert.object(result,
+                        'Expected Result Object For: ' + data.config.command
+                    );
+                    data[data.config.command] = result;
+                    return self._publishRedfishCommandResult(
+                        self.routingKey, 
+                        data.config.command, 
+                        data
+                    );
+                })
+                .then(function() {
+                    return waterline.workitems.findOne({ id: data.workItemId });
+                })
+                .then(function(workitem) {
+                        return waterline.workitems.setSucceeded(null, workitem);
+                })
+                .catch(function (err) {
+                    logger.error("Error Retrieving Redfish Data.", {
+                        data: data,
+                        error: err
+                    });
+                })
+                .finally(function() {
+                    self.removeConcurrentRequest(data.node, data.workItemId);
+                });
+            });
+        })
+        .catch(function(err) {
+            logger.error("Failed to initialize job", { 
+                error:err
+            });
+            self._done(err);
+        });
+    };
+
+    /**
+     * @function concurrentRequests
+     * @description manage concurrent work item command requests
+     */
+    RedfishJob.prototype.concurrentRequests = function(node, workItemId) {
+        assert.string(node);
+        assert.string(workItemId);
+        if(!_.has(this.concurrent, node)){
+            this.concurrent[node] = {};
+        }
+        if(!_.has(this.concurrent[node], workItemId)){
+            this.concurrent[node][workItemId] = 0;
+        }
+        if(this.concurrent[node][workItemId] >= this.maxConcurrent){
+            return true;
+        } else {
+            return false;
+        }
+    };
+
+    /**
+     * @function addConcurrentRequest
+     * @description add a new command request for this work item
+     */
+    RedfishJob.prototype.addConcurrentRequest = function(node, type) {
+        assert.object(this.concurrent[node]);
+        assert.number(this.concurrent[node][type]);
+        this.concurrent[node][type] += 1;
+    };
+
+    /**
+     * @function removeConcurrentRequest
+     * @description remove a completed command request for this work item
+     */
+     RedfishJob.prototype.removeConcurrentRequest = function(node, type) {
+        assert.object(this.concurrent[node]);
+        assert.number(this.concurrent[node][type]);
+        this.concurrent[node][type] -= 1;
+    };
+
+    /**
+     * Collect Chassis telemetry data for Redfish Chassis
+     * @param client the redfish API client object
+     */
+    RedfishJob.prototype.collectChassisData = function(client, command) {
+        assert.object(client, 'Redfish Client');
+        assert.string(command, 'Chassis Command');
+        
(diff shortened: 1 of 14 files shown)


lib/jobs/redfish-job.js
```diff
@@ -0,0 +1,226 @@
...
+            'Missing required OBM settings');
+            
+        var apiClient = new redfish.ApiClient();
+        apiClient.basePath = settings.uri.replace(/\/+$/, '');
+        if (settings.username && settings.password) {
+            var token = new Buffer(
+                settings.username + ':' + settings.password
+            ).toString('base64');
+            apiClient.defaultHeaders.Authorization = 'Basic ' + token;
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
```
yyscamper: This will impact globally, so such kind of work should be only done in redfish.ApiClient. Usually, an HTTP/HTTPs request library will give an option something like `rejectUnauthorized`, set it to false will get same result, but the impact is only limited in that request.

lib/jobs/redfish-job.js
```diff
@@ -0,0 +1,226 @@
...
+                        self.routingKey, 
+                        data.config.command, 
+                        data
+                    );
+                })
+                .then(function() {
+                    return waterline.workitems.findOne({ id: data.workItemId });
+                })
+                .then(function(workitem) {
+                        return waterline.workitems.setSucceeded(null, workitem);
```
yyscamper: incorrect indent

lib/jobs/redfish-job.js
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * @function _run
+     * @description the jobs internal run method
+     */
+    RedfishJob.prototype._run = function run() {
+        var self = this;
+        return waterline.workitems.update({name: "Pollers.Redfish"}, {failureCount: 0})
+        .then(function() {
+            self._subscribeRedfishCommand(self.routingKey, function(data) {
```
yyscamper: Since this is a huge callback function, we'd better to move it to a standalone function, not only for readable but also benefit unit-testing.

lib/jobs/redfish-job.js
```diff
@@ -0,0 +1,226 @@
...
+    RedfishJob.prototype.concurrentRequests = function(node, workItemId) {
+        assert.string(node);
+        assert.string(workItemId);
+        if(!_.has(this.concurrent, node)){
+            this.concurrent[node] = {};
+        }
+        if(!_.has(this.concurrent[node], workItemId)){
+            this.concurrent[node][workItemId] = 0;
+        }
+        if(this.concurrent[node][workItemId] >= this.maxConcurrent){
```
yyscamper: Just a minor suggestion, just return this expression is enough, we can save following 4 lines.

lib/jobs/redfish-job.js
```diff
@@ -0,0 +1,226 @@
...
+    };
+
+    /**
+     * @function removeConcurrentRequest
+     * @description remove a completed command request for this work item
+     */
+     RedfishJob.prototype.removeConcurrentRequest = function(node, type) {
+        assert.object(this.concurrent[node]);
+        assert.number(this.concurrent[node][type]);
+        this.concurrent[node][type] -= 1;
```
yyscamper: for sake of safety, the minus 1 operation should be done after judging the `concurrent` is bigger than 0

lib/jobs/run-work-items.js
```diff
@@ -207,6 +208,37 @@ function runWorkItemsJobFactory(
...
+     * @description execute the redfish worker item 
+     */
+    RunWorkItemsJob.prototype._executeRedfishPoller = function (workItem) {
+        assert.object(workItem.config, 'workItem.config');
+        var self = this;
+
+        return Promise.resolve()
+        .then(function() {
+            assert.ok(workItem.config,'workItem.config');
+            assert.string(workItem.config.command, 'workItem.config.command');
```
yyscamper: Duplicated assertion with line#217 and why not move these assertion together with line#217?

lib/jobs/run-work-items.js
```diff
@@ -207,6 +208,37 @@ function runWorkItemsJobFactory(
...
+        })
+        .then(function () {
+            return Promise.resolve(
```
yyscamper: This line is redundant.

lib/jobs/run-work-items.js
```diff
@@ -207,6 +208,37 @@ function runWorkItemsJobFactory(
             );
         });
     };
+    
+    /**
+     * @function _executeRedfishPoller
+     * @description execute the redfish worker item 
+     */
+    RunWorkItemsJob.prototype._executeRedfishPoller = function (workItem) {
```
yyscamper: After examine this function,  I see it only does synchronous tasks, so for me, this async design is a little complicated or redundant.
Though `Promise` can be used to split the synchronous task to sub-tasks so it can be done in several tick-clock not once. But I still don't find there is a need to do this splitting as well.

spec/lib/jobs/create-default-pollers-spec.js
```diff
@@ -143,13 +159,44 @@ describe("Job.Pollers.CreateDefault", function () {
...
+        };
+
+        var job = new CreateDefaultPollers(
+            { nodeId: nodeId, pollers: pollers },
+            { graphId: uuid.v4() },
+            uuid.v4()
+        );
+
+        return job.run()
+        .catch(function(err) {
```
yyscamper: How about the `job.run()` is fulfilled not rejected? so that your assertion logic will not be checked but the unit-test will pass. So you should explicitly fail the test case if the job.run is fulfilled. Below is an example:
https://github.com/RackHD/on-tasks/blob/master/spec%2Flib%2Futils%2Fjob-utils%2Fracadm-tool-spec.js#L156-L164

If no complex assertion logic in catch block, we should using following line to check the promise is rejected, this is the safe way and it will fail if `job.run` is fulfilled:
```
return job.run().to.be.rejectedWith(Error, 'xxxxx');
```

spec/lib/jobs/redfish-job-spec.js
```diff
@@ -0,0 +1,174 @@
...
+                .then(function(data) {
+                    expect(data[0]).to.equal(chassisData[1].body);
+                });
+            });
+        });
+        
+        it("should fail collectChassis for unknown command", function() {
+            redfishApi.listChassisAsync = sandbox.stub().resolves(listChassisData);
+            return redfishJob.collectChassisData(redfishApi, 'unknown')
+            .catch(function(data) {
```
yyscamper: Should explicitly fail the test if `redfishJob.collectChassisData` is fulfilled.

spec/lib/jobs/redfish-job-spec.js
```diff
@@ -0,0 +1,174 @@
...
+                expect(data.message).to.equal('Unsupported Chassis Command: unknown');
+            });
+        });
+        
+        it("should fail collectChassis with no data", function() {
+            redfishApi.listChassisAsync = sandbox.stub().resolves(listChassisData);
+            redfishApi.getPowerAsync = sandbox.stub().resolves([null, {body: undefined}]);
+            return redfishJob.collectChassisData(redfishApi, 'power')
+            .catch(function(data) {
```
yyscamper: Should explicitly fail the test if promise is fulfilled.

spec/lib/jobs/redfish-job-spec.js
```diff
@@ -0,0 +1,174 @@
...
+                expect(data.message).to.equal('No Data Found For Command: power');
+            });
+        });
+        
+        it("should fail getPowerAsync with error", function() {
+            redfishApi.listChassisAsync = sandbox.stub().resolves(listChassisData);
+            redfishApi.getPowerAsync = sandbox.stub().rejects({response: {text : 'error text'}});
+            return redfishJob.collectChassisData(redfishApi, 'power')
+            .catch(function(data) {
```
yyscamper: Should explicitly fail the test if promise is fulfilled.

spec/lib/jobs/base-job-spec.js
```diff
@@ -195,7 +241,7 @@ describe("Base Job", function () {
                 });
             });
         });
-
+
```
yyscamper: Suggest to clean out these unwanted spaces

yyscamper:
:-1: this PR due to the global setting for `process.env.NODE_TLS_REJECT_UNAUTHORIZED` and the unit-testing case design for rejected promise.

Meanwhile, I see there is an enhancement opportunity, redfish-job, ipmi-job and snmp-job share some common code, like function `concurrentRequests`, `addConcurrentRequest` and `removeConcurrentRequest`, they should be better move to a common place (like job-util).

lib/jobs/redfish-job.js
jlongever: This is meant to be temporary. Since the API client uses super agent requests, there is no way to pass rejectUnauthorize to it. The suggested workaround is to set this ENV https://github.com/visionmedia/superagent/issues/188.

I'll add a finally block to restore the value.

lib/jobs/redfish-job.js
jlongever: This follows the exact same pattern as the other poller _run() implementantions.  Are you suggesting we refractor those as well?

lib/jobs/redfish-job.js
jlongever: Again, copied pattern from the other poller job implementations.

lib/jobs/redfish-job.js
jlongever: Copied pattern from other poller job implementation.  However I agree, but removeConcurrentRequest is never called unless a request was added.

lib/jobs/run-work-items.js
jlongever: Its just where I ended up putting them.  I can move them if you feel it breaks functionality.

lib/jobs/run-work-items.js
jlongever: This copies the same pattern as the other poller work item code.  Are you suggesting those should be refactored?

jlongever:
@yyscamper I'll address some of your concerns.  However the common code refactoring you're suggesting is outside the scope of this work (and impacts very little).  Please propose or create a story to address it. Thanks.

lib/jobs/redfish-job.js
yyscamper: Restore it in the finally block is OK for me, meanwhile, I see from you posted link, the "0" should be number `0`

yyscamper:
@jlongever , it's OK for me, we can open another story to track some enhancement, please at least fix the `process.env` and unit testing for rejected promise.

lib/jobs/redfish-job.js
yyscamper: @jlongever , please also add a comment for this

jlongever: "0" should get converted to an integer.

jlongever: @yyscamper  NODE_TLS_REJECT_UNAUTHORIZED is a environment variable,  which are strings.

jlongever:
@yyscamper ^^

spec/lib/utils/job-utils/redfish-tool-spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+describe("redfish-tool", function() {
+    var redfishTool;
+    
+    before(function() {
+         helper.setupInjector([
+            helper.require('/lib/utils/job-utils/redfish-tool.js'),
+        ]);
+        redfishTool = helper.injector.get('JobUtils.RedfishTool');
+    });
```
yyscamper: I think you forgot to backup the `process.env.NODE_TLS_REJECT_UNAUTHORIZED` before starting this test suite and recovery it after test suite done. We cannot ensure mocha runs the test cases in a sequence just like we wrote. So it's better to do the backup and recovery global variables in order to not impact other test suites.

yyscamper:
Thanks @jlongever , it's :+1:  after fixing the process.env back & recovery in unit-testing. Others looks good to me.

spec/lib/utils/job-utils/redfish-tool-spec.js
jlongever: There is no need to back up the value again in these unit-tests.  Other tests relying on TLS (which seems wrong for a unit test to begin with) should set this variable accordingly.

yyscamper: What you said is true for test case itself because test cases should always define clear input arguments. But this may impact some test libraries, especially those http request mocking or assertion library, such as *supertest* and *nock*, we heavily rely them on some unit testing. I haven't study those library's implementation, so I am not sure it will use this env variable or not.

The env variable that defined by node.js itself is sensitive and its impact is very broad. It will impact every http request's behavior if it is set by my understanding. I think you don't want to place a line to set this env everywhere to request https, Keep it in Node.js's default setting is convenient. However, if the env is owned by RackHD itself, I think the impact is relative limited.

Anyway, cleanup the spot is always a recommendation from me though there may be no bugs even no cleanup at present. 
Meanwhile, I think this also doesn't bring much code work for your case.

zyoung51:
+1

lib/jobs/redfish-job.js
benbp: I agree with @yyscamper, I think it will be more readable and testable if the code is split up into helper functions called from `_run()`. I think just because other code is similar doesn't mean there isn't room for improvement in future code. Not a blocking issue for me though.

benbp:
:+1: same on the unit test env.

lib/jobs/redfish-job.js
jlongever: @benbp I'm following the same convention as all of the RackHD pollers, and this seems minor, so there was no immediate need to break from that convention in this PR.   If we think there is room to improve the poller job design then we should submit a proposal, user story or another PR to address it.

jlongever: @yyscamper @benbp let's keep things consistent util it's time to refactor,  then we can refactor the entire design at once and not carry multiple variations of the design.   IMHO.

yyscamper: Sound good to me. It's better to take an action to do all refactor for ipmi/snmp/redfish pollers together, not do it one by one.

PR #143, opened 2016-03-18
resolves issue https://github.com/RackHD/RackHD/issues/138, change to run graph with new API and fix unittest

yyscamper:
All Changes:
1. resolves https://github.com/RackHD/RackHD/issues/138, change to run graph with new API
2. Create new workflowTool utility to wrap some helper functions in on-tasks repos.
3. Fix the unittest problem of isc-dhcp-lease-poller about the incorrect calcuation of future date.

**NOTE**
1. The unit-test fails is expected as this PR **depends** on https://github.com/RackHD/on-core/pull/102, so **DON'T** merge this PR until https://github.com/RackHD/on-core/pull/102 is merged.
2. We haven't created unit-testing before for the job `run-graph` and `run-sku-graph`, so my PR doesn't includes those unit-test change. I will create another PR to fix this technical debt.

@RackHD/corecommitters @ytjohn @tldavies @iceiilin @WangWinson @cgx027 @sunnyqianzhang @pengz1

### DIFF
--- a/spec/lib/jobs/isc-dhcp-poller-job-spec.js
+++ b/spec/lib/jobs/isc-dhcp-poller-job-spec.js
@@ -7,12 +7,13 @@ describe('ISC DHCP Poller Job', function () {
     var base = require('./base-spec');
     var uuid;
 
-    /// create a future end date
-    var todayDate = new Date();
-    // add 30 seconds
-    var newdate = new Date(todayDate.getTime() + (10000));
+    // create a future end date
+    // consider different timezone and the daylight saving, add 2 days bases on now will always
+    // generate a future date if discard timezone while parsing.
+    var futureDate = new Date();
+    futureDate.setDate(futureDate.getDate() + 2); //add 2 days to current
     // use toISOString, it's the closest to Y-m-d H:i:s
-    newdate = newdate.toISOString();
+    var newdate = futureDate.toISOString();
     // modify ISO string to get Y-m-d H:i:s
     newdate = newdate.split('.')[0].replace('T', ' ').replace(/-/g, '/');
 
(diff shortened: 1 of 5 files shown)


ytjohn:
:+1:

spec/lib/jobs/isc-dhcp-poller-job-spec.js
```diff
@@ -7,12 +7,13 @@ describe('ISC DHCP Poller Job', function () {
...
 
-    /// create a future end date
-    var todayDate = new Date();
-    // add 30 seconds
-    var newdate = new Date(todayDate.getTime() + (10000));
+    // create a future end date
+    // consider different timezone and the daylight saving, add 2 days bases on now will always
+    // generate a future date if discard timezone while parsing.
+    var futureDate = new Date();
+    futureDate.setDate(futureDate.getDate() + 2); //add 2 days to current
```
yyscamper: @tldavies could you please review this change, as I see you are the author who created the first line code. 

This is to fix the unit-test failure for isc-dhcp-lease-poller, I guess this problem is caused by the daylight saving change, because we had not found this issue before. Previous calculation for future only adds extra 10000ms, but the below line `newdate = newdate.split('.')[0].replace('T', ' ').replace(/-/g, '/');` will remove some critical timing info, so when it is eventually parsed, the time may not be in the future.

My fix is to add extra 2 days into the calculation, so regardless the daylight saving and timezone, the new date will always be in the future.

I admit the best solution is to clarify those date format difference (UTC/ISO/Local, .etc) and understand all necessary Javascript Date function, but that's will take relative more time for me and I think for unit-testing, simply add 2 days to generate future days is enough.

yyscamper:
test this please

anhou:
:+1:

PR #144, opened 2016-03-22
Adding support for Ansible tasks to return results via file.

stevemendes:
Currently there is no way for a task that runs an Ansible playbook to return any results to the workflow.  This change adds support for Ansible playbooks to write their results to a file, whose contents are inserted into the context for that task.

This change would allow ansible-job.js to:
   1) Create a file name in the system 'tmp' area called 'ansible-<taskId>.out' 
   2) add the file name to the extraVars that get passed to the playbook
   3) after the playbook completes, the contents of the file (if it exists) are added to the task's context

It is the responsibility of the playbook to pick up the file name and write to it.  If the file is missing, the operation completes normally and a msg is logged.    The file is removed (unlinked) afterwards.

### DIFF
--- a/lib/jobs/ansible-job.js
+++ b/lib/jobs/ansible-job.js
@@ -3,7 +3,9 @@
 'use strict';
 
 module.exports = ansibleJobFactory;
-var di = require('di');
+var di = require('di'),
+    os = require('os'),
+    path = require('path');
 
 di.annotate(ansibleJobFactory, new di.Provide('Job.Ansible.Playbook'));
 di.annotate(ansibleJobFactory, new di.Inject(
@@ -12,7 +14,9 @@ di.annotate(ansibleJobFactory, new di.Inject(
     'Logger',
     'JobUtils.Ansible.Playbook',
     'Assert',
-    'Services.Waterline'
+    'Services.Waterline',
+    'Promise',
+    'fs'
     )
 );
 
@@ -22,16 +26,33 @@ function ansibleJobFactory(
         Logger,
         AnsibleTool,
         assert,
-        waterline
+        waterline,
+        Promise,
+        nodeFs
 ) {
     var logger = Logger.initialize(ansibleJobFactory);
+    var fs = Promise.promisifyAll(nodeFs);
 
-     function AnsibleJob(options, context, taskId) {
+    function AnsibleJob(options, context, taskId) {
         AnsibleJob.super_.call(this, logger, options, context, taskId);
         this.ansibleCmd = undefined;
         this.nodeId = this.context.target;
         this.routingKey = context.graphId;
-        this.extraVars = options.vars ? JSON.stringify(options.vars) : null;
+        this.taskKey = taskId;
+        // file for Ansible playbook to write results to:
+        this.outFileName = path.join(os.tmpdir(),'ansible-' + this.taskKey + '.out'); 
+
+
+        if(options.vars === undefined) {
+            this.extraVars = '{"ansibleResultFile":"' + this.outFileName + '"}';
+        } else {
+            this.extraVars = options.vars;
+            this.extraVars.ansibleResultFile = this.outFileName;
+            this.extraVars = JSON.stringify(this.extraVars);
+        }
+        if (this.context.ansibleResultFile === undefined) {
+            this.context.ansibleResultFile = [];
+        }
         logger.debug("ansible job", {
                 options: options,
                 extraVars: this.extraVars || 'none'
@@ -65,12 +86,29 @@ function ansibleJobFactory(
             return self._publishAnsibleResult(self.routingKey, data);
         })
         .then(function() {
+            return fs.readFileAsync(self.outFileName, 'utf8');
+        })
+        .then(function(contents) {
+            assert.arrayOfString(self.context.ansibleResultFile);
+            self.context.ansibleResultFile.push(contents);
+            fs.unlink(self.outFileName, function(err) {
+                if (err) {
+                    logger.warning("Failed to delete " + self.outFileName);
+                }
+            });
+        })
+        .then(function() {
             self.ansibleCmd = undefined;
             self._done();
         })
         .catch(function(err) {
             self.ansibleCmd = undefined;
-            self._done(err);
+            if(err.code === 'ENOENT') {
+                logger.warning("Result file not found. " + err.message);
+                self._done();
+            } else {
+                self._done(err);
+            }
         });
     };
 
@@ -82,6 +120,5 @@ function ansibleJobFactory(
             this.ansibleCmd.kill();
         }
     };
-
     return AnsibleJob;
 }


jlongever:
ok to test

lib/jobs/ansible-job.js
```diff
@@ -78,10 +95,23 @@ function ansibleJobFactory(
      * @memberOf AnsibleJob
      */
      AnsibleJob.prototype._cleanup = function cleanup() {
+        var self = this;
+        try {
+            var data = fs.readFileSync(self.outFileName, 'utf8');
```
yyscamper: The synchronous version API is only used in unit-testing, if there is no strong reason that we have to use the synchronous API,  otherwise we need use the async API for better performance concern.

lib/jobs/ansible-job.js
```diff
@@ -78,10 +95,23 @@ function ansibleJobFactory(
...
+            self.context.ansibleResultFile.push(data);
+            fs.unlink(self.outFileName, function(err) {
+                if (err) {
+                    logger.warning("Uable to remove temporary Ansible " +
+                        "result file " + self.outFileName + " " + err);
+                }
+            });
+        } catch (e) {
+            logger.info("Error reading Ansible result file " +
```
yyscamper: logger.error?

lib/jobs/ansible-job.js
```diff
@@ -3,7 +3,10 @@
 'use strict';
 
 module.exports = ansibleJobFactory;
-var di = require('di');
+var di = require('di'),
+    os = require('os'),
+    fs = require('fs'),
```
yyscamper: The `fs` has been included in the injectable table, you can move it to below list `di.inject(...)`

lib/jobs/ansible-job.js
```diff
@@ -31,7 +34,20 @@ function ansibleJobFactory(
...
+        this.outFileName = path.join(os.tmpdir(),'ansible-' + this.taskKey + '.out'); 
+
+        this.extraVars = '{"ansibleResultFile":"' + this.outFileName + '"}';
+        if(options.vars !== null) { 
+             this.extraVars = options.vars;
+             this.extraVars.ansibleResultFile = this.outFileName; 
+             this.extraVars = JSON.stringify(this.extraVars);
+        }
+        if (this.context.ansibleResultFile === undefined) {
+            this.context.ansibleResultFile = [];
```
yyscamper: If `this.context.ansibleResultFile` is object/null/number or other non-collection types, then the line#101 `self.context.ansibleResultFile.push(data)` will fail as no `push` method is defined. So it's better to add an assertion for this.

yyscamper:
@benbp, please also review this PR.
This is another good example of how to store the returned result of a job, I don't think the context is a good way, as it introduces the dependency between tasks. But I agree on using the context right now as a temporary solution before the task returning value design is completed.

lib/jobs/ansible-job.js
stevemendes: I tried using the async version at first, but I found that sometimes the contents of the file did not end up in the context.  I think the context was getting set before the file read was complete, so I changes to the sync version.

lib/jobs/ansible-job.js
stevemendes: I don't think it should be an error, since the Ansible playbook may not always write a result file (e.g.: if there is no result to return).  So it is not an error if the file is not present.  Perhaps it should be a warning?

lib/jobs/ansible-job.js
yyscamper: If so, that should be a problem if no content in context. What's more, the pushing result seems to be better in the `_run` routine instead of '_cleanup`, what's your concern for this?

lib/jobs/ansible-job.js
yyscamper: Warning should be better that error, but if that's not a error case, we can keep using the `info` but with a little change for the message, like `no ansible result file is present` or something else.

yyscamper: Just the `Error` word may confuse user.

lib/jobs/ansible-job.js
stevemendes: agreed.  will change.

lib/jobs/ansible-job.js
stevemendes: Will change to warning and reword message.  thx.

lib/jobs/ansible-job.js
stevemendes: I'll move code into _run, that looks better.  And I'm not sure I understand the 1st part of your comment (the 'context.ansibleResultFile' var is initialized up front - is that enough?

yyscamper: My meaning is that it's a problem that need be resolved as you mentioned sometimes the file content was not in the context.

yyscamper:
@stevemendes, I think you still miss the unittest updating and new cases adding, thus travisci failed and coverage decreased.

lib/jobs/ansible-job.js
benbp: Either way, it would probably be better to use a promisified version of the async readFile call, that way the job can't complete and pass execution to the next task (which may read context) before it is actually done reading the data and setting the context. You can see some example usage of this here:

Promisify fs methods: https://github.com/RackHD/on-core/blob/0be650eb0c5d21f4ee854463ab0133d3d878fea6/lib/common/file-loader.js#L17
Call the promisified method: https://github.com/RackHD/on-core/blob/0be650eb0c5d21f4ee854463ab0133d3d878fea6/lib/common/file-loader.js#L48

stevemendes: Switched to promisified version (thx) but still need readFileSync for some reason (when I try Async, I don't get my results in the context)

benbp:
Resolves https://github.com/RackHD/RackHD/issues/102

lib/jobs/ansible-job.js
benbp: It's because you aren't blocking on resolution of the async request I believe, so by the time you update the context, we've already killed the rest of the job and persisted the old context.

lib/jobs/ansible-job.js
```diff
@@ -22,16 +26,33 @@ function ansibleJobFactory(
...
         this.ansibleCmd = undefined;
         this.nodeId = this.context.target;
         this.routingKey = context.graphId;
-        this.extraVars = options.vars ? JSON.stringify(options.vars) : null;
+        this.taskKey = taskId;
+        // file for Ansible playbook to write results to:
+        this.outFileName = path.join(os.tmpdir(),'ansible-' + this.taskKey + '.out'); 
+
+
+        if(options.vars == undefined) {
```
anhou: this need to be fixed

lib/jobs/ansible-job.js
```diff
@@ -65,6 +86,23 @@ function ansibleJobFactory(
             return self._publishAnsibleResult(self.routingKey, data);
         })
         .then(function() {
+            try {
+                var data = fs.readFileSync(self.outFileName, 'utf8');
```
anhou: @stevemendes  based on your previous comments with @yyscamper , so fs.readFileAsync has some problems?  a example for it is https://github.com/RackHD/on-http/blob/master/lib/services/schema-api-service.js#L41

lib/jobs/ansible-job.js
stevemendes: done

lib/jobs/ansible-job.js
stevemendes: Thank you for the example, anhou.  It is better now.

lib/jobs/ansible-job.js
stevemendes: I see.  anhou provided example.  Fixed now.  Thx!

stevemendes:
Actually - do not pull yet - the change to Async is NOT working as expected.  I need to run more tests.

lib/jobs/ansible-job.js
```diff
@@ -65,6 +86,24 @@ function ansibleJobFactory(
             return self._publishAnsibleResult(self.routingKey, data);
         })
         .then(function() {
+            try {
```
zyoung51: You can leave out the try/catch blocks and handle it in the .catch();
return fs.readFileAsync(self.outFileName, 'utf8').then(function(contents) {  
  return fs.unlinkAsync(self.outFileName);
}).catch(function(err) {
  // error handling
});

stevemendes: I was able to rework the error handling and put the call to end the workflow in the callback of the file read.  (I just saw the part of your note regarding adding a promise - and although I had trouble with error handling, it seems to work OK now - let me know if you still think the promise should be in there).  Thx.

benbp:
LGTM to me now, the last thing I'll request is can you squash the commits down so we don't have 13 commits for <100 LoC? Thanks!

:+1: after the squash.

jlongever:
test this please

stevemendes:
Squash is complete.

benbp:
Thanks!

lib/jobs/ansible-job.js
```diff
@@ -65,12 +86,29 @@ function ansibleJobFactory(
             return self._publishAnsibleResult(self.routingKey, data);
         })
         .then(function() {
+            return fs.readFileAsync(self.outFileName, 'utf8');
+        })
+        .then(function(contents) {
+            assert.arrayOfString(self.context.ansibleResultFile);
+            self.context.ansibleResultFile.push(contents);
+            fs.unlink(self.outFileName, function(err) {
```
yyscamper: @stevemendes , the fs.unlink should also use the Promise, otherwise the next then will be executed before the unlink completes.

Below is the sample code:
```
.then(function() {
    return fs.readFileAsync(self.outFileName, 'utf8');
})
.then(function() {
    ...
    return fs.unlinkAsync(self.outFileName);
})
.then(function() {
    ...
    self._done();
})
...
```

yyscamper:
:+1: after changing `unlink` to Promise style.

lib/jobs/ansible-job.js
stevemendes: It doesn't matter if the unlink does not complete before the workflows exits, so I did not use the promise there.  Should we still use the promise anyway?

benbp: Yeah I also read the code as not concerning the job success or execution with the results of unlink, which in this case I think is fine, especially since you log failures.

yyscamper: OK, so my review result is :+1: now.

stevemendes:
Thx for all your help & bearing with me as I ramp up :)

benbp:
No problem @stevemendes. Thanks for the contribution!

PR #147, opened 2016-03-28
adding default version for CoreOS install

heckj:
per feedback from https://github.com/RackHD/on-http/pull/187, adding a default 'version' option for the `Task.Os.Install.CoreOS` task so that we can leverage it from the RackHD demonstration install


benbp:
:+1:, see https://github.com/RackHD/on-http/pull/187#issuecomment-202515388 for reason why `version` can't be combined with `repo`

anhou:
:+1:

yyscamper:
:+1:

PR #148, opened 2016-03-28
bootstrap file changed to reflect new kernel

DavidjohnBlodgett:
@RackHD/corecommitters @tldavies @VulpesArtificem @rolandpoulter  this is one of a 2 part PR (also in https://github.com/RackHD/on-imagebuilder/pull/31) to support the new micro kernel.

### DIFF
--- a/lib/task-data/tasks/bootstrap-ubuntu.js
+++ b/lib/task-data/tasks/bootstrap-ubuntu.js
@@ -7,13 +7,13 @@ module.exports = {
     injectableName: 'Task.Linux.Bootstrap.Ubuntu',
     implementsTask: 'Task.Base.Linux.Bootstrap',
     options: {
-        kernelVersion: '3.19.0-56-generic'
-        kernelFile: 'vmlinuz-{{ options.kernelVersion}}',
-        initrdFile: 'initrd.img-{{ options.kernelVersion}}',
+        kernelVersion: '3.19.0-56-generic',
+        kernelFile: 'vmlinuz-{{ options.kernelVersion }}',
+        initrdFile: 'initrd.img-{{ options.kernelVersion }}',
         kernelUri: '{{ api.server }}/common/{{ options.kernelFile }}',
         initrdUri: '{{ api.server }}/common/{{ options.initrdFile }}',
-        basefs: 'common/base.trusty.{{ options.kernelVersion}}.squashfs.img',
-        overlayfs: 'common/discovery.{{ options.kernelVersion}}.overlay.cpio.gz',
+        basefs: 'common/base.trusty.{{ options.kernelVersion }}.squashfs.img',
+        overlayfs: 'common/discovery.{{ options.kernelVersion }}.overlay.cpio.gz',
         profile: 'linux.ipxe',
         comport: 'ttyS0'
     },
--- a/lib/task-data/tasks/bootstrap-ubuntu.js
+++ b/lib/task-data/tasks/bootstrap-ubuntu.js
@@ -7,12 +7,12 @@ module.exports = {
     injectableName: 'Task.Linux.Bootstrap.Ubuntu',
     implementsTask: 'Task.Base.Linux.Bootstrap',
     options: {
-        kernelFile: 'vmlinuz-3.16.0-25-generic',
-        initrdFile: 'initrd.img-3.16.0-25-generic',
+        kernelFile: 'vmlinuz-3.19.0-56-generic',
+        initrdFile: 'initrd.img-3.19.0-56-generic',
         kernelUri: '{{ api.server }}/common/{{ options.kernelFile }}',
         initrdUri: '{{ api.server }}/common/{{ options.initrdFile }}',
-        basefs: 'common/base.trusty.3.16.0-25-generic.squashfs.img',
-        overlayfs: 'common/discovery.overlay.cpio.gz',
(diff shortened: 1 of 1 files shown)


benbp:
@DavidjohnBlodgett also need some doc/build updates at https://github.com/RackHD/RackHD/blob/master/packer/ansible/roles/images/tasks/main.yml#L15-L19 and https://github.com/RackHD/docs/blob/25c037d64ecd54b7e4960efe7e976923c559dd2f/docs/rackhd/ubuntu_package_installation.rst (search for 3.16).

benbp:
Please don't merge until ^^ are updated.

anhou:
:+1: after update with @benbp 's comments

lib/task-data/tasks/bootstrap-ubuntu.js
```diff
@@ -7,12 +7,12 @@ module.exports = {
...
-        kernelFile: 'vmlinuz-3.16.0-25-generic',
-        initrdFile: 'initrd.img-3.16.0-25-generic',
+        kernelFile: 'vmlinuz-3.19.0-56-generic',
+        initrdFile: 'initrd.img-3.19.0-56-generic',
         kernelUri: '{{ api.server }}/common/{{ options.kernelFile }}',
         initrdUri: '{{ api.server }}/common/{{ options.initrdFile }}',
-        basefs: 'common/base.trusty.3.16.0-25-generic.squashfs.img',
-        overlayfs: 'common/discovery.overlay.cpio.gz',
+        basefs: 'common/base.trusty.3.19.0-56-generic.squashfs.img',
+        overlayfs: 'common/discovery.3.19.0-56-generic.overlay.cpio.gz',
```
yyscamper: How about to refactor the code using the task rendering? so it is more easy to change all micorkernel files to a new version by only change the options version:
```
options: {
        version: '3.19.0-56-generic',
        kernelFile: 'vmlinuz-{{options.version}}',
        initrdFile: 'initrd.img-{{options.version}}',
        kernelUri: '{{ api.server }}/common/{{ options.kernelFile }}',
        initrdUri: '{{ api.server }}/common/{{ options.initrdFile }}',
        basefs: 'common/base.trusty.{{options.version}}.squashfs.img',
        overlayfs: 'common/discovery.{{options.version}}.overlay.cpio.gz',
        profile: 'linux.ipxe',
        comport: 'ttyS0'
    },
```

yyscamper:
Still :+1: with a refactor suggestion.

lib/task-data/tasks/bootstrap-ubuntu.js
benbp: Great idea! I second this, though maybe call it `kernelVersion` instead of `version`?

jlongever: Why just version?  Why not the entire path/filename?

benbp: @jlongever because the `version` value is shared across 4 variables, if I understand your question correctly?

benbp: @DavidjohnBlodgett ^^

jlongever: @benbp as long as we're rendering option values why not pass each full filename and path in as an option?  What if I want to boot my own custom overlay (or initrd, kernel, etc) called `my.custom.overlay.version.1.cpio.gz` put in `my_overlays/` instead of `/common`;  my graph would pass `overlayfs` and rendered as:
`overlayfs: {{option.overlayfs}}` = `my_overlays/my.custom.overlay.version.1.cpio.gz`.

Are we intentionally enforcing a file naming convention?  Just throwing it out there.

benbp: I think the extra options are just for convenience on our end when changing the code, but for a use case like yours, yeah I would encourage just overwriting the whole overlayfs value from the grah.

yyscamper: Yes, my proposal just wants to simplify our future change on this task definition, you can still override all options with your customized values regardless of the `version`. 
Furthermore, maybe the `common/` can also be extracted as a option, such as `baseUrl` or `baseDir` .

anhou: Talking about refactoring, I suggest more. the URI/URL could be independent with kernel/initrd/basefs/overlayfs so that the static file server could be independent, it's very useful when a separate service for static file is used(also may benefit to RackHD in switch),  so just like this.
{
   ***
   baseUri: {{api.server}/common
   kernelFile: ***
   initrdFil: ***
   basefs: ***
   overlayfs: ***
   ***
}
and for ipxe profile https://github.com/RackHD/on-http/blob/master/data/profiles/linux.ipxe also need to be refactored, 
kernel <%=baseUri%>/<%=kernelFile%>
initrd <%=baseUri%>/<%=initrdFile%>
imgargs imgargs <%=kernelFile%> initrd=<%=initrdFile%> ****** BASEFS=<%=baseUri%>/<%=basefs%> OVERLAYFS=<%=baseUri%>/<%=overlayfs%> ***

so the hardcoded 'http' could be removed from 
https://github.com/RackHD/on-imagebuilder/blob/master/roles/initrd/provision_initrd/files/local#L109-L121

I know this is out of the original purpose of this PR, so it could be refactored in another PR,  so still :+1:  from me for this PR with another PR refactoring

PR #152, opened 2016-03-31
Add new Task to run a UEFI application (if compute is running in UEFI)

keedya:
depends on : https://github.com/RackHD/on-http/pull/201
depends on : https://github.com/RackHD/on-taskgraph/pull/81


keedya:
@RackHD/corecommitters @tannoa2 @uppalk1 @zyoung51

lib/jobs/uefi-job.js
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright 2016, EMC, Inc.
```
benbp: ugh, this is a new one. We'll have to figure out how to ignore it :)

spec/lib/jobs/uefi-job-spec.js
```diff
@@ -0,0 +1,82 @@
...
+        expect(job.nodeId).to.equal('testid');
+    });
+
+    it("should have a profile value", function () {
+        expect(job.profile).to.equal('testprofile');
+    });
+
+    it("should set up message subscribers", function () {
+        var cb;
+        return job._run().then(function () {
```
benbp: There doesn't really seem to be a reason to have the `_run()` method return a promise at all. In test, you could just call `job._run()` and then check your stubs, since it will all be called synchronously.

keedya:
Addressed @benbp  Comments

benbp:
:

lib/jobs/uefi-job.js
```diff
@@ -0,0 +1,57 @@
...
+        assert.string(options.repo);
+        this.nodeId = this.context.target;
+        this.profile = this.options.profile;
+    }
+    util.inherits(UefiJob, BaseJob);
+
+    UefiJob.prototype._run = function () {
+        var self = this;
+        self._subscribeRequestProfile(function () {
+            return self.profile;
```
yyscamper: Since the `profile` is in the base-task's `requiredOptions` list and it is a must-have options for `subscribeRequestProfile`, so I suggest adding an assertion logic for the existing of `profile`

lib/jobs/uefi-job.js
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = uefiJobFactory;
+di.annotate(uefiJobFactory, new di.Provide('Job.Uefi'));
```
yyscamper: I have some concern about the job's name, the `Job.Uefi` is a so general word, seems it will cover all UEFI related work, but as you described in the PR, this is used to run UEFI application, so `Job.Uefi.App` or `Job.RunUefiApp` seems better.

lib/jobs/uefi-job.js
yyscamper: I have tried to clone @keedya 's code to my local, the jshint doesn't report such error. 
I also did some search, seems this is probably due to the incorrect file encoding, change the file-encoding to utf-8 will resolve this issue.
@keedya , please check whether it is due to the file-encoding. If it is a real false-alarm, I think we can open an issue for houndci.

yyscamper:
:+1: after my two questions are either by clarified or fixed.

tannoa2:
+1

lib/jobs/uefi-job.js
keedya: @yyscamper didn't find any file format problems in on-tasks. I had to fix format in on-taskgraph but I am still seeing the issue

lib/jobs/uefi-job.js
keedya: @yyscamper We can change the name in the future if we need separate jobs for uefi related workflows.

lib/jobs/uefi-job.js
zyoung51: @yyscamper Do you have a suspicion that the workflow engine code that already validated the presence of the requiredOptions is broken?  Should an issue be opened against this code: https://github.com/RackHD/on-core/blob/9b94bbc61254f936d7624451baf4858703860414/lib/workflow/task-graph.js#L325

zyoung51:
+1 after the file encoding issues are fixed:

file -i lib/jobs/uefi-job.js
lib/jobs/uefi-job.js: text/x-c++; charset=utf-8

Everything else in the repo is us-ascii encoded.

keedya:
@zyoung51  all set!!

lib/jobs/uefi-job.js
keedya: @yyscamper added the assertion as requested

brianparry:
:+1:

lib/jobs/uefi-job.js
yyscamper: @zyoung51, the baseTask requiredPropertes only check the existing, so a number or even null will pass, so the assertion logic to check whether it is string is still necessary in job.

lib/jobs/uefi-job.js
yyscamper: ok

yyscamper:
:+1:

PR #153, opened 2016-04-01
modify redfish discovery to optionally catalog EMC elements

zyoung51:
- Modify the redfish discovery task to add newly cataloged
  system/chassis to the graph context
- Modify the EMC redfish catalog to iterate through chassis
  or the specified target
- Add condition job for basic logic handling in the graph flow

### DIFF
--- a/lib/jobs/condition.js
+++ b/lib/jobs/condition.js
@@ -1,0 +1,43 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = conditionJobFactory;
+var di = require('di');
+
+di.annotate(conditionJobFactory, new di.Provide('Job.Evaluate.Condition'));
+di.annotate(conditionJobFactory, new di.Inject(
+    'Job.Base',
+    'Util',
+    'Logger',
+    'Assert'
+    )
+);
+
+function conditionJobFactory(
+        BaseJob,
+        util,
+        Logger,
+        assert
+) {
+    var logger = Logger.initialize(conditionJobFactory);
+
+     function ConditionJob(options, context, taskId) {
+        ConditionJob.super_.call(this, logger, options, context, taskId);
+        assert.string(options.when);
+    }
+    util.inherits(ConditionJob, BaseJob);
+
+    /**
+     * @memberOf ConditionJob
+     */
+    ConditionJob.prototype._run = function run() {
+        var when = this.options.when.toLowerCase();
+        if( when === 'true' ) {
+            return this._done();
+        }
+        this._done(new Error('condition evaluated to false'));
+    };
+
+    return ConditionJob;
+}
--- a/lib/jobs/emc-redfish-catalog.js
+++ b/lib/jobs/emc-redfish-catalog.js
@@ -42,10 +42,8 @@ function EmcRedfishCatalogJobFactory(
                                    options,
                                    context,
                                    taskId);
-        assert.string(this.context.target);
-        this.nodeId = this.context.target;
-        this.redfish = new RedfishTool();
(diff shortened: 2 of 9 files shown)


lib/jobs/condition.js
```diff
@@ -0,0 +1,40 @@
...
+     function ConditionJob(options, context, taskId) {
+        ConditionJob.super_.call(this, logger, options, context, taskId);
+    }
+    util.inherits(ConditionJob, BaseJob);
+
+    /**
+     * @memberOf ConditionJob
+     */
+    ConditionJob.prototype._run = function run() {
+        var when = this.options.when.toLowerCase();
```
yyscamper: Needs an assertion for `options.when` or need a default handling routine if `options.when` is not defined.

lib/jobs/emc-redfish-catalog.js
```diff
@@ -42,10 +42,8 @@ function EmcRedfishCatalogJobFactory(
                                    options,
                                    context,
                                    taskId);
-        assert.string(this.context.target);
-        this.nodeId = this.context.target;
-        this.redfish = new RedfishTool();
-        this.redfish.setup(this.nodeId);
+        this.chassis = this.context.chassis ||  [ this.context.target ] || [];
```
yyscamper: Change the `chassis` to plural will be better to express its meaning.

yyscamper: Change the `chassis` to plural will be better to express its meaning.

lib/task-data/base-tasks/condition.js
```diff
@@ -0,0 +1,14 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: 2016

spec/lib/jobs/condition-job-spec.js
```diff
@@ -0,0 +1,29 @@
...
+    it("should run and finish with success", function() {
+        var job = new ConditionJob({when: 'true'}, {}, graphId);
+        job._run();
+        return job._deferred;
+    });
+
+    it("should run and finish with failure", function() {
+        var job = new ConditionJob({when: 'false'}, {}, graphId);
+        job._run();
+        return job._deferred.should.eventually.be.rejectedWith('condition evaluated to false');
```
yyscamper: The `eventually` should be put together with `reject`, please see this issue for detail: https://github.com/RackHD/RackHD/issues/54 , I tend to think this should be a issue for Chai-as-Promise, I will submit a issue for it. But before it is fixed, we should avoid doing like this.

yyscamper: The `eventually` should be put together with `reject`, please see this issue for detail: https://github.com/RackHD/RackHD/issues/54 , I tend to think this should be a issue for Chai-as-Promise, I will submit a issue for it. But before it is fixed, we should avoid doing like this.

yyscamper: The `eventually` should be put together with `reject`, please see this issue for detail: https://github.com/RackHD/RackHD/issues/54 , I tend to think this should be a issue for Chai-as-Promise, I will submit a issue for it. But before it is fixed, we should avoid doing like this.

yyscamper:
I think at least following comment needs be fixed before merge:
(1) `eventually` should not be together with `reject`

tannoa2:
+1

zyoung51:
test this please

lib/jobs/condition.js
benbp: @yyscamper the assert is in the constructor.

benbp:
👍

PR #154, opened 2016-04-06
Create task and parser for Dell perccli tool

pengz1:
This PR is to create task and parser for Dell perccli tool catalog function. Workflow for catalog will be created in C6320 SKU pack.


pengz1:
@RackHD/corecommitters

anhou:
:+1:

lib/task-data/tasks/catalog-perccli.js
```diff
@@ -0,0 +1,23 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Catalog perccli',
+    injectableName: 'Task.Catalog.perccli',
+    implementsTask: 'Task.Base.Linux.Catalog',
+    options: {
+        adapter: '0',
```
yyscamper: For a catalog job, it better to scan all adapters and record all adapters' catalog data, is there any difficult to implement like this?

pengz1: Show all commands should work. However most servers and storage only have one controller. Not sure if Ben have any comments here. @benbp

yyscamper:
Still :+1: since we only scan the first adapter using MegaRaid tool, if we can find a solution to get adapter's information will be better.

pengz1:
@benbp , looks you are the first one to create MegaRaid catalog, do you know why you use  "adapter: '0'" instead of  "adapter: 'all'"?

benbp:
@pengz1 sorry, I missed your first @ mention of me. I don't remember the exact reason for "adapter: '0'". I think I may have been working on some old hardware that didn't support 'all'? Or maybe I just didn't know about it? Or I wanted the catalog to be easier to parse? Who knows. Regardless, I don't see why we can't just use "adapter: 'all'" if it works, though I would double check the command parser with that output.

pengz1:
@benbp . Thanks for confirmation. It should have no problem with parser as the command output itself is JSON. We actually just stored those JSON output in catalog. However, as till now we don't have a platform w/ more than 1 controllers, I will keep the code as it is.

yyscamper:
Since storcli still only scan the adapter 0, so it's OK for perccli. We can add a follow up items to check whether we need scan all adapters.

PR #157, opened 2016-04-07
Fix run-sku-graph job to have a reference to its sub-graph ID. Add te…

benbp:
…sts for run graph jobs

Right now the run sku graph job is actually broken. It will run a sub-graph specified by SKU, but it won't hold the correct reference to that graph, and so the job won't track state change AMQP events from the graph (i.e. it will never complete).

I'm also adding tests for the run graph/run sku graph jobs since there weren't any.

I removed the assert statements in the `_subscribeGraphFinished` callback because if they actually trigger the job will just hang indefinitely since we never call `_done()` OR `_done(error)`.

This may be totally unrelated to https://github.com/RackHD/RackHD/issues/154, but it may provide some clues (or may not).

@RackHD/corecommitters @heckj @zyoung51 @VulpesArtificem

### DIFF
--- a/lib/jobs/run-sku-graph.js
+++ b/lib/jobs/run-sku-graph.js
@@ -10,9 +10,9 @@ di.annotate(runSkuGraphJobFactory, new di.Provide('Job.Graph.RunSku'));
     new di.Inject(
         'Job.Base',
         'Services.Waterline',
-        'Protocol.TaskGraphRunner',
         'Logger',
         'Assert',
+        'Constants',
         'uuid',
         'Util',
         'JobUtils.WorkflowTool'
@@ -21,9 +21,9 @@ di.annotate(runSkuGraphJobFactory, new di.Provide('Job.Graph.RunSku'));
 function runSkuGraphJobFactory(
     BaseJob,
     waterline,
-    taskGraphProtocol,
     Logger,
     assert,
+    Constants,
     uuid,
     util,
     workflowTool
@@ -41,7 +41,7 @@ function runSkuGraphJobFactory(
         RunSkuGraphJob.super_.call(this, logger, options, context, taskId);
 
         this.nodeId = this.options.nodeId;
-        assert.isMongoId(this.nodeId);
+        assert.string(this.nodeId);
 
         if (!this.options.instanceId) {
             this.options.instanceId = uuid.v4();
@@ -57,8 +57,7 @@ function runSkuGraphJobFactory(
         var self = this;
 
         this._subscribeGraphFinished(function(status) {
-            assert.string(status);
-            if (status === 'succeeded') {
+            if (status === Constants.Task.States.Succeeded) {
                 self._done();
             } else {
                 self._done(new Error("Graph " + self.graphId +
@@ -76,17 +75,19 @@ function runSkuGraphJobFactory(
                 return;
             }
             if (!node.sku) {
+                // It's okay if there is no SKU, it just means there is
+                // nothing to do.
                 self._done();
                 return;
             }
 
-            return waterline.skus.findOne({ id: node.sku })
+            return waterline.skus.needOne({ id: node.sku })
             .then(function(sku) {
-                if (!sku) {
-                    return;
-                }
                 var graphName = sku.discoveryGraphName;
                 var graphOptions = sku.discoveryGraphOptions || {};
+                // If we don't specify the instanceId in the options
+                // then we can't track when the graph completes
+                graphOptions.instanceId = self.graphId;
 
                 if (!graphName) {
                     self._done();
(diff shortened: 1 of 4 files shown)


lib/jobs/run-sku-graph.js
```diff
@@ -41,7 +41,7 @@ function runSkuGraphJobFactory(
         RunSkuGraphJob.super_.call(this, logger, options, context, taskId);
 
         this.nodeId = this.options.nodeId;
-        assert.isMongoId(this.nodeId);
+        assert.string(this.nodeId);
```
benbp: Doing this so we can start moving away from mongo-specific string types (in favor of more flexible IDs that work cross-db).

spec/lib/jobs/run-sku-graph-spec.js
```diff
@@ -0,0 +1,162 @@
...
+
+    it('should noop if there is no sku', function() {
+        delete fakeNode.sku;
+        waterline.nodes.findByIdentifier.resolves(fakeNode);
+        var job = new RunSkuGraphJob({ nodeId: fakeNode.id }, { target: fakeNode.id }, uuid.v4());
+        job._run();
+
+        // The assertion here is that the job promise should just be resolved
+        // without having to trigger the _subscribeGraphFinished callback.
+        return job._deferred;
```
jlongever: Should there be an explicit assertion with some message that makes the expectation clear on failure?  Same at ln 152. Just a thought.

jlongever:
👍

zyoung51:
+1

lib/jobs/run-sku-graph.js
zyoung51: Any thoughts on monkey patching in something like an assert.nodeid () method? May be nice to assert on something more specific than just a string

benbp: I like that. Would be trivial since we already monkey-patch assert https://github.com/RackHD/on-core/blob/master/lib/common/assert.js

spec/lib/jobs/run-sku-graph-spec.js
benbp: How do you mean @jlongever? Like catch the rejected promise and raise a new, but different error with a message that is basically what the comment says?

benbp: Oh, or like `return expect(job._deferred).to.be.fulfilled;` ?

jlongever: Yeah that's what i was thinking.  Wouldn't the test throw an "unhandled rejection" if the run throws and not necessarily fail?  Maybe I'm thinking of a different scenario.

lib/jobs/run-sku-graph.js
```diff
@@ -76,17 +75,23 @@ function runSkuGraphJobFactory(
...
+                // It's okay if there is no SKU, it just means there is
+                // nothing to do.
                 self._done();
                 return;
             }
 
             return waterline.skus.findOne({ id: node.sku })
             .then(function(sku) {
                 if (!sku) {
+                    self._done();
```
yyscamper: Shouldn't throw an error? Since sku is assigned to the node so the SKU document should have existed in the database.

benbp: If we throw an error, that would mean that all discovery graphs would fail against nodes without a SKU assigned, which I think is behavior we definitely don't want. One option would be to mark the run-sku-graph task in the SKU discovery graph with `ignoreFailure: true`, but then we would ignore legitimate failure scenarios as well.

benbp:
@jlongever updated the test assertions.

benbp:
@zyoung51 I threw a ticket in our backlog for the `isNodeId` method: https://www.pivotaltracker.com/story/show/117401369

jlongever:
Ugh, ^^ random failure.  test this please

yyscamper:
:+1:

lib/jobs/run-sku-graph.js
yyscamper: @benbp , I agree the discovery graph should not fail if no SKU is specified, but we have handled the no-sku case in this line: https://github.com/RackHD/on-tasks/pull/157/files#diff-a4a2ba55165c416b61a5d04fe65e0dcaL78. For above line that I commented, the node has been assigned a valid SKU, if SKU document doesn't exist, so how the SKU was assigned before?

benbp: Ah good point, I missed that. I will update.

benbp:
@yyscamper updated to fail on specified, but missing, sku document.

yyscamper:
thanks @benp, +1

PR #160, opened 2016-04-11
Load result file into context on failure, in case it contains error msg.

stevemendes:
When an Ansible script returns a failure, this job was failing to read the result file that the playbook wrote out and store the contents in the context.  It would be useful  to load the contents even on failure so the playbook can return an error message in the context, alerting the caller of why the playbook failed.

### DIFF
--- a/lib/jobs/ansible-job.js
+++ b/lib/jobs/ansible-job.js
@@ -42,7 +42,6 @@ function ansibleJobFactory(
         // file for Ansible playbook to write results to:
         this.outFileName = path.join(os.tmpdir(),'ansible-' + this.taskKey + '.out'); 
 
-
         if(options.vars === undefined) {
             this.extraVars = '{"ansibleResultFile":"' + this.outFileName + '"}';
         } else {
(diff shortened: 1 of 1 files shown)


jlongever:
Jenkins, ok to test

lib/jobs/ansible-job.js
```diff
@@ -107,11 +97,27 @@ function ansibleJobFactory(
                 logger.warning("Result file not found. " + err.message);
                 self._done();
             } else {
-                self._done(err);
+                loadResultFile(self.outFileName,self.context)
```
jlongever: I think you might want to return here: 
`return loadResultFile(self.outFileName,self.context);`

lib/jobs/ansible-job.js
```diff
@@ -107,11 +97,27 @@ function ansibleJobFactory(
...
             }
         });
     };
 
+    function loadResultFile(resultFile,thisContext) {
+        return fs.readFileAsync(resultFile, 'utf8')
+        .then(function(contents) {
+            assert.arrayOfString(thisContext.ansibleResultFile);
+            thisContext.ansibleResultFile.push(contents);
+            fs.unlink(resultFile, function(err) {
```
jlongever: same, probably want to return.

benbp: The unlink here I think is meant to be inconsequential to the success/failure of the job.

lib/jobs/ansible-job.js
```diff
@@ -107,11 +97,27 @@ function ansibleJobFactory(
                 logger.warning("Result file not found. " + err.message);
                 self._done();
             } else {
-                self._done(err);
+                loadResultFile(self.outFileName,self.context)
+                .then(function() {
+                    self._done(err);
+                });
```
benbp: Might be good to add a `.catch` on here as well, in case your assert statement fails in `loadResultFile`? Otherwise there's a possibility for silent failures.

stevemendes: thx.  Done.

lib/jobs/ansible-job.js
stevemendes: Thx.  Done.

jlongever:
+1

yyscamper:
:+1:

PR #162, opened 2016-04-18
Wipe nvram in flahsing Quanta BIOS task

iceiilin:
Partial fix of ODR 620 about failure in BIOS update. NVRAM structure might change among different BIOS versions. Change to /N to wipe nvram settings when updating.

### DIFF
--- a/lib/task-data/tasks/flash-ami.js
+++ b/lib/task-data/tasks/flash-ami.js
@@ -22,7 +22,7 @@ module.exports = {
             // to a file and then send that file back + parse it, we
             // should be okay.
             'cd /opt/ami;' + 'sudo ./afulnx_64 {{ options.file }}' +
-                ' /P /B /K /X /R /ME > /tmp/renasar-afu-flash.log;' +
+                ' /P /B /K /N /ME > /tmp/renasar-afu-flash.log;' +
                 'sleep 3;tail -n 100 /tmp/renasar-afu-flash.log'
         ]
     },


stuart-stanley:
@derrickostertag since he worked on this before, I believe. My concern: is the options file that is part of the payload a delta or a complete overwrite? If the former, this change would be a problem. If the latter, this is +1

zyoung51:
+1

@stuart-stanley @derrickostertag I found this was a required step when upgrading/downgrading between 3A14 and newer versions.  While this could be skipped when upgrading/downgrading between 17, 18 and 19, skipping it when upgrading from 14 to a newer version or downgrading back to 14 from a newer version resulted in bricking the Quanta.

yyscamper:
good :+1:

stuart-stanley:
Still wondering if the options.file is a delta or a complete-set. If the former, I'm not sure this is a fix. If the latter, +1 no problem....

derrickostertag:
If the nvram of the BIOS gets wiped, depending on the defaults settings, we may lose the ability to control or talk to the BMC until settings can be manually set. If this is what has to be done, we need to verify it and put in the release notes.

iceiilin:
@stuart-stanley options.file is rendered completely by user's payload. I am not sure if I understand your concerns here.

iceiilin:
@derrickostertag I downgrade BIOS from 3A18 to 3A16 with this wiping flag, and still can talk to BMC.

derrickostertag:
@iceiilin Good. And going from 3A14 to 3A16+ BIOS settings are the same? if so +1

stuart-stanley:
@iceiilin if the user is expecting the options.file that is downloaded to be a DELTA against the current config, wiping the nvram will reset a bunch of things they were NOT expecting to change.
If the options.file is always the *complete* set of config information, then there is no problem...

zyoung51:
@stuart-stanley FWIW, its the preservation of the settings that actually causes it to be bricked.  The quanta procedure wipes the nvram to avoid this condition.

A user needs to provide a full file or expect non-deterministic behavior. The nvram utility does a full dump and it is expected that the user edits the settings they want changed and then provides that full file back.

iceiilin:
test this please

PR #163, opened 2016-04-18
added sftp job/task/tests

VulpesArtificem:
adds sftp job and tasks as part of in band management

depends on https://github.com/RackHD/on-core/pull/116 
@RackHD/corecommitters

### DIFF
--- a/lib/jobs/sftp-job.js
+++ b/lib/jobs/sftp-job.js
@@ -5,7 +5,6 @@
 var di = require('di');
 module.exports = sftpJobFactory;
 
-var Client = require('ssh2').Client;
 di.annotate(sftpJobFactory, new di.Provide('Job.Sftp'));
 di.annotate(sftpJobFactory, new di.Inject(
     'Job.Base',
@@ -14,6 +13,7 @@ di.annotate(sftpJobFactory, new di.Inject(
     'Assert',
     'Promise',
     'Services.Encryption',
+    'ssh',
     'Services.Waterline'
 ));
 
@@ -24,6 +24,7 @@ function sftpJobFactory(
     assert,
     Promise,
     cryptService,
+    ssh,
     waterline
 ) {
     var logger = Logger.initialize(sftpJobFactory);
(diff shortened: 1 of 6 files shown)


lib/jobs/sftp-job.js
```diff
@@ -0,0 +1,119 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+module.exports = sftpJobFactory;
+
+var Client = require('ssh2').Client;
```
yyscamper: Since you add the `ssh2` into the injectableTable (see https://github.com/RackHD/on-tasks/pull/163/files#diff-168726dbe96b3ce427e7fedce31bb0bcR11), so you'd better to use it using `di`

lib/jobs/sftp-job.js
```diff
@@ -0,0 +1,119 @@
...
+
+        assert.string(this.context.target);
+        this.nodeId = this.context.target;
+        this.timeout = options.timeout;
+        this.fileSource = options.fileSource;
+        this.fileDestination = options.fileDestination;
+    }
+    util.inherits(SftpJob, BaseJob);
+
+    SftpJob.prototype._run = function run() {
```
yyscamper: Since the base-job has the function with the same name `run`, so I suggest changing to another name to avoid confusion.

lib/jobs/sftp-job.js
```diff
@@ -0,0 +1,119 @@
...
+        var self = this;
+        return waterline.nodes.needByIdentifier(this.context.target)
```
yyscamper: Use `self.nodeId` instead of `this.context.target`? If not, then the assignment of `self.nodeId` (https://github.com/RackHD/on-tasks/pull/163/files#diff-9c80485bef6f959152431eaa9b8160f9R37) will have no reference.

lib/jobs/sftp-job.js
```diff
@@ -0,0 +1,119 @@
...
+        .then(function(node) {
+                return self.runSftp(node.sshSettings, new Client());
+            })
+            .then(function() {
```
yyscamper: indent error, and same with below `catch` block

spec/lib/jobs/sftp-job-spec.js
```diff
@@ -0,0 +1,185 @@
...
+        afterEach(function() {
+            this.sandbox.restore();
+        });
+
+        it('should use node credentials to sftp a file',function() {
+            this.sandbox.stub(sftpJob, 'runSftp').resolves();
+            return sftpJob._run()
+            .then(function() {
+                expect(waterline.nodes.needByIdentifier).to.have.been.calledOnce;
+                expect(sftpJob.runSftp).to.have.been.calledWith();
```
yyscamper: Miss the arguments in the `calledWith`

lib/jobs/sftp-job.js
```diff
@@ -0,0 +1,119 @@
...
+            })
+            .catch(function(e) {
+                self._done(e);
+            });
+    };
+
+    SftpJob.prototype.runSftp = function(sshSettings, sshClient) {
+        var self = this;
+        return new Promise(function(resolve, reject) {
+            if(self.timeout) {
```
yyscamper: Please confirm the behavior if no timeout is specified, the SSH client will have its default timeout?

yyscamper:
:+1: after my comment is either fixed or clarified

yyscamper:
test this please

lib/jobs/sftp-job.js
```diff
@@ -0,0 +1,119 @@
...
+                if (hasErr) {
+                    logger.error("Failure transferring file", {
+                        source: self.fileSource,
+                        destination: self.fileDestination,
+                        host: sshSettings.host
+                    });
+
+                    reject(new Error(
+                            "Encountered a failure transferring "+self.fileSource+
+                            "to "+self.fileDestination+"on remote host"+ sshSettings.host
```
anhou: only support remote as destination for upload, don't support local as destination for download?

lib/jobs/sftp-job.js
VulpesArtificem: yes, there's a default timeout of 20 seconds for ssh handshake and keep alive packets are disabled by default.

lib/jobs/sftp-job.js
VulpesArtificem: yes, we only upload files from the server to nodes.

lib/jobs/sftp-job.js
yyscamper: OK, thanks. Then it will be better to move the default timeout to RackHD code, so other user doesn't need to search to `ssh` library to check its default behavior.

VulpesArtificem:
test this please

VulpesArtificem:
test this please

lib/jobs/sftp-job.js
benbp: Right now there's no need/request to download files via sftp, so starting small with just uploads (which is needed soon).

benbp:
👍

yyscamper:
:+1:

lib/jobs/sftp-job.js
anhou: @VulpesArtificem @benbp gotta, :+1:

PR #165, opened 2016-04-20
run sku-specific-graph from skupack

yyscamper:
This is a portion of code about this proposal: https://groups.google.com/forum/#!topic/rackhd/nzmsIM6LF3Y

In summary, I want to move all SKU specific information to skupack and keep the SKU model more clean, below is an example of how to specify the sku specific graph in sku's config.json:
```json
{
    "rules": [...],
    "skuConfig": {
        "postDiscoveryGraph": {
            "name": "graph.my.testing",
            "options": {
                   "defaults": {
                         "foo": "bar"
                   }
              }
        }
    }
}
```

For back-compatible, my code still support defining the SKU specific graph via SKU document, but I marked it as deprecated.
 
There will be more follow-up PRs to eventually achieve this goal:
1. Rearrange the discovery tasks (see my google group topic)
2. Modify the Dell SKU pack, Move the dell perccli catalog into this SKU specific graph, so it can be automatically triggered after the Dell node is discovered.
3. Update related document.

@RackHD/corecommitters @zyoung51 @pengz1 @iceiilin @WangWinson @cgx027 @sunnyqianzhang

### DIFF
--- a/lib/jobs/run-sku-graph.js
+++ b/lib/jobs/run-sku-graph.js
@@ -10,23 +10,27 @@ di.annotate(runSkuGraphJobFactory, new di.Provide('Job.Graph.RunSku'));
     new di.Inject(
         'Job.Base',
         'Services.Waterline',
+        'Services.Environment',
         'Logger',
         'Assert',
         'Constants',
         'uuid',
         'Util',
-        'JobUtils.WorkflowTool'
+        'JobUtils.WorkflowTool',
+        '_'
     )
 );
 function runSkuGraphJobFactory(
     BaseJob,
     waterline,
+    env,
     Logger,
     assert,
     Constants,
     uuid,
     util,
-    workflowTool
+    workflowTool,
+    _
 ) {
     var logger = Logger.initialize(runSkuGraphJobFactory);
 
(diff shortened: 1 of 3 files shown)


zyoung51:
@yyscamper there is a story to move the config.json contents into the sku model which is mutually exclusive with this.

yyscamper:
@zyoung51 :
Ahh! If so, seems this PR need be retired.
Could you please share the history why you move the config.json to SKU model? As for me, the SKU model is bound with database, it's not so flexible because any new config entry will lead to the database schema change.
Meanwhile, could you please add me in the cc list for future SKU pack discussion, since a lot work in Shanghai will leverage the SKU pack.

amymullins:
Link to the story: https://www.pivotaltracker.com/projects/1492890/stories/117656885
Captured out of a thread between @zyoung51 , @benbp  and @rolandpoulter

yyscamper:
Thanks @amymullins , I'm appreciated if you can keep my team be notified for future skupack change. Meanwhile, I will spare some time to scan your backlog in future.

yyscamper:
Close this PR, since my work is in a wrong direction

PR #167, opened 2016-04-20
Add waiting time for the new bmc to take effect after flashing quanta

iceiilin:
From my test, we need to wait for secondes before the new BMC FW take effect after updating. If catalog BMC task follows immediately after flashing, it might fail in some entries.
Sleep for 90s after flashing BMC FW, which is suggested in BMC update script provided by Quanta.

### DIFF
--- a/lib/task-data/tasks/flash-quanta-bmc.js
+++ b/lib/task-data/tasks/flash-quanta-bmc.js
@@ -17,6 +17,9 @@ module.exports = {
             // Flash files
             'sudo /opt/socflash/socflash_x64 -s option=x ' +
                 'flashtype=2 if={{ options.downloadDir }}/{{ options.file }}',
+            // Wait for the new BMC to take effect, suggested in script provided by Quanta
+            // Otherwise following bmc related tasks (if any) might possibly fail
+            'sleep 90'
         ]
     },
     properties: {


iceiilin:
test this please

jlongever:
known random jenkins failures,  test this please

yyscamper:
@iceiilin : could you please add the PR description as comment into the task definition, so anyone can see the history why we add 90 seconds delay.

Anyway, for this PR, it is still :+1:

lib/task-data/tasks/flash-quanta-bmc.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
             // Flash files
             'sudo /opt/socflash/socflash_x64 -s option=x ' +
                 'flashtype=2 if={{ options.downloadDir }}/{{ options.file }}',
+            // Wait for the new version to take effect
```
anhou: wait for the new version? could you help refine the annotations ?

zyoung51:
Can you poll for the mc info to be updated rather than doing a hardcoded sleep?  Adding @uppalk1 as we need to make a similar change to the skupack implementation.

pengz1:
+1

iceiilin:
@zyoung51 @uppalk1 I wrote a script about polling mc info today, but I prefer to leave this PR as it is now and add this script into skupack.

I would not prefer to add it into templates/ so that linux command task can download and run it after running the update command, since there are already many templates  without clear catalogs and I don't want confuse others with this one. I think put this script in skupack is a good idea to make clear structure.

I paste my scripts here for your reference. It works in both T41 and D51.
```
#!/bin/bash
counter=90
until [ $counter -le 0 ]; do
    sudo ipmitool mc info
    if [ $? == 0 ]; then
        exit 0
    fi
    sleep 1
    let counter-=1
    echo ${counter}
done
exit 1
```

Any thoughts?

zyoung51:
@iceiilin Thanks for the update.  We will add that polling routine to the skupack flash_bmc script as part of the current enhancements to preserve the lan settings.

PR #168, opened 2016-04-21
added copykey job and tasks

VulpesArtificem:
more support for in-band management: added a job/tasks for copying an ssh public key to a node. Also refactored the sshExec method from ssh-job into command-util. 

should probably wait on https://github.com/RackHD/on-tasks/pull/150 to merge 
@benbp

PR #171, opened 2016-04-24
Validate ssh

VulpesArtificem:
Added a job/tasks for validating an inband connection via ssh and updating the ssh settings on a target node; additionally includes the copy-key job/tasks originally in https://github.com/RackHD/on-tasks/pull/168
@RackHD/corecommitters

### DIFF
--- a/lib/jobs/copy-publickey.js
+++ b/lib/jobs/copy-publickey.js
@@ -1,0 +1,71 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+module.exports = copyKeyJobFactory;
+
+di.annotate(copyKeyJobFactory, new di.Provide('Job.CopyKey'));
+di.annotate(copyKeyJobFactory, new di.Inject(
+    'Job.Base',
+    'Util',
+    'Logger',
+    'Assert',
+    'Promise',
+    'Services.Waterline',
+    'ssh',
+    'JobUtils.Commands'
+));
+
+function copyKeyJobFactory(
+    BaseJob,
+    util,
+    Logger,
+    assert,
+    Promise,
+    waterline,
+    ssh,
+    CommandUtil
+) {
+    var logger = Logger.initialize(copyKeyJobFactory);
+    var commandUtil;
+    function CopyKeyJob(options, context, taskId) {
+        CopyKeyJob.super_.call(this, logger, options, context, taskId);
+        assert.string(this.context.target);
+        this.nodeId = this.context.target;
+        commandUtil = new CommandUtil(this.nodeId);
+    }
+    util.inherits(CopyKeyJob, BaseJob);
+
+    CopyKeyJob.prototype._run = function run() {
+        var self = this;
+        return waterline.nodes.needByIdentifier(self.nodeId)
+        .then(function(node) {
+            var commands = [
+                {cmd: 'mkdir -p .ssh'},
+                {cmd: 'echo '+node.sshSettings.publicKey+' >> .ssh/authorized_keys'}
+            ];
+            return Promise.map(commands, function(commandData) {
+                return commandUtil.sshExec(
+                    commandData, node.sshSettings, new ssh.Client()
+                );
+            });
+        })
+        .then(function(result) {
+            logger.debug("Received copykey result", {
+                stdout: result.stdout,
+                stderr: result.stderr
+            });
+        })
+        .then(function() {
+            self._done();
+        })
+        .catch(function(err) {
+            logger.debug("Error copying public key", {error: err});
+            self._done(err);
+        });
+    };
+
+    return CopyKeyJob;
+}
+
--- a/lib/jobs/copy-publickey.js
+++ b/lib/jobs/copy-publickey.js
@@ -5,7 +5,7 @@
 var di = require('di');
 module.exports = copyKeyJobFactory;
 
-di.annotate(copyKeyJobFactory, new di.Provide('Job.CopyKey'));
+di.annotate(copyKeyJobFactory, new di.Provide('Job.CopySshKey'));
 di.annotate(copyKeyJobFactory, new di.Inject(
     'Job.Base',
     'Util',
@@ -28,12 +28,18 @@ function copyKeyJobFactory(
     CommandUtil
 ) {
     var logger = Logger.initialize(copyKeyJobFactory);
-    var commandUtil;
+
+    /**
+     * A job to copy an ssh public key from a node document's sshSettings field
+     * to .ssh/authorized keys on the node
+     *
+     * @constructor
+     */
     function CopyKeyJob(options, context, taskId) {
         CopyKeyJob.super_.call(this, logger, options, context, taskId);
         assert.string(this.context.target);
         this.nodeId = this.context.target;
-        commandUtil = new CommandUtil(this.nodeId);
+        this.commandUtil = new CommandUtil(this.nodeId);
     }
     util.inherits(CopyKeyJob, BaseJob);
 
@@ -41,27 +47,22 @@ function copyKeyJobFactory(
         var self = this;
         return waterline.nodes.needByIdentifier(self.nodeId)
         .then(function(node) {
+            assert.string(node.sshSettings.publicKey);
             var commands = [
                 {cmd: 'mkdir -p .ssh'},
                 {cmd: 'echo '+node.sshSettings.publicKey+' >> .ssh/authorized_keys'}
             ];
-            return Promise.map(commands, function(commandData) {
-                return commandUtil.sshExec(
+            return Promise.each(commands, function(commandData) {
+                return self.commandUtil.sshExec(
                     commandData, node.sshSettings, new ssh.Client()
                 );
-            });
-        })
-        .then(function(result) {
-            logger.debug("Received copykey result", {
-                stdout: result.stdout,
-                stderr: result.stderr
             });
         })
         .then(function() {
             self._done();
         })
         .catch(function(err) {
-            logger.debug("Error copying public key", {error: err});
+            logger.error("Error copying public key", {error: err});
             self._done(err);
         });
     };
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -54,6 +54,7 @@ function installOsJobFactory(
         this._validateOptions();
         this._convertOptions();
         this._encryptPassword();
+        this._provideUserCredentials();
     }
     util.inherits(InstallOsJob, BaseJob);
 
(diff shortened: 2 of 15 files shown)


lib/jobs/copy-publickey.js
```diff
@@ -0,0 +1,71 @@
...
+            logger.debug("Received copykey result", {
+                stdout: result.stdout,
+                stderr: result.stderr
+            });
+        })
+        .then(function() {
+            self._done();
+        })
+        .catch(function(err) {
+            logger.debug("Error copying public key", {error: err});
```
heckj: shouldn't this be `logger.error`?

lib/jobs/copy-publickey.js
```diff
@@ -0,0 +1,71 @@
...
+        CopyKeyJob.super_.call(this, logger, options, context, taskId);
+        assert.string(this.context.target);
+        this.nodeId = this.context.target;
+        commandUtil = new CommandUtil(this.nodeId);
+    }
+    util.inherits(CopyKeyJob, BaseJob);
+
+    CopyKeyJob.prototype._run = function run() {
+        var self = this;
+        return waterline.nodes.needByIdentifier(self.nodeId)
```
heckj: do you want to include an `assert.string(node.sshSettings.publicKey)` in here?

lib/jobs/validate-ssh.js
```diff
@@ -0,0 +1,108 @@
...
+        this.nodeId = this.context.target;
+        assert.ok(options.users || this.context.users);
+        this.timeout = options.timeout || 10000;
+        this.retries = options.retries || 2;
+        this.backoffDelay = options.backoffDelay || 1000;
+        this.users = _.compact([].concat(options.users).concat(this.context.users));
+    }
+    util.inherits(ValidationJob, BaseJob);
+
+    ValidationJob.prototype._run = function run() {
```
heckj: can you include some light docs on what you're intending this job to enable - looks like "test to make sure the provided credentials work, and if so, set them into the node sshSettings". Is that accurate?

heckj:
:+1: with log message fix and some minimal docs outlining intention for the jobs

lib/jobs/copy-publickey.js
benbp: I think it just needs a rebase with https://github.com/RackHD/on-tasks/pull/168

lib/jobs/copy-publickey.js
benbp: @heckj makes sense to be defensive here, though also the sshSettings object is enforced by a schema on API upload (though not on creation by this job).

benbp:
👍 along with the docs recommendation, because the variable/file naming doesn't really illuminate fully the functionality of the code.

benbp:
test this please

anhou:
test this please

lib/jobs/copy-publickey.js
```diff
@@ -0,0 +1,71 @@
...
+    util,
+    Logger,
+    assert,
+    Promise,
+    waterline,
+    ssh,
+    CommandUtil
+) {
+    var logger = Logger.initialize(copyKeyJobFactory);
+    var commandUtil;
```
yyscamper: I just fixed a bug in this PR: https://github.com/RackHD/on-tasks/pull/172, the `commandUtil` should not be shared by all job instances. Please fix it as well.

lib/jobs/install-os.js
```diff
@@ -95,13 +95,21 @@ function installOsJobFactory(
...
-
         if (this.options.users) {
             _.forEach(this.options.users, function(user) {
                 assert.string(user.name);
                 assert.string(user.password);
                 assert.number(user.uid);
             });
+            this.context.users = this.options.users;
+        }
+        if (this.options.rootPassword || this.options.rootSshKey) {
```
yyscamper: These codes are not suitable to put here, they don't matched with the function name `_validateOptions`, I hope to move them to another new function, or at least move to `_convertOptions` if you don't like to create a new function.

lib/jobs/ssh-job.js
```diff
@@ -50,7 +46,9 @@ function sshJobFactory(
         return waterline.nodes.needByIdentifier(self.nodeId)
         .then(function(node) {
             return Promise.reduce(self.commands, function(results, commandData) {
-                return self.sshExec(commandData, node.sshSettings, new ssh.Client())
+                return commandUtil.sshExec(
```
yyscamper: You need tweak this code and below ones as this PR (https://github.com/RackHD/on-tasks/pull/172) has been merged for hotfix.

lib/jobs/install-os.js
yyscamper: Meanwhile, add a reminder to you, we plan to change the `sshKey` and `rootSshKey` to plural, this is the story (https://www.pivotaltracker.com/projects/1492892/stories/117029201). If your code can take consideration both the singular and plural cases, then we can reduce associated change when doing former story.

lib/jobs/validate-ssh.js
```diff
@@ -0,0 +1,108 @@
...
+) {
+    var logger = Logger.initialize(validationJobFactory);
+    function ValidationJob(options, context, taskId) {
+        ValidationJob.super_.call(this, logger, options, context, taskId);
+        assert.string(this.context.target);
+        this.nodeId = this.context.target;
+        assert.ok(options.users || this.context.users);
+        this.timeout = options.timeout || 10000;
+        this.retries = options.retries || 2;
+        this.backoffDelay = options.backoffDelay || 1000;
```
yyscamper: How about user set these variables to `0`, could you please confirm whether `0` is a valid input, if so, your code need take special handling for number `0`.

lib/task-data/base-tasks/copy-key.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+
+module.exports = {
+    friendlyName: 'Copy Key',
+    injectableName: 'Task.Base.Ssh.CopyKey',
+    runJob: 'Job.CopyKey',
```
yyscamper: It's better to name it as `Job.CopySshKey`

lib/task-data/base-tasks/copy-key.js
```diff
@@ -0,0 +1,13 @@
...
+    requiredOptions: [],
```
yyscamper: Since you have designed some assertion logic for some SSH options, so it's better to add them into the `requireOptions`, the same for `validate-ssh` base-task.

lib/utils/job-utils/command-util.js
```diff
@@ -104,6 +106,63 @@ function commandUtilFactory(
         });
      };
 
+    CommandUtil.prototype.sshExec = function(cmdObj, sshSettings, sshClient) {
+        return new Promise(function(resolve, reject) {
+            if(cmdObj.timeout) {
+                setTimeout(function() {
+                    var seconds = cmdObj.timeout / 1000;
```
yyscamper: Since the `seconds` is only used in the log message, Personally I don't think it's worth to do the division

lib/jobs/validate-ssh.js
```diff
@@ -0,0 +1,108 @@
...
+    ValidationJob.prototype._run = function run() {
+        var self = this;
+        return waterline.lookups.findByTerm(self.nodeId)
+        .then(function(lookups) {
+            return self.testCredentials(lookups, self.users, self.retries, self.backoffDelay);
+        })
+        .then(function(sshSettings) {
+            return waterline.nodes.updateByIdentifier(
+                self.nodeId,
+                {sshSettings: sshSettings}
```
yyscamper: I just want to confirm this update operation is what you want, since the job name says it is to do SSH validation, but the update of sshSettings doesn't match with the word `validation`.

yyscamper:
Two future comment below:

(1) please add a test case for your new change in install-os job.

(2) We had `linux-catalog` and `linux-command` for micorkernel, I also think it will more flexible to design the job 'copy ssh key` to a general one just as the `linux-command`, it aims to executed a set of shell commands. For the task of copying of ssh keys, we can leverage the template, below is some pseudocode:
```javascript
{
    friendlyName: 'Copy Public Key',
    injectableName: 'Task.Ssh.Command',
    implementsTask: 'Task.Base.Ssh.CopyKey',
    options: {
           command: 'mkdir -p .ssh && echo {{options.sshKey | context.sshKey}} >> .ssh/authorized_keys'
    },
    properties: {}
}
```

:+1: after all my comment is either fixed or clarified. The design of general SSH command job can be moved to another story if you agree.

VulpesArtificem:
@yyscamper regarding your #2 comment: That functionality already exists with the ssh-job and tasks. 
```
{
    friendlyName: 'Copy Public Key',
    injectableName: 'Task.Ssh.CopyPublicKey',
    implementsTask: 'Task.Base.Ssh',
    options: {
           command: 'mkdir -p .ssh && echo {{options.sshKey | context.sshKey}} >> .ssh/authorized_keys'
    },
    properties: {}
}
```
should do what you want. The reason for a separate copyKey job is to get the publicKey from the sshSettings field of a target node's database document.

lib/jobs/validate-ssh.js
VulpesArtificem: `0` is a valid option for all of those fields, it will disable the option.

lib/jobs/validate-ssh.js
VulpesArtificem: yes, once we find valid settings we also want to make them available on the node document.

lib/jobs/copy-publickey.js
```diff
@@ -0,0 +1,72 @@
...
+        this.nodeId = this.context.target;
+        this.commandUtil = new CommandUtil(this.nodeId);
+    }
+    util.inherits(CopyKeyJob, BaseJob);
+
+    CopyKeyJob.prototype._run = function run() {
+        var self = this;
+        return waterline.nodes.needByIdentifier(self.nodeId)
+        .then(function(node) {
+            assert.string(node.sshSettings.publicKey);
```
benbp: What if sshSettings does not exist?

lib/task-data/tasks/validate-ssh.js
```diff
@@ -0,0 +1,18 @@
...
+
+module.exports = {
+    friendlyName: 'Validate Ssh',
+    injectableName: 'Task.Ssh.Validation',
+    implementsTask: 'Task.Base.Ssh.Validation',
+    options: {
+        timeout: null,
+        retries: null,
+        backoffDelay: null,
+        users: null
```
benbp: Will you add the root variables here too, so a graph definition can fill them in with "defaults" ?

lib/jobs/validate-ssh.js
benbp: But what if someone wants no backoff delay I think is the question.

lib/task-data/tasks/validate-ssh.js
benbp: Or do it as a follow-on.

PR #174, opened 2016-04-27
fixed error where we update the node sshSettings with the wrong key

VulpesArtificem:
fix for a bug where validate-ssh updates the node with the wrong key which will cause http validation errors when we later try to access the document

@RackHD/corecommitters


benbp:
CC @heckj 

We'll also want a PR against the 1.2 release branch I'm assuming.

benbp:
👍

heckj:
👍 
@VulpesArtificem would you submit a secondary PR to the `release/1.2-branch` as well please?

PR #175, opened 2016-04-27
fixed error where we update the node sshSettings with the wrong key

VulpesArtificem:
fix for a bug where validate-ssh updates the node with the wrong key which will cause http validation errors when we later try to access the document

@RackHD/corecommitters

### DIFF
--- a/lib/jobs/validate-ssh.js
+++ b/lib/jobs/validate-ssh.js
@@ -104,7 +104,7 @@ function validationJobFactory(
             var conn = new ssh.Client();
             var sshSettings = {
                 host: addr,
-                username: credentials.name,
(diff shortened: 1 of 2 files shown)


heckj:
👍 to match #174

benbp:
👍

benbp:
@davidzuhn @johren okay to merge to 1.2 or should I wait? Don't know whether I'll inadvertently kick off cascading builds or whether we should wait for a merge set or something.

yyscamper:
:+1:

davidzuhn:
If it's ready to merge, please do.   It *will* kick off builds, and that's the point of CI.

PR #179, opened 2016-05-02
Issue103

zhangy42:
Link to the requirement:  https://github.com/RackHD/RackHD/issues/103

Sample tasks will need help to add. 

The Source code and its unit tests should be done.

Added Joel as it is redfish relevant

@benbp @DavidjohnBlodgett @jlongever


jlongever:
-1 
What is the reason for adding a superagent client task?  I would prefer to avoid adding superagent, HTTP client libraries are built into node.  There is also already an HTTP client utility available in redfish-tool.js:

https://github.com/RackHD/on-tasks/blob/master/lib/utils/job-utils/redfish-tool.js#L118

benbp:
@jlongever would it maybe make sense to rename/move the pieces of that tool to just be something like httpTool, so it's clearer that that generic utility is available?

jlongever:
@benbp Sure.

zhangy42:
@jlongever @benbp , Just confirm my understanding, you want me: 

1. Rename following things:
REST-tool.js -> HTTP-tool.js
RestFactory() -> HttpFactory()
RestTool() -> HttpTool()

2. Modify REST-job.js and all the spec scripts to make sure that they call above tool correctly. 

3. Keep the job script REST-job.js, RestJobFacotry(), RestRequest(), Rest() and RestJob().

zhangy42:
@jlongever, I checked your code in Refish (the part you showed in the link, please do correct me if there are other sources). I found that it actually does not meet the need in this case (issue 103: https://github.com/RackHD/RackHD/issues/103), because 1. the user want to authenticate through cookie/token (Detail are here https://www.emc.com/techpubs/vipr/authenticate_controller_rest_api-1.htm), so that we need to provide an editable header. 2. the user want to fill the body xml, so that we also need to keep 'Content-type' in header editable.

So my suggestion is that: we can make a new util called HTTP-tool.js and put both function into the util. Or you can use my current implementation (with header editable but you just put it 'Content-type: application/json').

jlongever:
@zhangy42 Maybe update the current HTTP client utility to take an optional header parameter?  I'd prefer not to duplicate functionality.

zhangy42:
We need to move the code out anyway, current http client (you referred) lives in a module about redfish. Currently, there does not have a .js file/module called HTTP client utility. So by anyway we need to create a new util called HTTP util (for example HTTP-client-tool.js), and  then we need to 1. change your redfish code to make sure that it calls correct HTTP util. 2. change my REST util that make sure it is calling correct HTTP util.

jlongever:
@zhangy42 The code originated in redfish-tool.js because the Redfish jobs was the only thing using the client.  As @benbp suggested,  we should just move the code from redfish-tool.js into http-client-tool.js (or something common),  add an option to set a custom header and wrapper the calls inside redfish-tool.js to minimize the changes to the redfish code.

zhangy42:
@jlongever Thank you Joel. In order to make sure that I can change the HTTP client tool in Redfish successfully. Do you know how to test/debug the Redfish component, is unit test good enough?

jlongever:
@zhangy42 Unit test probably won't be enough to fully verify the change.  We'll have to do some functional tests running Redfish.Discovery.Graph workflows against a Redfish endpoint.  Nothing in the integration test suite for this yet.

zhangy42:
Let me first make the change and probably I need some help for testing the change.

index.js
```diff
@@ -16,6 +16,9 @@ module.exports = {
         core.helper.simpleWrapper(
             core.helper.requireGlob(__dirname + '/lib/task-data/**/*.js'),
             'Task.taskLibrary'
-        )
+        ),
+         core.helper.requireWrapper('superagent', 'Superagent')
```
yyscamper: The `requireWrapper` always need to explicitly pass the directory name if this is not `on-core` repo, we had found issues due to this multiple times before. 
This is the correct usage:  https://github.com/RackHD/on-tasks/blob/master/index.js#L11

lib/jobs/REST-job.js
```diff
@@ -0,0 +1,197 @@
+#! /usr/bin/env node
```
yyscamper: This shebang need be removed, as this is not a js-script that can be executed.

lib/jobs/redfish-job.js
```diff
@@ -267,12 +269,75 @@ function redfishJobFactory(
             return data;
         });
     };
+
+    /**
+     * Check fabric service status and generate alert on status change
+     * @param redfish client utility
+     */
+    RedfishJob.prototype.fabricServiceDataAlert = function(redfish, currentData) {
```
jlongever: I think you need to rebase your branch, you're pushing commits already in master.
https://github.com/RackHD/on-tasks/blob/master/lib/jobs/redfish-job.js#L277

yyscamper:
@zhangy42 : 
(1) Please rebase the code, I see there is a lot code doesn't belong to your pull request
(2) Please compress your commit history, 100+ commit is sooooo big!

WangWinson:
=。= b
Just going to take a look what happens here, bug gave up when I see 100+ commits.

zhangy42:
Let me copy out my changes and make another branch and make another PR.

PR #182, opened 2016-05-05
Add ARP cache poller

jlongever:
- Poll /proc/net/arp and parse entry data.
- Update lookups with new, updated and removed ARP entries.
- unit-tests

@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @zyoung51 @geoff-reid

### DIFF
--- a/lib/jobs/arp-poller.js
+++ b/lib/jobs/arp-poller.js
@@ -1,0 +1,133 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = arpPollerFactory;
+di.annotate(arpPollerFactory, new di.Provide('Job.ArpPoller'));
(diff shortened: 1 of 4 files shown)


lib/jobs/arp-poller.js
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * @memberOf ArpPollerJob
+     */
+    ArpPollerJob.prototype._run = function() {
+        var self = this;
+        Rx.Observable.interval(self.pollIntervalMs)
+        .subscribe(
+            self.arpCacheHandler.bind(self),
+            function(error) {
+                logger.error('ARP Poller Error', {error:error});
```
benbp: Shouldn't we fail the job here too?

benbp: Actually, this will likely never get hit. Errors from the `onNext` subscribe callback don't get handled by `onError` :(

PR #183, opened 2016-05-05
added a parser for ip command

VulpesArtificem:
Added an ip addr parser to the command parser; also refactored updateLookups from linux-catalog-job to command util to make it available to the ssh-job. @RackHD/corecommitters @rolandpoulter @DavidjohnBlodgett

### DIFF
--- a/lib/utils/job-utils/command-parser.js
+++ b/lib/utils/job-utils/command-parser.js
@@ -1033,33 +1033,61 @@ function commandParserFactory(Logger, Promise, _) {
 
         try {
             var interfaces = _.transform(data.stdout.split('\n'), function(result, line) {
+                //divide the output of ip addr show; ip link show into an array of arrays
+                // where each sub-array represents the details of an interface and
+                // each element is one line of output
                 if (line.startsWith(':', 1)) {
+                    //if we're looking at the first line of a new interface
+                    // i.e. 1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc....
+                    // push an array containing that line into the result
                     result.push([line]);
                 } else {
+                    //otherwise push the line into the last interface array
                     result[result.length - 1].push(line);
                 }
-            });
+            }, []);
 
             var parsed = _.transform(interfaces, function(res, info) {
+                //iterate over the interface arrays to parse them into key value pairs
                 var nameLine = info[0].substring(3).split(': ');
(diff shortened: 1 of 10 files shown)


spec/lib/utils/job-utils/stdout-helper.js
```diff
@@ -3644,6 +3644,41 @@ module.exports.racadmJobStatusData = 'Security Alert: Certificate is invalid - s
                                     '----------------------------------------------------------\n' +
                                     '\n';
 
+module.exports.ipAddrLinkOutput = '1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default\n'+
```
benbp: In general we've started just moving the data out to another dedicated file, then loading it in through this file.

benbp:
👍 with a recommendation to move the test data into its own file.

lib/utils/job-utils/command-parser.js
```diff
@@ -1023,6 +1024,58 @@ function commandParserFactory(Logger, Promise, _) {
...
+        }
+
+        try {
+            var interfaces = _.transform(data.stdout.split('\n'), function(result, line) {
+                if (line.startsWith(':', 1)) {
+                    result.push([line]);
+                } else {
+                    result[result.length - 1].push(line);
+                }
+            });
```
yyscamper: Miss the initial accumulator value `[]`

lib/utils/job-utils/command-parser.js
```diff
@@ -1023,6 +1024,58 @@ function commandParserFactory(Logger, Promise, _) {
         }
     };
 
+    CommandParser.prototype[ip] = function(data) {
```
yyscamper: I recommend to add some sample data in each step during parsing, so that brings convenience for other user to read the code.

yyscamper:
:+1: after at least fix following comment:
- @benbp 's comment
- My comment "Miss the initial accumulator value []"

lib/utils/job-utils/command-parser.js
anhou: second from me. at least some steps could be well described.

lib/utils/job-utils/command-parser.js
VulpesArtificem: you don't need to add an explicit accumulator unless you want the accumulator to be a different type than the input object, but I will add one for clarity

PR #184, opened 2016-05-06
Fix task timeout, support schedulerOverrides. Miscellaneous code cleanup

benbp:
Resolves https://github.com/RackHD/RackHD/issues/197

Fixing task timeout functionality, and also supporting the old "schedulerOverrides" option still in use in most places. Also the unit tests weren't valid for testing this, so updating those as well.

@RackHD/CoreCommitters @vulpesartificem @stuart-stanley @hohene

### DIFF
--- a/lib/task.js
+++ b/lib/task.js
@@ -98,7 +98,6 @@ function factory(
         // hint to whatever is running the task when it has failed
         self.failedStates = [TaskStates.Failed, TaskStates.Timeout, TaskStates.Cancelled];
 
-
         self.renderContext = {
             server: Task.configCache,
             api: {
@@ -113,6 +112,16 @@ function factory(
         self.renderContext.api.base = self.renderContext.api.server + '/api/current';
         self.renderContext.api.files = self.renderContext.api.base + '/files';
         self.renderContext.api.nodes = self.renderContext.api.base + '/nodes';
+
+        var timeout = self.definition.options.$taskTimeout;
+        // Support schedulerOverrides for backwards graph definition compatibility
+        if (!timeout && self.definition.options.schedulerOverrides) {
+            timeout = self.definition.options.schedulerOverrides.timeout;
+        }
+        if (typeof timeout !== 'number') {
+            timeout = 24 * 60 * 60 * 1000;  // default to 24 hour timeout
+        }
+        self.$taskTimeout = timeout;
 
         return self;
     }
@@ -290,6 +299,9 @@ function factory(
                 return self;
             })
             .finally(function() {
+                if (self.timer) {
+                    clearTimeout(self.timer);
+                }
                 if (self.error) {
                     var data = {
                         error: self.error.toString(),
@@ -309,17 +321,20 @@ function factory(
         var self = this;
 
         self.instantiateJob();
+
         // TODO: it may better to define this in the database in the case that
         // a task runner crashes, and the task is re-run, the timeout clock
         // will reset. It's somewhat of a design question whether we want
         // timeouts to be for each task iteration or for the max time allowed period.
-        if (self.definition.options.timeout) {
-            assert.number(self.definition.options.timeout, "Task timeout value");
+        // If timeout is undefined, this evaluates to false as well.
+        if (self.$taskTimeout > 0) {
             self.timer = setTimeout(function() {
-                var _timeoutSecs = self.definition.options.timeout / 1000;
+                // This ends up cancelling the job, which causes the promise returned
+                // by the `self.job.run()` call below to then reject, which gets
+                // caught by the promise chain in `task.run()`.
                 self.cancel(new Errors.TaskTimeoutError(
-                    "Task did not complete within " + _timeoutSecs));
-            }, self.definition.options.timeout);
+                    "Task did not complete within " + self.$taskTimeout + "ms"));
+            }, self.$taskTimeout);
         }
         logger.info("Running task job.", {
             taskId: self.instanceId,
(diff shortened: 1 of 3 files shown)


spec/lib/task-spec.js
```diff
@@ -508,6 +502,64 @@ describe("Task", function () {
...
+            var definition = _.cloneDeep(noopDefinition);
+            definition.options.schedulerOverrides = { timeout: 1 };
+            var task = Task.create(definition, {}, {});
+            task.renderAll = sinon.stub().resolves();
+            sinon.spy(task, 'cancel');
+
+            return task.run().then(function() {
+                expect(task.state).to.equal('timeout');
+                expect(task.error).to.be.an.instanceof(Errors.TaskTimeoutError);
+                expect(task.error.message).to.equal("Task did not complete within 1ms");
```
hohene: is the message "within 1ms" correct?  or would it be within + timeout + ms ?

benbp: @hohene oh you mean like "Task did not complete within timeout 1ms" ?

benbp: Or do you mean referencing the test timeout variable I'm using rather than hardcoding it in the assertion string?

hohene: Yeah - you have 1ms hard coded in the message, thought it would be your 'timeout' variable in the message.  ie. "Task did not complete within " + timeout + "ms" (if it were python code.) :-)
or is it really 1ms?

benbp: @hohene I'm just hardcoding it in the test assertion against the error string, which will be generated via `{ timeout: 1 }` on line 549 (https://github.com/RackHD/on-tasks/pull/184#discussion-diff-62383049R549). In the non-test code it will produce the right message. Happy to change it here too though if you want.

hohene: Nope - looks good to me! thanks!
+1 on the rest!

lib/task.js
```diff
@@ -309,17 +309,24 @@ function factory(
         var self = this;
 
         self.instantiateJob();
+        // Support schedulerOverrides for backwards graph definition compatibility
+        var timeout = self.definition.options.timeout;
+        if (!timeout && self.definition.options.schedulerOverrides) {
+            timeout = self.definition.options.schedulerOverrides.timeout;
+        }
```
yyscamper: - How about setting a default timeout because we don't want a task to be hang forever? I remembered previously it is 24 hours.
- How about if the timeout is not a number? I see you removed the assertion.number for the timeout, what's the reason? Though the `undefined` works, but how about a string/an object or others

lib/task.js
```diff
@@ -309,17 +309,24 @@ function factory(
         var self = this;
 
         self.instantiateJob();
+        // Support schedulerOverrides for backwards graph definition compatibility
+        var timeout = self.definition.options.timeout;
```
yyscamper: I have a concern that there may be conflict in further if we choose the name `timeout`, this is a common word, it is likely to be an options for a job (such as the ssh command job).
So maybe we can copy the python design, use `__` as prefix if the variable is reserved for RackHD, so `timeout` ==> `__timeout`
How do you think?

spec/lib/task-spec.js
```diff
@@ -508,6 +502,64 @@ describe("Task", function () {
...
+            task.renderAll = sinon.stub().resolves();
+            sinon.spy(task, 'cancel');
+
+            return task.run().then(function() {
+                expect(task.state).to.equal('succeeded');
+                expect(task.error).to.equal(null);
+            });
+        });
+
+        it("should timeout with options.timeout", function() {
```
yyscamper: Sorry, I don't understand why this test case works well, because I don't see any Promise.delay, so how the timeout happens?
@benbp : could you please give some explanation.

spec/lib/task-spec.js
```diff
@@ -508,6 +502,64 @@ describe("Task", function () {
...
+
+            return task.run().then(function() {
+                expect(task.state).to.equal('timeout');
+                expect(task.error).to.be.an.instanceof(Errors.TaskTimeoutError);
+                expect(task.error.message).to.equal("Task did not complete within 1ms");
+                expect(task.cancel).to.have.been.calledOnce;
+            });
+        });
+    });
+
```
yyscamper: How about adding another test case to test the behavior that the input `timeout` is an invalid value, like `string` or `object` or `undefined`

spec/lib/task-spec.js
jlongever: I think @benbp is just making sure when the passed in timeout equals 1ms that the state equals `"timeout"` with the correct error message indicating the timed out value.  It's not a functional test that requires real delays?

heckj:
@benbp remember to make a dupe for the 1.2 branch when you've made any updates/rebased/etc.

spec/lib/task-spec.js
yyscamper: @jlongever , if so, it requires to stub the `setTimeout` according to the code. Maybe the 1ms is the critical value, it just make the test passed.

jlongever:
test this please

lib/task.js
benbp: @yyscamper thanks, I had forgotten about the default timeout. I'll add that in.

I checked comparisons for various types, they all evaluate to false:

```
> 'test' > 0 || [] > 0 || {} > 0
false
```

lib/task.js
benbp: Ah, good point @yyscamper. `__timeout` could work, or what about `taskTimeout`?

benbp: I think a descriptive name might be better than `__` as it seems a little odd to me to have a python-style internal variable in a JSON definition. Maybe that's just me though.

spec/lib/task-spec.js
benbp: @yyscamper @jlongever I am using real delays, but not any that involve I/O. `Promise.delay` eventually gets called here: https://github.com/RackHD/on-tasks/blob/master/lib/jobs/noop-job.js#L37. I'm taking advantage of the fact that even during normal execution the setTimeout of 1ms will trigger before the job is started and completes (since `Promise.delay(0)` isn't synchonous). However, this could be more deterministic, so I'll update the test task definition to add a 2ms delay to the job, if that sounds alright.

spec/lib/task-spec.js
benbp: Will do. When we switch to JSON schema for task definitions then it will be easier to deal with these types of things too.

lib/task.js
benbp: @yyscamper or, as another suggestion along the lines of conventional "magic" values in javascript, what about `$timeout` or `$taskTimeout` ?

benbp:
@yyscamper updated for your comments. I used `$taskTimeout` but let me know if you'd prefer something other than that.

yyscamper:
@benbp ,$taskTimeout is better, more javascriptical :)
👍

VulpesArtificem:
👍

PR #185, opened 2016-05-09
Add support for Api proxy

shanemsully:
In order to support running the RackHD on-taskgraph and on-http services on a host that does not have Layer-2 network access to the PXE network we need a host that does have access to the PXE network and can proxy requests to the southbound API endpoint over a management network.

The proxy server, nginx in my POC, adds headers to the requests coming from the PXE client that contain headers that will be used during profile, template, and task rendering.

The headers are
X-Real-IP - the actual IP address of the PXE client
X-RackHD-api-proxy-ip - The IP address of the API proxy server
X-RackHD-api-proxy-port - The port of the API proxy server

The proxy server URL is added to the lookup record for the PXE client so that it can be used during task rendering since the original HTTP request is not available at that time.

This is one of three pull requests. There will also be subsequent changes to on-core and on-http to support this.

on-core PR: https://github.com/RackHD/on-core/pull/131
on-http PR: https://github.com/RackHD/on-http/pull/274

[test-coverage.tar.gz](https://github.com/RackHD/on-tasks/files/255464/test-coverage.tar.gz)

### DIFF
--- a/lib/jobs/run-graph.js
+++ b/lib/jobs/run-graph.js
@@ -58,6 +58,7 @@ function runGraphJobFactory(
         this.options.graphOptions.instanceId = this.graphId;
         this.graphTarget = this.options.graphOptions.target;
         this.domain = this.options.domain || Constants.Task.DefaultDomain;
+        this.proxy = context.proxy;
     }
     util.inherits(RunGraphJob, BaseJob);
 
@@ -79,7 +80,8 @@ function runGraphJobFactory(
             self.graphTarget,
             self.options.graphName,
             self.options.graphOptions,
-            self.domain
+            self.domain,
+            self.proxy
         ).catch(function(error) {
             logger.error("Error starting graph for task.", {
                 taskId: self.taskId,
--- a/lib/task.js
+++ b/lib/task.js
@@ -76,8 +76,9 @@ function factory(
         self.instanceId = taskOverrides.instanceId || uuid.v4();
         // Add convenience nodeId attribute to be used for rendering some task data option values
         if (_.has(context, 'target')) {
-            self.nodeId = context.target;
+            self.nodeId = context.target;     
         }
+        
         self.name = taskOverrides.name || self.definition.injectableName;
         self.friendlyName = taskOverrides.friendlyName || self.definition.friendlyName;
         self.waitingOn = taskOverrides.waitingOn || [];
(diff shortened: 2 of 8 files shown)


lib/jobs/run-graph.js
```diff
@@ -79,7 +80,8 @@ function runGraphJobFactory(
             self.graphTarget,
             self.options.graphName,
             self.options.graphOptions,
-            self.domain
+            self.domain,
+            self.proxy
```
benbp: Would be great if you could add the same updates to https://github.com/RackHD/on-tasks/blob/master/lib/jobs/run-sku-graph.js#L97 as well!

benbp:
👍 after an update to run-sku-graph.

jlongever:
test this please

lib/task.js
```diff
@@ -97,15 +98,20 @@ function factory(
...
         self.failedStates = [TaskStates.Failed, TaskStates.Timeout, TaskStates.Cancelled];
-
+   
+        var server;
+        if (_.has(context, 'proxy')) {
+            server = context.proxy;
+        } else {
+            server = 'http://%s:%s'.format(
+                Task.configCache.apiServerAddress,
+                Task.configCache.apiServerPort
```
yyscamper: I'm afraid these two configuration entries may be deprecated, because we have switch them to the southbound API endpoint, see the example configuration (http://rackhd.readthedocs.io/en/latest/packaging_and_deployment.html#rackhd-configuration) and http://rackhd.readthedocs.io/en/latest/packaging_and_deployment.html#setting-up-http-https-endpoint

I want to invite @WangWinson @cgx027 to confirm how to handle such configuration entries.

yyscamper: I just realized this is not a new problem that brought by this PR, so this comment is not gating for me. If @WangWinson or @cgx027 confirms this is a problem need be resolved, I can accept move this work to another PR.

cgx027: A typical endpoint look like below (in monorail.json or config.json). I would suggest to get the address and port from the endpoint with northbound-api-router. 

```
"httpEndpoints": [
        {
            "address": "0.0.0.0",
            "port": 8080,
            "httpsEnabled": false,
            "proxiesEnabled": true,
            "authEnabled": false,
            "routers": "northbound-api-router"
        },
        {
            "address": "172.31.128.1",
            "port": 9080,
            "httpsEnabled": false,
            "proxiesEnabled": true,
            "authEnabled": false,
            "routers": "southbound-api-router"
        }
    ]
```

Since apiServerAddress and apiServerPort will also work, I agree this is not a gating item.

WangWinson: `apiServerAddress` and `apiServerPort` still works. So it's OK here

yyscamper:
:+1:

yyscamper:
Jenkins: ok to test

PR #197, opened 2016-05-12
Add task to wait for a completion Uri trigger

zyoung51:
Add a task to wait for a Uri trigger to resolve an issue
with ESXi installation early completion.

Resolves https://github.com/RackHD/RackHD/issues/221

### DIFF
--- a/lib/jobs/wait-completion-uri.js
+++ b/lib/jobs/wait-completion-uri.js
@@ -1,0 +1,76 @@
+
+// Copyright 2016, EMC, Inc.
+'use strict';
+
+var di = require('di');
+
+module.exports = waitCompletionUri;
+di.annotate(waitCompletionUri, new di.Provide('Job.Wait.Completion.Uri'));
+    di.annotate(waitCompletionUri,
+    new di.Inject(
+        'Job.Base',
+        'Logger',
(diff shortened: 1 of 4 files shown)


yyscamper:
@zyoung51 :
Thanks for the enhancement for ESXi installation.
I think the 2nd completionUri is still not the absolutely completed point, so there is still some gap between this point and the SSH really works, as the ssh-validation lasts for only 100 seconds, do you think  this value is enough to cover the time gap bases on your testing? Whether we still need extend the ssh-validation waiting time?

yyscamper:
Still :+1: for this PR

zyoung51:
@yyscamper ODR-697 needs to be fixed independently of this.  Every OS does a reboot after the completionUri is retrieved, but none of the OS workflows have a delay between the OS installation task completing and the SSH task running.  Because of this, the SSH task reports a false positive 'OS is ready' event.

yyscamper:
@zyoung51 , I agree, that's an independent task

zyoung51:
@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @jlongever

zyoung51:
test this please

PR #205, opened 2016-05-13
add code to handle no default OBM settings

DavidjohnBlodgett:
resolves https://github.com/RackHD/RackHD/issues/35

Thanks @davequick for the assist.

@RackHD/corecommitters @heckj @tldavies @VulpesArtificem @rolandpoulter 

-DJ

### DIFF
--- a/lib/jobs/obm-control.js
+++ b/lib/jobs/obm-control.js
@@ -44,8 +44,6 @@ function obmControlJobFactory(
         assert.string(this.options.action, 'action');
         assert.ok(_.contains(_.methods(ObmService.prototype), this.options.action),
                 'OBM action is a known action');
-        // Defaults for this should have been set by the definition
-        assert.ok(this.options.obmServiceName, 'obm service name');
 
         this.nodeId = this.context.target || this.options.nodeId;
         assert.string(this.nodeId, 'nodeId');
@@ -64,18 +62,33 @@ function obmControlJobFactory(
             assert.ok(node, "Node should exist to run OBM command");
             assert.ok(node.obmSettings, "Node should have OBM settings to run OBM command");
 
-            var settings = _.find(node.obmSettings, function(setting) {
-                return setting ? setting.service === self.obmServiceName : false;
-            });
+            if (self.obmServiceName) {
+                self.settings = _.find(node.obmSettings, function(setting) {
+                    return setting ? setting.service === self.obmServiceName : false;
+                });
+                // throw error if the exact obm service doesn't exist
+                assert.ok(self.settings, "Node should have OBM settings for service: " +
+                    self.obmServiceName);
+            } else {
+                //find the default obm service
+                if(node.obmSettings.length === 0){
+                    throw new Error("No OBM service assigned to this node.");
+                }
+                if(node.obmSettings.length === 1) {
+                    self.settings = node.obmSettings[0];
+                    logger.debug("Defaulting to obmServiceName "+ self.settings.service +
+                        " for node because only one OBM service present.");
+                } else {
+                    throw new Error("More than one OBM service assigned to this node." +
+                        " (no default set for which to use)");
+                }
+            }
 
-            assert.ok(settings, "Node should have OBM settings for service: " +
-                self.obmServiceName);
-
-            var obmServiceFactory = injector.get(self.options.obmServiceName);
+            var obmServiceFactory = injector.get(self.settings.service);
             var obmService = ObmService.create(
-                self.nodeId, 
+                self.nodeId,
                 obmServiceFactory,
-                settings, 
+                self.settings,
                 self.options
             );
 
--- a/spec/lib/jobs/obm-control-spec.js
+++ b/spec/lib/jobs/obm-control-spec.js
@@ -29,6 +29,7 @@ describe("Job.Obm.Node", function () {
             helper.require('/lib/jobs/obm-control.js'),
             helper.require('/lib/services/base-obm-service.js'),
             helper.require('/lib/services/ipmi-obm-service.js'),
+            helper.require('/lib/services/noop-obm-service.js'),
             helper.require('/lib/services/obm-service.js'),
             helper.di.simpleWrapper(mockWaterline, 'Services.Waterline')
         ]);
@@ -131,6 +132,92 @@ describe("Job.Obm.Node", function () {
             });
         });
 
+        it('should fail if there is no OBM services and no default', function(){
+            // user is not passing any default obm service to use for this test...
+            var testOptions = {
+                action: 'reboot',
+                delay: 1,
+                retries: 1
+            };
+
+            job = new Job(testOptions, { target: '54da9d7bf33e0405c75f7111' }, uuid.v4());
+            job._subscribeActiveTaskExists = sinon.stub().resolves();
+            job.killObm = sinon.stub().resolves();
+            var node = {
+                obmSettings: []
+            };
+            mockWaterline.nodes.findByIdentifier.resolves(node);
+
+            return expect(job.run()).to.be.rejectedWith(Error,
+                 'No OBM service assigned to this node.');
+
+        });
+
+        it('should fail if there is more than one OBM service with no default', function(){
+          // user is not passing any default obm service to use for this test...
+          var testOptions = {
+              action: 'reboot',
+              delay: 1,
+              retries: 1
+          };
+
+          job = new Job(testOptions, { target: '54da9d7bf33e0405c75f7111' }, uuid.v4());
+          job._subscribeActiveTaskExists = sinon.stub().resolves();
+          job.killObm = sinon.stub().resolves();
+
+          var node = {
+              obmSettings: [
+                  {
+                      service: 'noop-obm-service',
+                      config: {}
+                  },
+                  {
+                      service: 'foo',
+                      config: {
+                          "user": "admin",
+                          "password": "admin",
+                          "host": "10.0.0.254"
+                      }
+                  }
+              ]
+          };
+          mockWaterline.nodes.findByIdentifier.resolves(node);
+
+          return expect(job.run()).to.be.rejectedWith(Error,
+               'More than one OBM service assigned to this node.' +
+               ' (no default set for which to use)');
+
+        });
+
+        it('should set default OBM setting if only one exists', function(){
+            // user is not passing any default obm service to use for this test...
+            var testOptions = {
+                action: 'reboot',
+                delay: 1,
+                retries: 1
+            };
+
+            job = new Job(testOptions, { target: '54da9d7bf33e0405c75f7111' }, uuid.v4());
+            job._subscribeActiveTaskExists = sinon.stub().resolves();
+            job.killObm = sinon.stub().resolves();
+
+            var node = {
+                obmSettings: [
+                    {
+                        service: 'noop-obm-service',
+                        config: {}
+                    }
+                ]
+            };
+            mockWaterline.nodes.findByIdentifier.resolves(node);
+
+            return job.run()
+            .then(function() {
+                expect(job.settings).to.equal(node.obmSettings[0]);
+            });
+        });
+
+
         it('should run an OBM command with nodeId specified in options', function() {
             var node = {
                 obmSettings: [
@@ -179,7 +266,7 @@ describe("Job.Obm.Node", function () {
             .then(function() {
                 var ipmiObmServiceFactory = helper.injector.get('ipmi-obm-service');
                 expect(ObmServiceSpy).to.have.been.calledWith(
-                    job.nodeId, ipmiObmServiceFactory, node.obmSettings[0], 
+                    job.nodeId, ipmiObmServiceFactory, node.obmSettings[0],
                     testOptions);
                 expect(ObmService.prototype.reboot).to.have.been.calledOnce;
             });
--- a/lib/task-data/tasks/clear-sel.js
+++ b/lib/task-data/tasks/clear-sel.js
@@ -7,9 +7,7 @@ module.exports = {
     injectableName: 'Task.Obm.Node.ClearSEL',
     implementsTask: 'Task.Base.Obm.Node',
     options: {
-        action: 'clearSEL',
-        obmServiceName: 'ipmi-obm-service'
+        action: 'clearSEL'
     },
-    properties: {}	
+    properties: {}
 };
-
--- a/lib/task-data/tasks/identify-off.js
+++ b/lib/task-data/tasks/identify-off.js
@@ -7,8 +7,7 @@ module.exports = {
     injectableName: 'Task.Obm.Node.IdentifyOff',
     implementsTask: 'Task.Base.Obm.Node',
     options: {
-        action: 'identifyOff',
-        obmServiceName: 'ipmi-obm-service'
+        action: 'identifyOff'
     },
     properties: {}
 };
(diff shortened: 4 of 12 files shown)


DavidjohnBlodgett:
@benbp when you get a minute... think you could give this some love?

lib/jobs/obm-control.js
```diff
@@ -64,18 +64,33 @@ function obmControlJobFactory(
...
 
-            var settings = _.find(node.obmSettings, function(setting) {
+            //var settings = _.find(node.obmSettings, function(setting) {
+            self.settings = _.find(node.obmSettings, function(setting) {
                 return setting ? setting.service === self.obmServiceName : false;
             });
 
-            assert.ok(settings, "Node should have OBM settings for service: " +
+            if(!self.settings){
+                console.log("I'm in the block");
```
yyscamper: These debug message should be removed, there is still multiple places contain such un-necessary message.

lib/jobs/obm-control.js
```diff
@@ -64,18 +64,33 @@ function obmControlJobFactory(
...
             assert.ok(node.obmSettings, "Node should have OBM settings to run OBM command");
 
-            var settings = _.find(node.obmSettings, function(setting) {
+            //var settings = _.find(node.obmSettings, function(setting) {
+            self.settings = _.find(node.obmSettings, function(setting) {
                 return setting ? setting.service === self.obmServiceName : false;
             });
 
-            assert.ok(settings, "Node should have OBM settings for service: " +
+            if(!self.settings){
```
yyscamper: I think it will be more reasonable to find the default obm-setting only if no explict obmServiceName is specified or set to `"auto"`.  Because some code logic actually don't want the default behavior.

The logic looks like:
```javascript
if (self.obmServiceName) {
   // throw error if the exact obm service doesn't exist
} else {
   //find the default obm service
}
```

lib/jobs/obm-control.js
```diff
@@ -64,18 +64,33 @@ function obmControlJobFactory(
...
+                console.log("I'm in the block");
+                logger.info("No settings found that match obmServiceName provided to job.");
+                if(node.obmSettings.length === 1) {
+                    console.log("I'm where I want to be!!!");
+                    self.settings = node.obmSettings[0];
+                    logger.info("Defaulting to obmServiceName "+ self.settings.service +
+                    " for node because only one OBM service present.");
+                }else if(node.obmSettings.length > 1){
```
yyscamper: Miss a condition that `node.obmSetting.length === 0`, it should throw an error.

spec/lib/jobs/obm-control-spec.js
```diff
@@ -131,6 +132,63 @@ describe("Job.Obm.Node", function () {
             });
         });
 
+        it('should fail if there is more then one OBM service', function(done){
```
yyscamper: typo: `more then` should be `more than`

spec/lib/jobs/obm-control-spec.js
```diff
@@ -131,6 +132,63 @@ describe("Job.Obm.Node", function () {
...
+                  expect(e).to.have.property('message').that.equals(
+                      'more than one OBM service assigned to this node,' +
+                      ' no default set for which to use'
+                  );
+                  done();
+              } catch (e) {
+                  done(e);
+              }
+          });
+        });
```
yyscamper: Actually, this code block can be much more simple:
```javascript
return expect(job.run()).to.be.rejectedWith(AssertionError, 'more than one OBM service assigned to this node, no default set for which to use');
```

spec/lib/jobs/obm-control-spec.js
```diff
@@ -131,6 +132,63 @@ describe("Job.Obm.Node", function () {
...
+                    }
+                ]
+            };
+            mockWaterline.nodes.findByIdentifier.resolves(node);
+
+            return job.run()
+            .then(function() {
+                // hmmm, how to verify settings has been set...
+                expect(job.settings).to.equal(node.obmSettings[0]);
+            });
```
yyscamper: No a bug here, just to reply to your comment:

Previous, we tested whether the correct obm service has been set by judging whether the correct obmServiceFatory function is called, the code snippt is as following:
```javascript
            return job.run()
            .then(function() {
                var ipmiObmServiceFactory = helper.injector.get('ipmi-obm-service');
                expect(ObmServiceSpy).to.have.been.calledWith(
                    job.nodeId, ipmiObmServiceFactory, node.obmSettings[0],
                    testOptions);
            });
```

Of course, you choose to expose the result obm setting to job's prototype is an alternative simpler way.

lib/jobs/obm-control.js
DavidjohnBlodgett: nice catch, I thought I had removed them.

lib/jobs/obm-control.js
DavidjohnBlodgett: @yyscamper does this case, you brought up, not get handled by the assert found just below, on line 81?

spec/lib/jobs/obm-control-spec.js
DavidjohnBlodgett: understood, fixed in next push

spec/lib/jobs/obm-control-spec.js
DavidjohnBlodgett: thanks for the feedback, I tried implementing this, but it seems to fail, i.e. not return the expected behavior... I would not rule out that I did not implement it correctly though ;-) I'll keep playing with it.

spec/lib/jobs/obm-control-spec.js
DavidjohnBlodgett: haha, you caught me... ya I was having difficulty wrapping my brain around a valid way to test this... and so resulted in just exposing the value I wanted to check as the only way I could think of to make it readable... do you, or anyone else object to me exposing that private variable? Should I test this more like you described above?

lib/jobs/obm-control.js
DavidjohnBlodgett: I think I follow you here, thanks for the feedback! Just curious, could you help me understand what the case would be that we don't want the default behavior?

lib/jobs/obm-control.js
```diff
@@ -64,18 +64,33 @@ function obmControlJobFactory(
             assert.ok(node, "Node should exist to run OBM command");
             assert.ok(node.obmSettings, "Node should have OBM settings to run OBM command");
 
-            var settings = _.find(node.obmSettings, function(setting) {
+            //var settings = _.find(node.obmSettings, function(setting) {
```
anhou: remove this line if it's not used

spec/lib/jobs/obm-control-spec.js
yyscamper: I think either way is OK to me, export the `setting` is reasonable to me.

spec/lib/jobs/obm-control-spec.js
yyscamper: OK, code is correct, You can keep with this.

lib/jobs/obm-control.js
yyscamper: If length ===0, line 81 will not be called, your code just miss this special condition. Change this line to `else` will work, of course you need to modify the error message to "no obm service or more than one obm services"

lib/jobs/obm-control.js
yyscamper: (1) In some situation, we are aware some obm services will not work, for example, snmp-obm-service works for ipdu/switch, ipmi-obm-service does't work, so we want to explicitly use that particular obm-service. 
(2) For passive node discovery, such as the discovery for management service via ipmi, the discovery for switch via snmp, only that particular obm-service is reliable. 
(3) As I know, only ipmi-obm-service supports set pxe boot, so if I want to set the pxe boot order, I should explicitly use the ipmi-obm-service, otherwise fails.

lib/jobs/obm-control.js
DavidjohnBlodgett: understood, fixed in next push.

spec/lib/jobs/obm-control-spec.js
DavidjohnBlodgett: ok, I got help from @VulpesArtificem to wrap my head around what I was doing wrong... and I updated the code to implement the changes you requested, fix will be in the next push.

spec/lib/jobs/obm-control-spec.js
DavidjohnBlodgett: ok, I'll keep it as is for now. Thanks!

DavidjohnBlodgett:
@yyscamper Hey Felix, wondering if you would just review the changes I made based on your suggestions. :D

lib/jobs/obm-control.js
```diff
@@ -64,18 +62,33 @@ function obmControlJobFactory(
...
+                assert.ok(self.settings, "Node should have OBM settings for service: " +
+                    self.obmServiceName);
+            } else {
+                //find the default obm service
+                if(node.obmSettings.length === 0){
+                    assert.fail(true,true, "No OBM service assigned to this node");
+                }else if(node.obmSettings.length === 1) {
+                    self.settings = node.obmSettings[0];
+                    logger.info("Defaulting to obmServiceName "+ self.settings.service +
+                    " for node because only one OBM service present.");
```
yyscamper: a minor indent error, the same for line 83

lib/jobs/obm-control.js
```diff
@@ -64,18 +62,33 @@ function obmControlJobFactory(
...
 
-            var settings = _.find(node.obmSettings, function(setting) {
+            self.settings = _.find(node.obmSettings, function(setting) {
                 return setting ? setting.service === self.obmServiceName : false;
             });
 
-            assert.ok(settings, "Node should have OBM settings for service: " +
-                self.obmServiceName);
+            if (self.obmServiceName) {
+                // throw error if the exact obm service doesn't exist
```
yyscamper: I think it's more reasonable to move line 65-67 here because you don't need to lookup obm-service if no explicit obmServiceName is specified.

lib/jobs/obm-control.js
```diff
@@ -64,18 +62,33 @@ function obmControlJobFactory(
...
+                    self.obmServiceName);
+            } else {
+                //find the default obm service
+                if(node.obmSettings.length === 0){
+                    assert.fail(true,true, "No OBM service assigned to this node");
+                }else if(node.obmSettings.length === 1) {
+                    self.settings = node.obmSettings[0];
+                    logger.info("Defaulting to obmServiceName "+ self.settings.service +
+                    " for node because only one OBM service present.");
+                }else if(node.obmSettings.length > 1){
```
yyscamper: not a bug, but I think `else` is enough here

yyscamper:
@DavidjohnBlodgett : thanks for your fix, the code is almost good.  I had some new comment, it's :+1: to me if you at least fix this one:
```
I think it's more reasonable to move line 65-67 here because you don't need to lookup obm-service if no explicit obmServiceName is specified.
```

yyscamper:
Meanwhile, @DavidjohnBlodgett, could you please squeeze the commit history as well?

lib/jobs/obm-control.js
DavidjohnBlodgett: understood, fix added to text push.

lib/jobs/obm-control.js
DavidjohnBlodgett: understood, fix added to next push.

lib/jobs/obm-control.js
DavidjohnBlodgett: ok, so I refactored this to use `else`, and in addition, folded in the use case for `node.obmSettings.length === 0` because I was finding I could not hit that branch of code with tests, due to the assert found on line 63, it seemed to always trigger when the property either does not exist or is an empty array.

DavidjohnBlodgett: actually I take back my comment about it triggering in both cases... just retested after realizing I had not re-introduced the array with zero elements test back into the mix... and discovered I was able to also hit that else statement... fixed in next push.

DavidjohnBlodgett:
@yyscamper ok, fixes in and commits squashed per request. :)

lib/jobs/obm-control.js
```diff
@@ -64,18 +62,31 @@ function obmControlJobFactory(
...
+                    return setting ? setting.service === self.obmServiceName : false;
+                });
+                // throw error if the exact obm service doesn't exist
+                assert.ok(self.settings, "Node should have OBM settings for service: " +
+                    self.obmServiceName);
+            } else {
+                //find the default obm service
+                if(node.obmSettings.length === 1) {
+                    self.settings = node.obmSettings[0];
+                    logger.info("Defaulting to obmServiceName "+ self.settings.service +
```
benbp: debug level seems more appropriate for this, but not a big deal either way.

lib/jobs/obm-control.js
```diff
@@ -64,18 +62,31 @@ function obmControlJobFactory(
...
+                        " for node because only one OBM service present.");
+                }else {
+                    assert.fail(true,true, "More than one OBM service assigned to this node," +
```
benbp: What is this? Do you mean to just throw an error here?

lib/jobs/obm-control.js
```diff
@@ -64,18 +62,31 @@ function obmControlJobFactory(
...
+                        " (no default set for which to use)" +
```
benbp: Also I'd think we have enough information to throw separate error messages depending on whether there are no services assigned, or multiple ones assigned, that way the user doesn't have to guess which one it is.

benbp:
@DavidjohnBlodgett I think you should also remove the 'ipmi-obm-service' default value for obmServiceName in all the obm task definitions, right? (e.g. https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/reboot.js#L11)

lib/jobs/obm-control.js
DavidjohnBlodgett: do you have a recommended level?

lib/jobs/obm-control.js
DavidjohnBlodgett: yes, I just did a quick search and this seemed like a simple enough way. Any suggestions for something different?

benbp: `throw new Error("More than one...<rest of message>");`

lib/jobs/obm-control.js
benbp: Since you're in there making other changes, `logger.debug`.

lib/jobs/obm-control.js
DavidjohnBlodgett: So if there is no obm settings defined, it actually is caught on line 63... this last case was a way for me to satisfy Felix's request to handle the edge case of having the obm property exist as an empty array... he recommended just wrapping them into a "catch all" error, and I choose it to simplify unit testing and code coverage. Should I break out this logic again and reimplement a unique error? @benbp @yyscamper

benbp: So shouldnt the last line of this error message just be removed then? @davidjohnblodgett

DavidjohnBlodgett:
@benbp @yyscamper pushed an updated (and squashed) commit implementing Ben's feed back.

yyscamper:
Thanks @DavidjohnBlodgett , it's :+1:
but still lack another approval, @RackHD/corecommitters

benbp:
👍

PR #206, opened 2016-05-15
Remove null IP set in ARP poller

jlongever:
Previous logic was incorrect.  Lookups should maintain the last seen
entry and not invalidate the IP on an incomplete ARP entry since we
don't intend to control the kernel's ARP cache life cycle.

@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2 @keedya


yyscamper:
+1

zyoung51:
+1

PR #211, opened 2016-05-17
Increase timeout on snmpbulkwalk to handle switches with more data

johren:
Some of the Cisco switches I was using failed during passive discovery because they would timeout doing the snmpbulkwalk.  It appears that the default timeout is 50 seconds when it took about 1m6s to perform the full bulk walk.  I increased the timeout to 120 seconds and the switches now discover successfully.

### DIFF
--- a/lib/utils/job-utils/net-snmp-tool.js
+++ b/lib/utils/job-utils/net-snmp-tool.js
@@ -35,7 +35,11 @@ function SnmpFactory(assert, parser, ChildProcess, Promise, _) {
         }
         var outputFormat = options.numericOutput ? '-Oqn' : '-Oq';
         // -Oq makes parsing easier by simplifying snmptool output
-        var args = [outputFormat, '-v2c', '-t120', '-c', this.community, this.host].concat(oid.split(' '));
+        if (!options.timeout && options.timeout !== 0) {
+            options.timeout = 120;
+        }
+        var args = [outputFormat, '-v2c', '-t' + options.timeout.toString(),
+                    '-c', this.community, this.host].concat(oid.split(' '));
         // for bulkget and bulkwalk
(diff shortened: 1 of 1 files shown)


benbp:
FYI @vulpesartificem

benbp:
:+1:

lib/utils/job-utils/net-snmp-tool.js
```diff
@@ -35,7 +35,7 @@ function SnmpFactory(assert, parser, ChildProcess, Promise, _) {
         }
         var outputFormat = options.numericOutput ? '-Oqn' : '-Oq';
         // -Oq makes parsing easier by simplifying snmptool output
-        var args = [outputFormat, '-v2c', '-c', this.community, this.host].concat(oid.split(' '));
+        var args = [outputFormat, '-v2c', '-t120', '-c', this.community, this.host].concat(oid.split(' '));
```
yyscamper: I think it's better to add the timeout in the `options` and let it default to 120 rather than hard-code. So it's flexible if other command want different timeout value.
```
if (!options.timeout && options.timeout !== 0) {
    options.timeout = 120;
}
var args = [outputFormat, '-v2c', '-t' + options.timeout.toString(), '-c', this.community, this.host].concat(oid.split(' '));
```

yyscamper:
Still :+1: with nice-to-have comment.

johren:
Good suggestion, @yyscamper.  I will test this out today and put up a change.

johren:
I added @yyscamper 's suggestion of putting timeout in the options.

yyscamper:
@johren , I also checked there was no existing unit-test for this file. We need improve this in future, I can accept no unit-test for your minor change.

PR #213, opened 2016-05-17
Add partitioning support for os installation

anhou:
temporarily only support CentOS, CentOS is a starting point, other Linux OS will be added
more description, please see docs PR: https://github.com/RackHD/docs/pull/278
related PR: https://github.com/RackHD/on-tasks/pull/285

also address ODR-654 https://hwjiraprd01.corp.emc.com/browse/ODR-654

@RackHD/corecommitters @RackHD/rackhd_dev

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -171,6 +171,36 @@ function installOsJobFactory(
                 }
             });
         }
+
+        if (this.options.installPartitions) {
+            _.forEach(this.options.installPartitions, function(partition) {
+                if(partition.mountPoint){
+                    assert.string(partition.mountPoint, "mountPoint must be a string");
+                } else {
+                    throw new Error('mountPoint is required');
+                }
+
+                if(partition.size){
+                    assert.string(partition.size, "size must be a string");
+                    if(isNaN(+partition.size) && partition.size !== "auto") {
+                        throw new Error('size must be a number string or "auto"');
+                    }
+                } else {
+                    throw new Error('size is required');
+                }
+
+                if (partition.fsType) {
+                    assert.string(partition.fsType, "fsType must be a string");
+                    if(partition.mountPoint === 'swap') {
+                        if(partition.fsType !== 'swap') {
+                            logger.warning("fsType should be 'swap' if mountPoint is 'swap'");
+                            // if fsType is not swap, correct it.
+                            partition.fsType = 'swap';
+                        }
+                    }
+                }
+            });
+        }
     };
 
     InstallOsJob.prototype._provideUserCredentials = function() {
--- a/spec/lib/jobs/install-os-job-spec.js
+++ b/spec/lib/jobs/install-os-job-spec.js
@@ -486,5 +486,48 @@ describe('Install OS Job', function () {
             expect(function() { job._validateOptions(); })
                 .to.throw(Error, 'Invalid ipv6 netmask.');
         });
+
+        it('should throw error when mountPoint is not specified', function() {
+            job.options.installPartitions = [{ size:'500', fsType:'ext3' }];
+            return expect(job._validateOptions.bind(job,{}))
+                   .to.throw('mountPoint is required');
+        });
+
+        it('should throw error when mountPoint is not a string', function() {
+            job.options.installPartitions = [{ mountPoint:{}, size:'500', fsType:'ext3' }];
+            return expect(job._validateOptions.bind(job,{}))
+                   .to.throw('mountPoint must be a string');
+        });
+
+        it('should throw error when size is not specified', function() {
+            job.options.installPartitions = [{ mountPoint:'/boot', fsType:'ext3' }];
+            return expect(job._validateOptions.bind(job,{}))
+                   .to.throw('size is required');
+        });
+
+        it('should throw error when size is not a string', function() {
+            job.options.installPartitions = [{ mountPoint:'/boot', size:{}, fsType:'ext3' }];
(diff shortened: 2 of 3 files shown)


lib/jobs/install-os.js
```diff
@@ -96,6 +96,23 @@ function installOsJobFactory(
                 }
             });
         }
+
+        if (this.options.installPartitions) {
+            _.forEach(this.options.installPartitions, function(partition) {
+                assert.string(partition.mountPoint);
+                assert.string(partition.size);
```
yyscamper: As I commented in the doc PR (https://github.com/RackHD/docs/pull/278), the `size` should be better to be a number.

spec/lib/jobs/install-os-job-spec.js
```diff
@@ -152,6 +152,38 @@ describe('Install OS Job', function () {
         );
     });
 
+    describe('test _validateOptions', function() {
```
yyscamper: Lack some negative test cases:
- it should throw AssertionError if `mountPoint` is not specified or not in correct type.
- it should throw AssertionError if `size` is not in correct type.
- it should throw AssertionError if `fsType` is not in correct type.

lib/jobs/install-os.js
anhou: size is not only a number, it also could be a 'auto' string.

yyscamper: I mean if the `size` is omitted, then you can default to auto mechanism, normally it is a number.

Of course, it is OK to explicitly support a "auto" string, but I think at least the `number` type should be support, because it is more straightforward.

spec/lib/jobs/install-os-job-spec.js
anhou: It seems that It also have more previous unittest debts in this function. anyway, let me add them all

lib/jobs/install-os.js
anhou: in my opinion. I don't think so. I don't want to let users or the document have more understanding for when  'size' is omitted. it's not straightforward.  I also don't think we provide both string/number at the same time. we have to understand more about the input.  it's not straightforward for quickly understanding.  either number or string, choose one. actually, I still have concerns about why provide more choices for 'installDisk' for this default sequences. I have to read more than 3 times to understand what it's happening.

lib/jobs/install-os.js
```diff
@@ -129,6 +129,23 @@ function installOsJobFactory(
                 }
             });
         }
+
+        if (this.options.installPartitions) {
+            _.forEach(this.options.installPartitions, function(partition) {
+                assert.string(partition.mountPoint, "mountPoint must be a string");
+                assert.string(partition.size, "size must be a string");
```
yyscamper: Seem need an extra assertion if `size` is not "auto", then it should be a number string.

spec/lib/jobs/install-os-job-spec.js
```diff
@@ -379,5 +379,28 @@ describe('Install OS Job', function () {
...
         });
         
+        it('should throw error when mountPoint is not a string', function() {
+            job.options.installPartitions = [{ mountPoint:{}, size:'500', fsType:'ext3' }];
+            return expect(job._validateOptions.bind(job,{}))
+                   .to.throw('mountPoint must be a string');
+        });
+
+        it('should throw error when size is not a string', function() {
+            job.options.installPartitions = [{ mountPoint:'/boot', size:{}, fsType:'ext3' }];
```
yyscamper: I would expect to test more invalid types, not just `{}` as user may input any kind of types. So we need array:
```javascript
[{}, 'foo', [1], null, undefined, NaN, 123].forEach(function(value) {
    job.options.installPartitions = [{ mountPoint:'/boot', size: value, fsType:'ext3' }];
    return expect(job._validateOptions.bind(job,{})).to.throw('size must be a string');
});
```
The same for other test cases.

anhou: @yyscamper  IMHO. I don't think so. this way is not increasing either line coverage or branch coverage. actually, the input such as  null, NAN, etc. is not testing our codes. it's testing assert-plus's 'assert.string()'. although it looks like it tests all kinds of cases. it's missing focus test for our codes. and it's not unit test's intention.

yyscamper: As my understanding, unit-test is to test a portion of code has expected behavior for every (as many as) possible input. The test coverage is just a kind of measurement, not the target. Even 100% coverage doesn't mean the unit-test is good enough to have taken every possible input into account.

Meanwhile, unit-test is to test the behavior on input, so it should not care the implementation detail. So in your case,  unit-test doesn't care you use `assert.string` or another assertion library or even you design the assertion by yourself. Unit-test just cares if I pass a non-string value, the validation should throw an error. Though we cannot list all non-string input, but at least we can test those common cases. If the `size` is from a calculation, maybe it ends up with `NaN`; if the `size` is unpurposely omitted, it is `undefined`;...

Anyway, this comment is not gating for me, just a recommendation to take as many as possible input into account .

anhou: Sorry, I disagree here. I think one of the key point for unittest case is  'unit'. and more for what' goot unit test case, what' bad unit test case, one reference is cycle complexity.

yyscamper:
:+1: with a nice-to-have comment.

anhou:
Updated with comments

iceiilin:
+1

PR #217, opened 2016-05-19
added tool for HTTP client and merged redfish tool with new HTTP client

zhangy42:
I added the people I know that are relevant. Please add if it is relevant to other people.

The original issue/request/requirement: https://github.com/RackHD/RackHD/issues/103

@benbp @jlongever @DavidjohnBlodgett

### DIFF
--- a/lib/jobs/REST-job.js
+++ b/lib/jobs/REST-job.js
@@ -1,0 +1,58 @@
+// Copyright 2016, EMC, Inc
+
+'use strict';
+
+var di = require('di');
+
+module.exports = RestJobFactory;
+di.annotate(RestJobFactory, new di.Provide('Job.Rest'));
+di.annotate(RestJobFactory, new di.Inject(
+    'Job.Base', 
+    'Logger', 
+    '_', 
+    'Assert', 
+    'Promise', 
+    'Util', 
+    'JobUtils.HttpTool'));
+
+function RestJobFactory(BaseJob, Logger, _, assert, Promise, util, HttpTool){
+    var logger = Logger.initialize(RestJobFactory);
+    
+    /**
+    * The interface that runs Rest Job from tasks
+    * @constructor
+    */
+    function RestJob(options, context, taskId){
+        var self = this;
+        self.options = options;
+        self.context = context;
+
+        RestJob.super_.call(self, logger, options, context, taskId);
+        self.restClient = new HttpTool();
+    }
+    
+    util.inherits(RestJob, BaseJob);
+    
+    RestJob.prototype._run = function run(){
+        var self = this;
+        logger.info("Runnging A Rest Call");
+        self.restClient.setupRequest(self.options)
+       .catch(function(err){
+            logger.error("Found error during setting up request.", {error: err});
+            self._done(err);
+        })
+        .then(self.restClient.runRequest()
+            .catch(function(err){
+                logger.error('Found error on running request.', {error: err});
+                self._done(err);
+            })            
+            .then(function(data){
+                self.context.restData = data;
+                self._done();
+            })
+        );
+
+    };
+
+    return RestJob;
+}
--- a/lib/utils/job-utils/HTTP-tool.js
+++ b/lib/utils/job-utils/HTTP-tool.js
@@ -1,0 +1,195 @@
+//Copyright 2016, EMC, Inc.
+
+/*jshint node: true*/
+/*jshint multistr: true*/ 
+
+'use strict';
+
+var di = require('di');
+
+module.exports = HttpFactory;
+
+di.annotate(HttpFactory, new di.Provide('JobUtils.HttpTool'));
+di.annotate(HttpFactory, new di.Inject(
+    'Assert',
+    'Promise',
+    '_'
+));
+
+function HttpFactory(assert, Promise, _) {
+    
+    /**
+    * Tool class that does HTTP methods
+    * @constructor
+    */
+    function HttpTool(){
+        this.settings = {};
+        
+        // ********** Helper Functions/Members **********
+        
+        /**
+        * Make sure that the method is a valid HTTP method
+        * @param {string} - testMethod: the intended method that for 
+        * http request.
+        * @return {boolean} - Whether the testmethod is valid
+        */
+        var validMethod = {'GET': true, 'PUT': true, 'POST':true, 
+            'DELETE':true, 'PATCH': true};
+        
+        this.isValidMethod = function(testMethod){
+            return (testMethod in validMethod);
+        };
+
+        this.isSettingValid = function(settings){
+            if (_.isEmpty(settings)) {
+                return false;
+            }
+            if ( ! ('url' in settings && 'method' in settings)) {
+                return false;
+            }
+            if (_.isEmpty(settings.url) || _.isEmpty(settings.method)) {
+                return false;
+            }
+            return true;
+        };
+    }
+
+    /**
+    * Set up the Request
+    * @param {object} settings - the object that provides the infomation 
+    * of all the works, example:
+    * {
+    *   url: "https://somewhat.website.com:some-port/some/path/to/file",
+    *   method: "GET"/"PUT"/"POST"/"DELETE"/"PATCH",
+    *   credential: {username: "foo", password: "bar"},
+    *   headers: {"token":"whatever-cookie-file"},
+    *   data: "Hello world",
+    *   verifySSL: false,
+    *   recvTimeoutMs: 2000
+    * }
+    * OR
+    * {
+    *   url: {
+    *       protocol: "https",
+    *       host: "somewhat.website.com",
+    *       port: "some-port",
+    *       path: "/some/path/to/file"
+    *   },
+    *   method: "GET"/"PUT"/"POST"/"DELETE"/"PATCH",
+    *   credential: {username: "foo", password: "bar"},
+    *   headers: {"token":"whatever-cookie-file"},
+    *   data: "Hello world",
+    *   verifySSL: false,
+    *   recvTimeoutMs: 2000
+    * }
+    *
+    * @return {Promise}
+    *   resolve() - do whatever next
+    *   reject(err) - do whatever you want with the err
+    */
+
+    HttpTool.prototype.setupRequest = function(settings) {
+        var self = this;
+        var err;
+        return new Promise(function(resolve, reject) {
+            if (! self.isSettingValid(settings)) {
+                err = new Error('Please provide at least url and'+
+                                     ' method to use HTTP tool!');
+                reject(err);
+            }
+            else {
+                if (self.isValidMethod(settings.method)) {
+                    self.settings = settings;
+                    resolve();
+                }
+                else {
+                    err = new Error('The method you provided is ' +
+                                        'not valid!');
+                    reject(err);
+                }
+            }});
+    };
+    
+    /**
+    * Run the request based on checked result
+    */ 
+    HttpTool.prototype.runRequest = function() {
+        var urlTool = require('url'),
+            httpTool,
+            self = this,
+            urlObject = {},
+            toolName = 'http';
+        
+        if (_.isEmpty(self.settings)) {
+            throw new Error('Request is not setup properly,'+
+                            ' please run setupRequest() first.');
+        }
+
+        // Parse the string into url options
+        if (typeof (self.settings.url) === 'string') {
+            urlObject = urlTool.parse(self.settings.url);
+        }
+        else {
+            urlObject = self.settings.url;
+        }
+
+        // protocole needs to be 'http:' and module is 'http'
+        if (urlObject.protocol.substr(-1) === ':') {
+            toolName = urlObject.protocol.substr(0, urlObject.protocol.length - 1);
+        }
+        else {
+            toolName = urlObject.protocol;
+            urlObject.protocol = urlObject.protocol + ':';
+        }
+        httpTool = require(toolName);
+
+        // set the rest parameters
+        urlObject.method = self.settings.method;
+        
+        if (! _.isEmpty(self.settings.headers)){
+            urlObject.headers = self.settings.headers;
+        }
+        if (! _.isEmpty(self.settings.credential)){
+            urlObject.auth = self.settings.credential.username +':'+
+                self.settings.credential.password;
+        }
+        urlObject.rejectUnauthorized = self.settings.verifySSL || false;
+        urlObject.recvTimeoutMs = self.settings.recvTimeoutMs;
+
+        return new Promise(function(resolve, reject) {
+            var request  = httpTool.request(urlObject, function(response){
+                var result = {
+                    httpVersion : response.httpVersion,
+                    httpStatusCode : response.statusCode,
+                    headers: response.headers,
+                    body : '',
+                    trailers: response.trailers
+                };
+                response.on('data', function(chunk){
+                    result.body += chunk;
+                });
+                response.on('end', function(){
+                    resolve(result);
+                });
+            });            
+            
+            if (urlObject.method === 'POST' || (
+                urlObject.method === 'PUT' || urlObject.method === 'PATCH'
+            )){
+                
+                if (_.isEmpty(self.settings.data)){
+                    self.settings.data = '';
+                }
+                request.write(self.settings.data);
+            }
+            
+            request.on('error', function(err){
+                reject(err);
+            });
+            
+            self.settings = {};
+            request.end();
+        });    
+    };
+    return HttpTool;
+}
--- a/lib/utils/job-utils/http-tool.js
+++ b/lib/utils/job-utils/http-tool.js
@@ -28,7 +28,7 @@ function HttpFactory(assert, Promise, _) {
         this.dataToWrite = '';
         // ********** Helper Functions/Members **********
         
-        this.validMethods = ['GET', 'PUT', 'POST', 'DELETE', 'PATCH'];
+        var validMethods = ['GET', 'PUT', 'POST', 'DELETE', 'PATCH'];
         
         /**
         * Make sure that settings has at least property of url and method.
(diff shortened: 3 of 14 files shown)


lib/utils/job-utils/HTTP-tool.js
```diff
@@ -0,0 +1,195 @@
...
+                                        'not valid!');
+                    reject(err);
+                }
+            }});
+    };
+    
+    /**
+    * Run the request based on checked result
+    */ 
+    HttpTool.prototype.runRequest = function() {
```
jlongever: There is a lot of setup work for each request.  Could you do a one-time setup that is stored in the httpTool instance and break out the actual request method?

lib/utils/job-utils/HTTP-tool.js
```diff
@@ -0,0 +1,195 @@
...
+            urlObject.headers = self.settings.headers;
+        }
+        if (! _.isEmpty(self.settings.credential)){
+            urlObject.auth = self.settings.credential.username +':'+
+                self.settings.credential.password;
+        }
+        urlObject.rejectUnauthorized = self.settings.verifySSL || false;
+        urlObject.recvTimeoutMs = self.settings.recvTimeoutMs;
+
+        return new Promise(function(resolve, reject) {
```
jlongever: Create separate method for this that reference the initialized urlObject.

lib/jobs/REST-job.js
```diff
@@ -0,0 +1,58 @@
...
+    
+    RestJob.prototype._run = function run(){
+        var self = this;
+        logger.info("Runnging A Rest Call");
+        self.restClient.setupRequest(self.options)
+       .catch(function(err){
+            logger.error("Found error during setting up request.", {error: err});
+            self._done(err);
+        })
+        .then(self.restClient.runRequest()
```
yyscamper: Invalid promise syntax, should be:
```javascript
.then(function() {
    return self.restClient.runRequest()
    .catch(...)
}
```

for lift the nested promise:
```javascript
.then(function() {
    return self.resetClient.runRequest();
})
.catch(function(err) {

})
....
```

lib/jobs/REST-job.js
```diff
@@ -0,0 +1,58 @@
...
+        self.restClient = new HttpTool();
+    }
+    
+    util.inherits(RestJob, BaseJob);
+    
+    RestJob.prototype._run = function run(){
+        var self = this;
+        logger.info("Runnging A Rest Call");
+        self.restClient.setupRequest(self.options)
+       .catch(function(err){
```
yyscamper: The catch will not let your Promise exit earlier,  the afterward `then` will still be executed with resolved `undefined` result. You should first chain `then` next `catch`.

lib/jobs/REST-job.js
```diff
@@ -0,0 +1,58 @@
...
+
+        RestJob.super_.call(self, logger, options, context, taskId);
+        self.restClient = new HttpTool();
+    }
+    
+    util.inherits(RestJob, BaseJob);
+    
+    RestJob.prototype._run = function run(){
+        var self = this;
+        logger.info("Runnging A Rest Call");
```
yyscamper: I think `logger.debug` is more reasonable.

lib/utils/job-utils/HTTP-tool.js
```diff
@@ -0,0 +1,195 @@
...
+        
+        // ********** Helper Functions/Members **********
+        
+        /**
+        * Make sure that the method is a valid HTTP method
+        * @param {string} - testMethod: the intended method that for 
+        * http request.
+        * @return {boolean} - Whether the testmethod is valid
+        */
+        var validMethod = {'GET': true, 'PUT': true, 'POST':true,
```
yyscamper: Define the `validMethod` as array is enough, why object?

lib/utils/job-utils/HTTP-tool.js
```diff
@@ -0,0 +1,195 @@
+//Copyright 2016, EMC, Inc.
```
yyscamper: As a conversion, the filename is all lower case, so change filename from `HTTP-tool.js` to `http-tool.js`

yyscamper:
The biggest problem to me is why reinvent a new wheel? A lot npm library can be used, why not to find a good one?

yyscamper:
:-1:  I think a lot question need be clarified and fixed.

zhangy42:
@yyscamper , @jlongever said that we should keep using default node default http request.

lib/utils/job-utils/HTTP-tool.js
zhangy42: I need a datastructure similar to "HashSet" in Java (with time complexity O(1) vs O(n)). And I find out that similar javascript practice of doing that is using a object and put them as keys.

yyscamper:
@zhangy42 : so your work is totally based on @jlongever 's previous work?

@jlongever : could you please explain more about why not using a mature external rest library?

jlongever:
@yyscamper  I didn't see a compelling reason to introduce another HTTP client library when I added the redfish-tool utility,  the native node HTTP library was sufficient.  What library did you have in mind?

jlongever:
@yyscamper here is the original PR where these discussions already took place:
https://github.com/RackHD/on-tasks/pull/179#issuecomment-216742401

yyscamper:
@jlongever : I agree the current simple implementation of http client can meet the requirement of redfish client, but as a **general** rest-job, I think following requirement will be likely to emerge in future:

- Authentication: I see @zhangy42 has mentioned this requirement in the closed PR.
- Various content type support: json/xml/yaml/...
- Follow redirection.
- Proxy support
- Multipart request
- ...

These requirement need a lot code if we choose to implement from scratch. But if we choose a mature external library, it will bring flexibility to enable those functionality in future and needless to re-test the redfish client code.

I think following libraries are the very famous ones (they have a lot stars ^^): 
- https://github.com/request/request
- http://unirest.io/nodejs.html
- https://github.com/danwrong/restler
- https://github.com/visionmedia/superagent

I don't have experience to use all these libraries, but if we need, we can do more search and practice to find the best one.

zhangy42:
We can probably keep current version and make another ticket of improving HTTP tool with some exiting 3rd party tools. Since current version can already meet the need of our customer.

zhangy42:
Since our problem is at "expandable", and current version meets the needs of our customer requirements, shall we keep current version of http tool and change to another better one when the case you mentioned occurs?

@yyscamper we used superagent in our testing, and that was my 1st draft to this issue. Probably superagent is the best tool and will bring least trouble for people to adapt if we need it in the future.

yyscamper:
@zhangy42: Sound good to me, let's first meet the basic requirement, so I'm not against not using an external library now, but code should be flexible to be replaced with an external library, because I don't want the the future change will lead to change redfish-client or other related modules again.

Next step, I will focus on reviewing your code itself.

lib/utils/job-utils/http-tool.js
```diff
@@ -0,0 +1,202 @@
...
+        
+        /**
+        * Make sure that settings has at least property of url and method.
+        * @return {boolean} whether settings is valid.
+        */ 
+        this.isSettingValid = function(settings){
+            if (_.isEmpty(settings)) {
+                return false;
+            }
+            if ( ! ('url' in settings && 'method' in settings)) {
```
yyscamper: Use `hasOwnProperty` here instead of `in`, below links explain why:
- http://stackoverflow.com/questions/13632999/if-key-in-object-or-ifobject-hasownpropertykey
- http://adripofjavascript.com/blog/drips/the-uses-of-in-vs-hasownproperty.html

lib/utils/job-utils/http-tool.js
```diff
@@ -0,0 +1,202 @@
...
+        var self = this;
+        var err;
+        return new Promise(function(resolve, reject) {
+            if (! self.isSettingValid(settings)) {
+                err = new Error('Please provide at least url and'+
+                                     ' method to use HTTP tool!');
+                reject(err);
+            }
+            else {
+                if (_.indexOf(self.validMethods, settings.method) > -1) {
```
yyscamper: The validation of `method` should belongs to the scope of function `isSettingValid`

lib/utils/job-utils/http-tool.js
```diff
@@ -0,0 +1,202 @@
...
+            if (_.isEmpty(settings.url) || _.isEmpty(settings.method)) {
+                return false;
+            }
+            return true;
+        };
+
+        /**
+        * Parse and convert setting into a urlObject that suitable for 
+        * http/https module in NodeJs.
+        * @return {object} - the objec that suitable for Node http/https module.
```
yyscamper: typo "objec"

lib/utils/job-utils/http-tool.js
```diff
@@ -0,0 +1,202 @@
...
+                urlObject.headers = settings.headers;
+            }
+            else {
+                urlObject.headers = {};
+            }
+            
+            urlObject.headers['Content-Length'] = 0;
+            if (settings.hasOwnProperty('data')){
+                this.dataToWrite = JSON.stringify(settings.data);
+                urlObject.headers['Content-Length'] = Buffer.byteLength(this.dataToWrite);
```
yyscamper: Since the data is JSON here, so it's better to set a Content-Type as json as well.

lib/utils/job-utils/http-tool.js
```diff
@@ -0,0 +1,202 @@
...
+            }
+            
+            if (! _.isEmpty(settings.credential)){
+                urlObject.auth = settings.credential.username +':'+
+                    settings.credential.password;
```
yyscamper: How about if either `username` or `password` is missing, I didn't see such validation logic.

lib/utils/job-utils/redfish-tool.js
```diff
@@ -50,105 +52,54 @@ function redfishToolFactory(
...
+        setups.credential = {username: self.settings.username, 
+            password: self.settings.password};
+        setups.verifySSl = self.settings.verifySSl || false;
+        setups.headers = {'Content-Type': 'application/json'};
+        setups.recvTimeoutMs = self.settings.recvTimeoutMs;
+        setups.data = data;
+
+        var http = new HttpTool();
+        
+        return new Promise(function(resolve){
```
yyscamper: You don't need to create a new Promise here, since `http.setupRequest` has already returned promise, you should could just return the promise chain.
If you need to promise chain resolves `responce`, you just need to change line#98 to `return response`

lib/utils/job-utils/http-tool.js
```diff
@@ -0,0 +1,202 @@
...
+    * Tool class that does HTTP methods
+    * @constructor
+    */
+    function HttpTool(){
+        this.settings = {};
+        this.urlObject = {};
+        this.dataToWrite = '';
+        // ********** Helper Functions/Members **********
+        
+        this.validMethods = ['GET', 'PUT', 'POST', 'DELETE', 'PATCH'];
```
yyscamper: This is better to be a class level variable rather  than instance level.

spec/lib/jobs/rest-job-spec.js
```diff
@@ -0,0 +1,101 @@
...
+    var testUrl = 'https://test.address.com:12345',
+        taskId = uuid.v4(),
+        options = {
+            url: testUrl + '/full/put',
+            method: 'PUT',
+            credential: {username:"foo", password:"bar"},
+            headers: {
+                "content-type": "application/json",
+                "some-token": "whatever-ssl-token"
+                },
```
yyscamper: indent error

spec/lib/jobs/rest-job-spec.js
```diff
@@ -0,0 +1,101 @@
...
+        .matchHeader('content-type', 'application/json')
+        .matchHeader('some-token', 'whatever-ssl-token')
+        .put('/full/put')
+        .basicAuth({user: 'foo', pass: 'bar'})
+        .reply(201, 'You are good');
+
+        context = {};
+        restJob = new RestJob(options, context, taskId);
+
+        restJob.run().then(function(){
```
yyscamper: The promise must be returned otherwise any assertion will not be captured by mocha.

spec/lib/jobs/rest-job-spec.js
```diff
@@ -0,0 +1,101 @@
...
+        });
+    });
+
+    it('Should reject on missing url', function(done){
+        options.url = null;
+
+        restJob = new RestJob(options, context, taskId);
+        var errMsg = 'Please provide at least url and method to use HTTP tool!';
+
+        expect(restJob.run()).to.be.rejectedWith(errMsg);
```
yyscamper: Promise need be returned.

spec/lib/jobs/rest-job-spec.js
```diff
@@ -0,0 +1,101 @@
...
+        done();
+    });
+
+    it('Should reject on missing method', function(done){
+        options.url = testUrl + '/get/good';
+        options.method = null;
+
+        restJob = new RestJob(options, context, taskId);
+        var err = 'Please provide at least url and method to use HTTP tool!';
+        expect(restJob.run()).to.eventually.be.rejectedWith(err);
```
yyscamper: `eventually` should not be chain with `rejectedWith`, please see this issue for reason: https://github.com/RackHD/RackHD/issues/54

spec/lib/jobs/rest-job-spec.js
```diff
@@ -0,0 +1,101 @@
...
+    });
+
+    it('Should reject on missing url', function(done){
+        options.url = null;
+
+        restJob = new RestJob(options, context, taskId);
+        var errMsg = 'Please provide at least url and method to use HTTP tool!';
+
+        expect(restJob.run()).to.be.rejectedWith(errMsg);
+        done();
```
yyscamper: The `done()` will be called before the above asserted is executed, because `restJob.run` is asynchronous

spec/lib/utils/job-utils/http-tool-spec.js
```diff
@@ -0,0 +1,178 @@
...
+        .basicAuth({user:'hello', pass:'world'})
+        .reply(200);
+
+        requestSettings.url = siteGen + '/basicAuth';
+        requestSettings.method = 'GET';
+        requestSettings.credential = {username:'hello', password:'world'};
+        // requestSettings.data = '{hello: world}';
+
+        httpTool.setupRequest(requestSettings)
+        .then(httpTool.runRequest()
```
yyscamper: Bad promise syntax. it should be put within a callback function:
```javascript
.then(function() {
    return httpTool.runRequest();
})
```

spec/lib/utils/job-utils/http-tool-spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+        requestSettings.url = siteGen + '/bad-get';
+        requestSettings.method = 'GET';
+        
+        httpTool.setupRequest(requestSettings)
+        .then(httpTool.runRequest()
+        .catch(function(err){
+            expect(err).to.have.property('status').to.equal(404);
+            done();
+        }));
```
yyscamper: since you are expecting the promise is rejected, so you need to explicitly fail the test case if the promise is fulfilled.
```javascript
return httpTool.setupRequest(requestSettings)
         .then(function() {
             return httpTool.runRequest();
         })
         .then(function() {
             done(new Error('Should never run here, otherwise the code has some bug.'));
         });
         .catch(function(err){
             expect(err).to.have.property('status').to.equal(404);
             done();
         }));
```

spec/lib/utils/job-utils/redfish-tool-spec.js
```diff
@@ -1,143 +1,101 @@
...
+                .to.be.calledOnce;    
+            return;    
+        })    
+        .then(function() {    
+            expect(redfishTool.settings).to.deep.equal({    
+                uri: 'http://fake'    
+            });    
+        });    
+    });    
+
```
yyscamper: Please remove these trailing spaces

yyscamper:
@zhangy42 : To test the promise in unit-testing, you need pay attention to following aspects:
(1) The promise need be returned, otherwise mocha will not catch the assertion error.
(2) If you use `done` callback, usually you need call both `done()` and `done(err)`, if you don't call `done(err)`, the mocha will catch a timeout error, but this is not expected.
(3) If you want to add assertion logic into `catch` block, remember to explicitly fail test if run to `then`, otherwise if the code goes to `then`, the mocha will give you a false success.
(4) The `eventually` should not be chain with `rejected` or `rejectedWith`.

zhangy42:
Thank you Felix!

spec/lib/utils/job-utils/http-tool-spec.js
yyscamper: if you want your style code, don't call the function, so the following is correct:
```javascript
.then(httpTool.runRequest)
```

zhangy42:
My code fix should finish now, please go and merge the code if you guys are happy with the code. Thank you!

jlongever:
@zhangy42 Just pulled down your changes and ran them against our Redfish endpoint and seeing workflow errors:

```
    waitingOn:
      51b32526-4982-4d7a-ba54-3213d55ed78d: succeeded
  4d352217-c74d-42cf-ad60-31ea5cb4b49d:
    error:
      """
        TypeError: Argument must be a string
            at HttpTool.setupUrlOptions (/home/jlongever/on-tasks/lib/utils/job-utils/http-tool.js:92:62)
            at /home/jlongever/on-tasks/lib/utils/job-utils/http-tool.js:178:39
            at tryCatcher (/home/jlongever/on-core/node_modules/bluebird/js/main/util.js:26:23)
            at Promise._resolveFromResolver (/home/jlongever/on-core/node_modules/bluebird/js/main/promise.js:480:31)
            at new Promise (/home/jlongever/on-core/node_modules/bluebird/js/main/promise.js:70:37)
            at HttpTool.HttpFactory.HttpTool.runRequest (/home/jlongever/on-tasks/lib/utils/job-utils/http-tool.js:176:16)
            at /home/jlongever/on-tasks/lib/utils/job-utils/redfish-tool.js:86:25
            at tryCatcher (/home/jlongever/on-core/node_modules/bluebird/js/main/util.js:26:23)
            at Promise._settlePromiseFromHandler (/home/jlongever/on-core/node_modules/bluebird/js/main/promise.js:507:31)
            at Promise._settlePromiseAt (/home/jlongever/on-core/node_modules/bluebird/js/main/promise.js:581:18)
            at Promise._settlePromiseAtPostResolution (/home/jlongever/on-core/node_modules/bluebird/js/main/promise.js:245:10)
            at Async._drainQueue (/home/jlongever/on-core/node_modules/bluebird/js/main/async.js:128:12)
            at Async._drainQueues (/home/jlongever/on-core/node_modules/bluebird/js/main/async.js:133:10)
            at Async.drainQueues (/home/jlongever/on-core/node_modules/bluebird/js/main/async.js:15:14)
            at process._tickDomainCallback (node.js:459:13)
      """
    friendlyName:     Redfish Client Discovery
    ignoreFailure:    false
    implementsTask:   Task.Base.Redfish.Discovery
    injectableName:   Task.Redfish.Discovery
    instanceId:       4d352217-c74d-42cf-ad60-31ea5cb4b49d
    name:             Task.Redfish.Discovery
    options:
      uri:       https://10.239.85.159/redfish/v1
```

zhangy42:
I hit close by mistake, sorry

jlongever:
@zhangy42 Validated latest commit locally against different Redfish targets.  :+1:

yyscamper:
:+1:

PR #218, opened 2016-05-24
Add OS install parameters for kickstart/preseed values

benbp:
This will make it easier to point our OS installer workflows toward custom kickstart/preseed files, and in general make changes in this location.

I'm inclined to not add documentation as I consider these more system specific options, since the kickstart files contain special callback scripts for our workflows. However, It's better that they're not hardcoded in the templates themselves (It is much, much harder to trace what files link to what files with OS installers, and will be easier to just look at a task definition and see all the files that will be downloaded).

Requires https://github.com/RackHD/on-http/pull/302

@RackHD/corecommitters @jlongever @heckj

### DIFF
--- a/lib/task-data/tasks/install-centos.js
+++ b/lib/task-data/tasks/install-centos.js
@@ -8,8 +8,9 @@ module.exports = {
     implementsTask: 'Task.Base.Os.Install',
     options: {
         osType: 'linux', //readonly options, should avoid change it
-
         profile: 'install-centos.ipxe',
+        kickstart: 'centos-ks',
+        kickstartUri: '{{api.templates}}/{{options.kickstart}}',
         hostname: 'localhost',
         comport: 'ttyS0',
         domain: 'rackhd',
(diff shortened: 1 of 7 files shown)


jlongever:
+1

anhou:
:+1:

lib/task-data/tasks/install-centos.js
```diff
@@ -8,8 +8,9 @@ module.exports = {
     implementsTask: 'Task.Base.Os.Install',
     options: {
         osType: 'linux', //readonly options, should avoid change it
-
         profile: 'install-centos.ipxe',
+        kickstart: 'centos-ks',
```
WangWinson: I have some concern on the `kickstart`, `autoya``preseed` naming. 
It does looks very clearly, but make the `option` param `schema` different from OS. Also, we may end up with updating the `doc` to reflect this , `kickstart` is **required** only for CentOS, ESXi , `preseed` is **required** only for ubuntu ...

IMHO, the template file name itself is already clear to tell what the file type is. So any chance to unify the param key naming? For example, `configTemplate`, `templateUrl` ?

PR #219, opened 2016-05-25
added a job/tasks to add switch Macs to lookups

VulpesArtificem:
solves https://github.com/RackHD/RackHD/issues/236
@RackHD/corecommitters @johren 
Added a job that gets the mac addresses (including management) and adds them to the lookups collection.

supports on https://github.com/RackHD/on-taskgraph/pull/107

### DIFF
--- a/lib/jobs/update-lookups-snmp.js
+++ b/lib/jobs/update-lookups-snmp.js
@@ -1,0 +1,55 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = updateLookupsJobFactory;
+
+var di = require('di');
+di.annotate(updateLookupsJobFactory, new di.Provide('Job.Snmp.Update.Lookups'));
+di.annotate(updateLookupsJobFactory, new di.Inject(
+    'Job.Base',
+    'Assert',
+    'Logger',
+    'Util',
+    'Services.Waterline',
+    '_'
+    )
+);
+
+function updateLookupsJobFactory(
+    BaseJob,
+    assert,
+    Logger,
+    util,
+    waterline,
+    _
+) {
+    var logger = Logger.initialize(updateLookupsJobFactory);
+
+    function UpdateLookupsJob(options, context, taskId) {
+        UpdateLookupsJob.super_.call(this, logger, options, context, taskId);
+        assert.string(this.context.target);
+        this.nodeId = this.context.target;
+    }
+    util.inherits(UpdateLookupsJob, BaseJob);
+
+    UpdateLookupsJob.prototype._run = function run() {
+        var self = this;
+        return waterline.catalogs.findLatestCatalogOfSource(self.nodeId, 'snmp-1')
+        .then(function(snmpData) {
+            return _.map(snmpData.data, function(val, key) {
+                if (key.startsWith("IF-MIB::ifPhysAddress")) {
+                    val = _.map(val.split(':'), function(chunk) {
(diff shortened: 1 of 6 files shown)


lib/jobs/update-lookups-snmp.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+    util.inherits(UpdateLookupsJob, BaseJob);
+
+    UpdateLookupsJob.prototype._run = function run() {
+        var self = this;
+        return waterline.catalogs.findLatestCatalogOfSource(self.nodeId, 'snmp-1')
+        .then(function(snmpData) {
+            return _.map(snmpData.data, function(val, key) {
+                if (key.startsWith("IF-MIB::ifPhysAddress")) {
+                    val = _.map(val.split(':'), function(chunk) {
```
benbp: A comment on why this is necessary might save future developers from confusion :)

benbp:
👍

lib/jobs/update-lookups-snmp.js
yyscamper: Yes, put some sample data here will help other developer to understand the code.

lib/jobs/update-lookups-snmp.js
```diff
@@ -0,0 +1,55 @@
...
+        assert.string(this.context.target);
+        this.nodeId = this.context.target;
+    }
+    util.inherits(UpdateLookupsJob, BaseJob);
+
+    UpdateLookupsJob.prototype._run = function run() {
+        var self = this;
+        return waterline.catalogs.findLatestCatalogOfSource(self.nodeId, 'snmp-1')
+        .then(function(snmpData) {
+            return _.map(snmpData.data, function(val, key) {
```
yyscamper: How about if catalog 'snmp-1' doesn't exist?
snmpData.data will throw a reference error because snmpData is `undefined`, but this error is not clear enough for debugging.

spec/lib/jobs/update-lookups-snmp-spec.js
```diff
@@ -0,0 +1,48 @@
...
+                "IF-MIB::ifPhysAddress_999001": "8:0:27:1e:27:4e"
+            }
+        };
+        waterline.catalogs.findLatestCatalogOfSource = sinon.stub().resolves(snmpObject);
+        waterline.lookups.upsertNodeToMacAddress = sinon.stub().resolves();
+        return job._run()
+        .then(function() {
+            expect(waterline.lookups.upsertNodeToMacAddress).to.be.calledThrice;
+            expect(waterline.lookups.upsertNodeToMacAddress).to.be
+                .calledWithExactly('someNodeId', '08:00:27:ca:c6:0a');
```
yyscamper: It's better to expect the arguments for all thrice calls.

spec/lib/jobs/update-lookups-snmp-spec.js
```diff
@@ -0,0 +1,48 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+var uuid = require('node-uuid');
+
+describe('update-lookups-job', function() {
```
yyscamper: it's better to add some negative test cases:
- The catalog source 'snmp-1' doesn't exist.
- Fail to insert lookup to database.

yyscamper:
+1 after comment is fixed.

benbp:
👍  after @yyscamper's comments as well.

PR #223, opened 2016-05-27
Add boot profile task/job

benbp:
The primary use of this is to be able to serve down a custom diskboot ipxe profile, in lieu of a way for us to intentionally ignore DHCP/PXE for a node while there is an active workflow against it.

Unfortunately, due to the way that we concatenate ALL ipxe profiles with `boilerplate.ipxe`, this task will not work effectively if it is the final task in a workflow, because the boilerplate ipxe script tries to DHCP _again_, and if we ignore those requests because there is no longer an active workflow, then it will never reach the custom script that was specified.

I think there could be more work done in that area ^^, but I'm not sure whether it's worth it, and think that this should be sufficiently simple to address the desired use case.

@RackHD/corecommitters @ytjohn @jlmcknight 

This was prompted by the following Slack discussion:

> jlmcknight [2:05 PM] 
> to expand on johnÕs question, we have a workflow that looks like this end to end:
> 
> pxe boot -> rackhd discovery -> reboot -> pxe boot into our overlayfs and do our custom catalog plus other comands (mem/cpu tests, cable validation, disk validation/partition wipe etc.) -> reboot -> pxe into ubuntu install -> reboot -> boot from disk -> ansible post install -> final reboot
> 
> That works for us on our current focus of dell c6320, but in trying to do that in vagrant, the reboot after ubuntu install is doing a pxe boot, getting error.ipxe downloaded, but when error.ipxe runs `exit` it will fail to boot from hard disk.  If we change `exit` to `sanboot Ñno-describe Ñdrive 0x80` in error.ipxe it works, but we donÕt want to make that change for vagrant when we donÕt have this problem in our bare metal installs.
> 
> Is there a way in the workflow (a task of some sort) we can shoehorn into the process after the ubuntu install that says "donÕt respond to ipxe for this hostÓ so we can control it at the workflow layer and not by editing profile files like `error.ipxe`?

### DIFF
--- a/lib/jobs/boot-profile.js
+++ b/lib/jobs/boot-profile.js
@@ -1,0 +1,55 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = bootProfileJobFactory;
+di.annotate(bootProfileJobFactory, new di.Provide('Job.BootProfile'));
+di.annotate(bootProfileJobFactory,
+    new di.Inject(
+        'Job.Base',
+        'Logger',
+        'Assert',
+        'Util'
+    )
+);
+function bootProfileJobFactory(BaseJob, Logger, assert, util) {
+    var logger = Logger.initialize(bootProfileJobFactory);
+
+    /**
+     * NOTE: Due to the way our boilerplate.ipxe script works, this job
+     * will not work if run as the final task within a workflow (boilerplate.ipxe
+     * causes ipxe to make additional DHCP requests prior to actually
+     * running the ipxe script we've served down, and those requests will get
+     * ignored if there is no currently active workflow for the node.
+     *
+     * @param {Object} options
+     * @param {Object} context
+     * @param {String} taskId
+     * @constructor
+     */
+    function BootProfileJob(options, context, taskId) {
+        BootProfileJob.super_.call(this, logger, options, context, taskId);
+        this.nodeId = this.context.target;
+        assert.string(this.nodeId, 'context.target (nodeId)');
+        this.profile = this.options.profile;
+        assert.string(this.profile, 'options.profile');
+    }
+    util.inherits(BootProfileJob, BaseJob);
+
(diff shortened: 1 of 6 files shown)


jlmcknight:
+1

heckj:
+1

ytjohn:
+1

lib/jobs/boot-profile.js
```diff
@@ -0,0 +1,55 @@
...
+
+    BootProfileJob.prototype._run = function () {
+        var self = this;
+
+        self._subscribeRequestProfile(function () {
+            return self.profile;
+        });
+
+        self._subscribeRequestProperties(function () {
+            self._done();
```
yyscamper: @benbp : What's the reason that finish the job after requesting properties? Your code just imposes a restriction that the upper layer should always first request profile then properties. Please confirm this is what you want?

benbp: @yyscamper the behavior of the HTTP server when serving profiles is to request a profile name first, then request properties that can be used for the render context for the profile (see https://github.com/RackHD/on-http/blob/master/lib/services/profiles-api-service.js#L257-L266). If we listen for requestProfile only, then http will fail on the requestProperties step. Additionally, this leaves the possibility open for a user to add a profile that requires some rendering context from the task.

I understand your concern about a restriction on the behavior of the upper layer, but it's pretty much the standard way we do this throughout the http profiles code and the job code. Maybe a better way to structure this would be refactoring requestProfile to return both a profile name and a render context all in one call, which would be less complex and impose fewer restrictions on the upper layer, but I also think this is just a pattern of behavior/technical debt that's pretty ingrained into the system that might not be worth the effort of replacing given how long it has been around.

yyscamper: Thanks @benbp , I agree this belongs to the catalog of technical debt and not obligation to this PR. Since the code logic works as expected, so this is not gating.
:+1:

PR #228, opened 2016-06-02
add task schema example: install-os-common + obm-control

yyscamper:
This is the initial PR for [Task-Schema](https://github.com/RackHD/RackHD/wiki/proposal-Task-Schema).

The primary purpose for this PR is to show some example about how the Task-Schema looks like and how the Task & Schema & Job is linked. I choose the most complex task `install-os` and a simple `obm-control` as example.

If the Task-Schema example is done, then the [Task-Annotation](https://github.com/RackHD/RackHD/wiki/proposal-task-annotation) can carry on in parallel.

Meanwhile, You should know that expect the unit-test, no other code will use the schema, so this PR is safe and will not impact our existing function. We will have subsequent PRs to gradually enrich the Task-Schema functionality.

@RackHD/corecommitters @WangWinson @iceiilin @cgx027 @pengz1 @leoyjchang @zyoung51 @jlongever

### DIFF
--- a/lib/task-data/schemas/rackhd-task-schema.json
+++ b/lib/task-data/schemas/rackhd-task-schema.json
@@ -1,0 +1,44 @@
+{
+    "id": "rackhd/schemas/v1/task-schema",
+    "type": "object",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "RackHD Tasks Schema Extention",
+    "description": "The extension for RackHD tasks schema, all RackHD tasks schema should adhere to this schema.",
+    "allOf": [
+        {
+            "$ref": "http://json-schema.org/draft-04/schema"
+        }
+    ],
+    "definitions": {
+        "describeJob": {
+            "type": "string",
+            "description": "This property is required to point to a job's di injectable name."
+        },
+        "copyright": {
+            "type": "string",
+            "description": "This attribute shall contain the copyright notice for the schema."
+        },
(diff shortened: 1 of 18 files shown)


WangWinson:
:+1: 
This is a very good start.

lib/task-data/schemas/rackhd-task-schema.json
```diff
@@ -0,0 +1,44 @@
...
+    "description": "The extension for RackHD tasks schema, all RackHD tasks schema should adhere to this schema.",
+    "allOf": [
+        {
+            "$ref": "http://json-schema.org/draft-04/schema"
+        }
+    ],
+    "definitions": {
+        "describeJob": {
+            "type": "string",
+            "description": "This property is required to point to a job's di injectable name."
```
anhou: don't need 'This property is required to point to' ?

anhou:
:+1:

lib/task-data/schemas/rackhd-task-schema.json
yyscamper: I think either is OK.
I see redfish schema has similar word, you can see the `readonly` description which is copied from redfish schema.

yyscamper:
Since there are some other stories depend on this PR and consider this is a very safe PR as the existing functionality will not be impacted. So I will merge this first after collecting enough approvals. If anyone has any comment, I will open subsequent PR to fix.

PR #230, opened 2016-06-03
Fix ODR#742: Add switchDevices/postInstallCommands in the options list of install esx task

yyscamper:
Fix this issue: https://hwjiraprd01.corp.emc.com/browse/ODR-742

Actually, This is a change after the workflow engine HA design, previous we did allow all options that passed by user to task, but then we changed to only allow options that in the task options' list.

The exact code is: https://github.com/RackHD/on-core/blob/master/lib/workflow/task-graph.js#L230.

@RackHD/corecommitters @zyoung51 @pengz1

### DIFF
--- a/lib/task-data/tasks/install-esx.js
+++ b/lib/task-data/tasks/install-esx.js
@@ -26,7 +26,24 @@ module.exports = {
         users: [],
         networkDevices: [],
         dnsServers: [],
-        installDisk: null
+        installDisk: null,
+
+        // If specified, this contains an array of objects with switchName and uplinks (optional)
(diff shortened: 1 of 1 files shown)


zyoung51:
+1

iceiilin:
:+1:

yyscamper:
Sorry, I didn't realized this PR misses an approval of CC when I  merge this PR.  Since this PR is pretty simple and safe, @RackHD/corecommitters, could you please review this PR as a complement, if you have any review comment, I will submit a hot fix for it.

anhou:
:+1:

PR #231, opened 2016-06-03
[1.2] Fix ODR#742: Add switchDevices/postInstallCommands in the options list of install esx task

yyscamper:
Fix this issue: https://hwjiraprd01.corp.emc.com/browse/ODR-742

Actually, This is a change after the workflow engine HA design, previous we did allow all options that passed by user to task, but then we changed to only allow options that in the task options' list.

The exact code is: https://github.com/RackHD/on-core/blob/master/lib/workflow/task-graph.js#L230.

@RackHD/corecommitters @zyoung51 @pengz1

### DIFF
--- a/lib/task-data/tasks/install-esx.js
+++ b/lib/task-data/tasks/install-esx.js
@@ -24,7 +24,24 @@ module.exports = {
         users: [],
         networkDevices: [],
         dnsServers: [],
-        installDisk: null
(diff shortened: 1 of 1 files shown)


yyscamper:
Please hold on the merge, I need to confirm whether this is really need in 1.2.0 branch.

PR #238, opened 2016-06-14
Perform task rendering up front

benbp:
The largest diff here is moving the task-graph class code into on-tasks from on-core. This makes the dependency structure a lot cleaner and allows for both on-http and on-taskgraph to exercise more TaskGraph/Task functionality at convenient times.

The second commit in the PR (https://github.com/RackHD/on-tasks/pull/238/commits/479876194c01206b10d37b1d4878ed2a13d916a6) adds in up front rendering logic to the `Task.create` method, primarily so that we can throw more validation errors during the synchronous API request, rather than waiting until the task actually executes within a workflow. This logic bypasses trying to render values from `context` until runtime, since those can be populated dynamically.

Requires https://github.com/RackHD/on-core/pull/158
@RackHD/corecommitters @VulpesArtificem @heckj

### DIFF
--- a/lib/task-graph.js
+++ b/lib/task-graph.js
@@ -8,6 +8,7 @@ di.annotate(taskGraphFactory, new di.Provide('TaskGraph.TaskGraph'));
 di.annotate(taskGraphFactory,
     new di.Inject(
         'Task.taskLibrary',
+        'Task.Task',
         'TaskGraph.Store',
         'Constants',
         'Assert',
@@ -19,6 +20,7 @@ di.annotate(taskGraphFactory,
 
 function taskGraphFactory(
     taskLibrary,
+    Task,
     store,
     Constants,
     assert,
@@ -242,6 +244,31 @@ function taskGraphFactory(
         definition.state = Constants.Task.States.Pending;
 
         return definition;
+    };
+
+    /**
+     * Attempt to create task objects for all tasks. Effectively this means
+     * we are exercising the task definition rendering functionality for all
+     * tasks with the information we have at this point in time, and so doing as much
+     * up front validation as we can.
+     */
+    TaskGraph.prototype.renderTasks = function() {
+        var self = this;
+        return Promise.map(_.values(self.tasks), function(taskDefinition) {
+            return Task.create(
+                taskDefinition,
+                { compileOnly: true, instanceId: taskDefinition.instanceId },
+                self.context
+            )
+            .then(function(task) {
+                // Overwrite definition options with the rendered options from
+                // the task object
+                self.tasks[taskDefinition.instanceId].options = task.options;
+            });
+        })
+        .then(function() {
+            return self;
+        });
     };
 
     TaskGraph.prototype.validate = function () {
@@ -506,6 +533,9 @@ function taskGraphFactory(
         })
         .then(function(graph) {
             return graph._populateTaskData();
+        })
+        .then(function(graph) {
+            return graph.renderTasks();
         })
         .then(function(graph) {
             graph.detectCyclesAndSetTerminalTasks();
--- a/lib/task-graph.js
+++ b/lib/task-graph.js
@@ -1,0 +1,517 @@
+// Copyright 2016, EMC, Inc.
+'use strict';
+
+var di = require('di');
+
+module.exports = taskGraphFactory;
+di.annotate(taskGraphFactory, new di.Provide('TaskGraph.TaskGraph'));
+di.annotate(taskGraphFactory,
+    new di.Inject(
+        'Task.taskLibrary',
+        'TaskGraph.Store',
+        'Constants',
+        'Assert',
+        'uuid',
+        'Promise',
+        '_'
+    )
+);
+
+function taskGraphFactory(
+    taskLibrary,
+    store,
+    Constants,
+    assert,
+    uuid,
+    Promise,
+    _
+) {
+    function TaskGraph(definition, context, domain) {
+        this.definition = definition;
+
+        if (this.definition.options && this.definition.options.instanceId) {
+            this.instanceId = this.definition.options.instanceId;
+        } else {
+            this.instanceId = uuid.v4();
+        }
+
+        // Bool
+        this.serviceGraph = this.definition.serviceGraph;
+        this.context = context || {};
+        this.context.graphId = this.instanceId;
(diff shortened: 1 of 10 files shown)


lib/task-graph.js
```diff
@@ -0,0 +1,547 @@
...
+        // If the graph has specifically defined options for a task, don't bother
+        // with whether they exist as a required option or not in the base definition.
+        if (_.has(self.definition.options, label)) {
+            allOptions = allOptions.concat(_.keys(self.definition.options[label]));
+        }
+
+        if (!_.isEmpty(self.definition.options)) {
+            _.forEach(allOptions, function(option) {
+                var taskSpecificOptions = self.definition.options[label];
+                if (_.has(taskSpecificOptions, option)) {
```
yyscamper: Though it's not responsible for this PR as it has existed in old code and you just move the code, but I think it's good chance to look at following problem:

This code means only if the user input `option` is in the task definition's options list, then it can be assigned to  the task. But sometimes the task may invent some options when build the workflow, the famous example is the task `When`: https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/redfish-discovery-graph.js#L13.

This code will just prevent the invented option from passing to the final task.

lib/task.js
```diff
@@ -149,6 +150,22 @@ function factory(
             });
     };
 
+    Task.prototype.getPath = function(obj, path) {
```
WangWinson: This function looks like directly copy from [catalog-searcher](https://github.com/RackHD/on-tasks/blob/master/lib/utils/job-utils/catalog-searcher.js#L20).  Actually, the lodash [_.get](https://lodash.com/docs#get) API should fulfill the case here, so I think we may not need to implement this ourselves.

yyscamper: Seems the lodash `get` can achieve the same functionality with `getPath`: http://devdocs.io/lodash~4/index#get

lib/task.js
```diff
@@ -404,7 +434,16 @@ function factory(
 
     Task.create = function create(definition, taskOverrides, context) {
         var task = new Task(definition, taskOverrides, context);
-        return task;
+        // If compileOnly is falsey, do rendering in the `.run()` step for
+        // compatibility with how the task runner code is written.
+        if (taskOverrides.compileOnly) {
+            return task.renderAll(task.nodeId, task.definition.options)
+            .then(function() {
```
WangWinson: Optional:  simplified version  `.return(task)`

lib/task-graph.js
```diff
@@ -0,0 +1,547 @@
...
+                { compileOnly: true, instanceId: taskDefinition.instanceId },
+                self.context
+            )
+            .then(function(task) {
+                // Overwrite definition options with the rendered options from
+                // the task object
+                self.tasks[taskDefinition.instanceId].options = task.options;
+            });
+        })
+        .then(function() {
```
WangWinson: Optional:  simplified version  `.return(self)`

lib/task.js
```diff
@@ -180,18 +201,27 @@ function factory(
...
         // Support OR syntax: {{ options.value | options.backupValue | default }}
         // ^^ Will return "default" if options.value and options.backupValue do not exist
         _.some(str.split(/[\s\n]?\|+[\s\n]?/), function(item) {
-            // If item is a string that should not be rendered yet, defer rendering
-            // until the next render stage
-            rendered = item.indexOf('.') > -1 ? catalogSearch.getPath(context, item) : item;
+            // If item is a string that should not be rendered yet (i.e. it requires
+            // data generated dynamically at workflow runtime), defer rendering
+            // until later
+            if (self._isDeferredRender(item.split('.')[0], self.state)) {
```
WangWinson: What's the purpose of passing `self.state` here? 
Looks like `_isDeferredRender` takes only 1 argument

lib/task.js
benbp: @WangWinson you caught me. I had actually moved this in because my initial iteration of this change had everything in on-core instead of on-tasks, and I guess I overlooked getting rid of it. I'll switch to `_.get`.

lib/task.js
benbp: I'll remove it. I got part of this code from a 6 month old version of the code that existed before we initially removed deferred rendering, and left some pieces in that should be removed.

lib/task.js
benbp: That is a simpler approach, didn't know about `.return`! Though at the same time I'd be worried about causing subtle bugs were other parts of this code ever changed to not pass `self` by reference, but rather another object that requires state changes within the async call chain.

lib/task-graph.js
benbp: Perhaps I'm misunderstanding what you're saying, but we do pass in options explicitly even if:

1. they are not in the original task definition at "compile" time
2. they are contained within the workflow.options.<task-label> field

https://github.com/RackHD/on-tasks/pull/238/files#diff-fe5719e39171626f6a6b53aa6aae63acR231

It's only when using the `workflow.options.defaults` shortcut that we check to see whether the existing task definition has the option or not.

yyscamper: Yes, it's the workflow.options.defaults shortcut, so we intentionally don't support pass invented option in `defaults` shortcut to all tasks? Some people have complained about this because they think the options in `defaults` shortcut can be passed to all tasks and not aware of such special limitation.

yyscamper: What's the purpose of such limitation?

yyscamper: Anyway, this is not a gating problem for me, we can move out the discussion to Slack.

yyscamper:
:+1:

yyscamper:
jenkins: test this please

lib/task-graph.js
benbp: @yyscamper correct. The reasoning for the limitation was because I didn't think it was ideal to pass a bunch of unnecessary options to other tasks that didn't need them. It only adds to the complexity of the task definition options, and might make one think "why does task X have option Y in its definition?"

benbp:
@yyscamper Jenkins will fail here because of the dependency on on-taskgraph. In fact Jenkins will fail on both the on-tasks PR and the on-core PR. I've verified locally with all changes synced but might be good to get a second person to affirm the results.

WangWinson:
:+1:

PR #241, opened 2016-06-15
Add new task to switch Discovery to create node/switch relation type

keedya:
This relation includes:

- relation type = "connectTo"
- source = "ethx"
- destination mac = Switch port mac address
- destination port name = Ethernetxx


keedya:
@RackHD/corecommitters  @zyoung51 @jlongever @uppalk1 @tannoa2

benbp:
FYI @VulpesArtificem

lib/jobs/switch-node-relations.js
```diff
@@ -0,0 +1,212 @@
...
+     * Iterate through switch mac addresses list
+     * for a match in nodes lldp catalog
+     * return a promise
+     */
+    function _findMatchingMac(nodeLldpData, lldpMacList) {
+        var relationsList = {};
+        return Promise.all(
+        _.forEach(_.keys(nodeLldpData.data), function (ethPort) {
+            //validate mac first
+            var lldpMac = catalogSearch.getPath(nodeLldpData.data[ethPort], 'chassis.mac');
```
VulpesArtificem: I think it's possible that chassis.mac may not be in the list of macs from IF-MIB::ifPhysAddress and it might instead be the value from LLDP-MIB::lldpLocChassisId (if LLDP-MIB::lldpLocChassisIdSubtype is macAddress) in which case we'll fail below even though there is data to correlate the switch to a given node.

keedya: @VulpesArtificem do u have any examples? Data I can work with?

VulpesArtificem: like here's the lldp data from a virtual node catalog: 
```
 {
    "node": "5761ca481b0388626870dddb",
    "source": "lldp",
    "data": {
      "eth0": {
        "via": "LLDP",
        "rid": "2",
        "age": "0 day, 00:00:11",
        "chassis": {
          "mac": "08:00:27:b1:13:43",
          "name": "my-arista",
          "descr": "Arista Networks EOS version 4.13.7M running on an Arista Networks vEOS",
          "mgmt-ip": "172.31.128.36",
          "Bridge": {
            "enabled": "on"
          },
          "Router": {
            "enabled": "off"
          }
        },
        "port": {
          "ifname": "Management1",
          "descr": "Not received"
        }
      }
    }
```
and here's some of the pertinent snmp data from the virtual switch:

```
"LLDP-MIB::lldpLocChassisIdSubtype_0": "macAddress",
"LLDP-MIB::lldpLocChassisId_0": "08 00 27 B1 13 43 "

"IF-MIB::ifPhysAddress_1": "8:0:27:aa:c8:8e",
"IF-MIB::ifPhysAddress_2": "8:0:27:ca:c6:a",
"IF-MIB::ifPhysAddress_999001": "8:0:27:1e:27:4e"
```

keedya: @VulpesArtificem in this case, u have the node connected to mgmt port. is that expected?

VulpesArtificem: I don't know; my concern is that it's still not showing the management port mac in the node's lldp data. 

```
"IF-MIB::ifName_1": "Ethernet1",
"IF-MIB::ifName_2": "Ethernet2",
"IF-MIB::ifName_999001": "Management1"
```
i'll also check on the node being attached to another port

yyscamper:
jenkins: test this please

keedya:
test this please

VulpesArtificem:
👍

lib/jobs/switch-node-relations.js
```diff
@@ -0,0 +1,212 @@
...
+    function _validateMacStrFormat(macString) {
+        var regex = /^(([A-Fa-f0-9]{1,2}[:]){5}[A-Fa-f0-9]{1,2}[,]?)+$/;
+        return regex.test(macString);
+    }
+
+    /**
+     * Adjust mac Address format to 2 hex digits
+     * return mac address string
+     */
+    function _adjustMacStrFormat(macString) {
```
brianparry: @keedya Do you also need to normalize case to compare macs?  Your regex accepts both upper and lower case hex digits.

lib/jobs/switch-node-relations.js
```diff
@@ -0,0 +1,212 @@
...
+     * create one if it doesn't exist
+     * return a promise
+     */
+    SwitchRelationsJob.prototype._updateSwitchRelations =
+        function updateSwitchRelations(lldpData, info) {
+            var self = this;
+            return waterline.nodes.findByIdentifier(lldpData.node)
+            .then(function (node) {
+                if (!node) {
+                    return Promise.reject('Could not find node with identifier ' + lldpData.node);
```
brianparry: @keedya This will work fine as is, but just FYI, you can also throw errors inside a `.then` callback.  These will be converted to rejections automatically.  The same is true when you want a Promise to fulfill with a specific value.  You can just return values directly without wrapping in `Promise.resolve`.  Generally you only need to return a Promise from a `.then` callback if that promise is pending an you need to wait for it to fulfill or reject prior to executing the next callback in the chain.

PR #242, opened 2016-06-16
Refactor task-schema design & finish all install-os task schemas

yyscamper:
Changes:
1. Refine schema unit-test helper structure, make it more powerful and more user friendly. 
2. Add more helper function (especially set/unset testing) to assist and simplify all schema unit-testing design. 
3. Add common-task-options schema.
4. Redesign the architecture of all install-os schemas.
5. Validate the task definition's `schemaRef`
6. Enable common-task-options unit-test for all task schema.
7. Refactor the schema unit-test helper to support non task schema.
8. Finish all install-os tasks' schema.

@RackHD/corecommitters @WangWinson @iceiilin


yyscamper:
test this please

anhou:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

anhou:
test this please

lib/task-data/schemas/common-task-options.json
```diff
@@ -0,0 +1,30 @@
...
+        "SchedulerOverrides": {
+            "type": "object",
+            "properties": {
+                "timeout": {
+                    "$ref": "#/definitions/Timeout"
+                }
+            }
+        }
+    },
+    "properties": {
```
brianparry: @yyscamper By convention, the definitions object encapsulates sub-schema that will be referenced elsewhere.  I think it would make more sense to move these properties into a schema called `options` within the `definitions` object. and change the id to `rackhd/schemas/v1/tasks/common`.  Other schema that reference this as a base would then use `$ref common/definitions/options`  This would allow us to easily add other common things in the future if needed.

lib/task-data/schemas/install-os-general.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "rackhd/schemas/v1/task-schema",
```
brianparry: What is the difference between install-os-general and install-os-base?

lib/task-data/schemas/install-centos.json
```diff
@@ -0,0 +1,19 @@
+{
+    "$schema": "rackhd/schemas/v1/task-schema",
+    "id": "rackhd/schemas/v1/tasks/install-centos",
+    "copyright": "Copyright 2016, EMC, Inc.",
+    "title": "Install CentOS",
+    "description": "The parameters for CentOS installation",
+    "describeJob": "Job.Os.Install",
+    "allOf": [
```
brianparry: Shouldn't this `allOf` operator be inside the `properties` object?

Like so:
```
"properties": {
    allOf: [
        { $ref: "Install-os-general },
        "properties": {
            "rackhdCallbackScript": {
                "$ref": "install-os-base#/definitions/RackHDCallbackScript"
            },
            "required": [ "rackhdCallbackScript" ]
        }
    ]
}
```

lib/task-data/schemas/common-task-options.json
benbp: I think this suggestion makes sense too.

lib/task-data/schemas/install-os-base.json
```diff
@@ -116,6 +165,15 @@
             "required": ["device"],
             "additionalProperties": false
         },
+        "NetworkDevices": {
+            "description": "The network cofiguration for each NIC",
```
benbp: "configuration" instead of "cofiguratoin"

spec/lib/task-data/schemas/schema-ut-helper.js
```diff
@@ -4,42 +4,190 @@
...
+/**
+ * validate task common options
+ * @memberof SchemaUnitTestHelper
+ * @param {Object} [overrideCanonicalData] - A particular canonical data to override the default one
+ */
+SchemaUnitTestHelper.prototype.validateCommonOptions = function(overrideCanonicalData) {
+    var self = this;
+    var canonical = overrideCanonicalData || self.canonicalData;
+    describe('validate task common options', function() {
+        [36000, 0, -1].forEach(function(value) {
```
benbp: I think it would be better to do a `forEach` loop of expect statements within a single `it` block, rather than dynamically generating 36000 `it` blocks, otherwise I believe the stdout from running the tests is going to be HUGE.

benbp:
👍 with Brian's comments. Great job, a lot of work to build all these schemas!!

lib/task-data/schemas/common-task-options.json
yyscamper: Also make sense to me. The only limit is that this schema cannot be verified directly in unit-test, I need to compose another one which references this schema to do verification. But this is OK.

lib/task-data/schemas/install-os-general.json
yyscamper: Actually, I fall into trouble about how to find a good schema name. I make a little complex structure to accommodate the slight difference between different install-os tasks and for sake of schema reuse:

- `install-os-base` is the basic install-os schema, with minimum parameters properties and includes all parameters' definitions, though some of the definition will not be used by install-os-base itself.

- `install-os-general` is the general parameters for most install-os tasks (Ubuntu/SUSE/CentOS/Redhat/ESXi), CoreOS is an exception because it only supports very limited parameters.  `install-os-general` extends from `install-os-base`.

- `install-centos`/`install-exsi`: These schema extends from `install-os-general`, to define the schema for these particular OS as they have some very special parameters support, such as the `rachdCallbackScript` for ESXi/CentOS and the `postInstallCommands`/`switchDevices` for ESXi.

- `install-coreos`: The special schema for CoreOS, as it is very different from other OS.

lib/task-data/schemas/install-centos.json
yyscamper: @brianparry : I see the redfish schema (https://github.com/RackHD/on-http/blob/master/static/DSP8010_1.0.0/json-schema/redfish-schema.1.0.0.json#L7) does in this way, so I copy it because I think it is more simple.
Meanwhile, I checked both syntax works well for the validator.

yyscamper:
@brianparry : I have addressed your comment about common task options, do you have still concern for your remaining un-addressed comment after my explanation?

panpan0000:
test this please.

panpan0000:
test this please

panpan0000:
test this please

brianparry:
:+1:

lib/task-data/schemas/install-centos.json
brianparry: @yyscamper Thanks for checking on this.  I agree that it makes sense to keep things consistent with Redfish schemas if they both work equivalently.

lib/task-data/schemas/install-os-general.json
brianparry: @yyscamper Yeah, I see what you mean.  At a high level, I think a structure that has a set of definitions like `install/definitions/os-base` (for os-install-base), `install/definitions/os` (for os-install-general), and `install/definitions/centos` (for centos) might make the intent of these definitions a little clearer.  In any case, I'm fine with merging as is.

yyscamper: Thanks @brianparry, some local people still have same concern that the install-os schema structure is a little complex. I will improve this later.

PR #243, opened 2016-06-17
Change existing jobs to use new OBMs model.

brianparry:
Depends on: 
https://github.com/RackHD/on-core/pull/164

### DIFF
--- a/lib/jobs/ipmi-catalog.js
+++ b/lib/jobs/ipmi-catalog.js
@@ -34,18 +34,6 @@ function ipmiCatalogJobFactory(
 ) {
     var logger = Logger.initialize(ipmiCatalogJobFactory);
 
-    function revealSecrets (options) {
-        if (options.password) {
-            options.password = encryption.decrypt(options.password);
-        }
-
-        if (options.community) {
-            options.community = encryption.decrypt(options.community);
-        }
-
-        return options;
-    }
-
     /**
      * @param {Object} options
      * @param {Object} context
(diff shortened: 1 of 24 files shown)


anhou:
test this please

panpan0000:
test this please

lib/jobs/ipmi-catalog.js
```diff
@@ -70,35 +58,28 @@ function ipmiCatalogJobFactory(
         var self = this;
         var cmdResp = [];
 
-        waterline.nodes.findByIdentifier(self.nodeId)
-        .then(function (node) {
-            assert.ok(node, 'No node for ipmi catalog');
+        return waterline.obms.findByNode(self.nodeId, 'ipmi-obm-service', true)
+            .then(function (obmSettings) {
+                assert.object(obmSettings, 'No OBM service available.');
+                assert.ok(!_.isEmpty(obmSettings.config), 'No OBM service config available.');
```
yyscamper: whether this means the obm will still be pass when inserting an obm without `config`? I think the obm model should prevent those obm service being added.

yyscamper:
:+1:

brianparry:
@panpan0000 @anhou Tests are failing because of the on-core dependency.  We are still working on changes to fix a 1.1 API compatibility issue in on-http.

lib/jobs/ipmi-catalog.js
brianparry: @yyscamper This is true.  If both the schema validator and the waterline model work properly, we should never have a model without the `config` property.  I added this mainly to make existing unit tests pass.

yyscamper: That's OK.

srinia6:
+1

benbp:
👍

PR #244, opened 2016-06-20
Create TaskOptionValidator inherit JsonSchemaValidator

WangWinson:
This is a sub class specific for task option validation.
As @yyscamper 's design:
- need `register` function to load all task schemas
- need a specific `validate` function to skip all `context` task option during validation
- customize `keyword` and `format` will be added later (on demand)

Dependency PR : https://github.com/RackHD/on-core/pull/165

@RackHD/corecommitters

### DIFF
--- a/lib/utils/task-option-validator.js
+++ b/lib/utils/task-option-validator.js
@@ -20,9 +20,9 @@ function taskOptionValidatorFactory(
     _,
     util
 ) {
-    var TaskSchemaFolder = path.resolve(__dirname, '../../lib/task-data/schemas');
-    var MetaSchemaFileName = 'rackhd-task-schema.json';
-    var ContextPattern = /\{\{[\s\S]*context\.[\s\S]*\}\}/;
+    var defaultTaskSchemaFolder = path.resolve(__dirname, '../../lib/task-data/schemas');
+    var defaultMetaSchemaFileName = 'rackhd-task-schema.json';
+    var contextPattern = /\{\{[\s\S]*context\.[\s\S]*\}\}/;
 
     function TaskOptionValidator () {
         JsonSchemaValidator.call(this, { allErrors: true, verbose: true });
@@ -32,12 +32,18 @@ function taskOptionValidatorFactory(
     util.inherits(TaskOptionValidator, JsonSchemaValidator);
 
     /**
-     * register the validator with all pre defined JSON schema
+     * register the validator with all pre defined JSON schemas
+     * @param  {String} schemaDir  Directory for schemas, 'lib/task-data/schemas'
+     *                             by default if not specified
+     * @param  {String} metaSchemaName  Meta Schema file name in schemaDir
      * @return {Promise}
      */
-    TaskOptionValidator.prototype.register = function () {
+    TaskOptionValidator.prototype.register = function (schemaDir, metaSchemaName) {
         var self = this;
-        return self.loader.getAll(TaskSchemaFolder, false)
+        schemaDir = schemaDir || defaultTaskSchemaFolder;
+        metaSchemaName = metaSchemaName || defaultMetaSchemaFileName;
+
+        return self.loader.getAll(schemaDir, false)
         .then(function (files) {
             return _.transform(files, function (result, v, k) {
                 result[k] = JSON.parse(v.contents);
@@ -45,9 +51,9 @@ function taskOptionValidatorFactory(
         })
         .tap(function (files) {
             // add meta schema
-            var metaSchemaFile = files[MetaSchemaFileName];
+            var metaSchemaFile = files[metaSchemaName];
             self.addMetaSchema(metaSchemaFile);
-            delete files[MetaSchemaFileName];
+            delete files[metaSchemaName];
         })
         .then(function (files) {
             // add all other schemas
@@ -70,13 +76,13 @@ function taskOptionValidatorFactory(
         }
 
         var errors = _.filter(this._ajv.errors, function(error) {
-            return !ContextPattern.test(error.data + '');
+            return !contextPattern.test(error.data + '');
         });
 
         if (_.isEmpty(errors)) {
             return true;
         } else {
-            var err = new Error('JSON schema validation failed - ' + this._ajv.errorsText());
+            var err = new Error('JSON schema validation failed - ' + this._ajv.errorsText(errors));
             err.errorList = errors;
             throw err;
         }
--- a/lib/utils/task-option-validator.js
+++ b/lib/utils/task-option-validator.js
@@ -1,0 +1,93 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+var di = require('di');
+var path = require('path');
+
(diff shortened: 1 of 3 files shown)


lib/utils/task-option-validator.js
```diff
@@ -0,0 +1,93 @@
...
+
+function taskOptionValidatorFactory(
+    JsonSchemaValidator,
+    FileLoader,
+    _,
+    util
+) {
+    var TaskSchemaFolder = path.resolve(__dirname, '../../lib/task-data/schemas');
+    var MetaSchemaFileName = 'rackhd-task-schema.json';
+    var ContextPattern = /\{\{[\s\S]*context\.[\s\S]*\}\}/;
```
yyscamper: As JavaScript naming convention, The first character for above 3 variable names need be in low case.

lib/utils/task-option-validator.js
```diff
@@ -0,0 +1,93 @@
...
+                result[k] = JSON.parse(v.contents);
+            }, {});
+        })
+        .tap(function (files) {
+            // add meta schema
+            var metaSchemaFile = files[MetaSchemaFileName];
+            self.addMetaSchema(metaSchemaFile);
+            delete files[MetaSchemaFileName];
+        })
+        .then(function (files) {
```
yyscamper: Previous promise doesn't return any value, so  the `files` here will be undefined.

lib/utils/task-option-validator.js
```diff
@@ -0,0 +1,93 @@
...
+    };
+    
+    /**
+     * validate JSON data with given JSON schema
+     * @param  {Object|String} schema  JSON schema Object or schema ref id
+     * @param  {Object} data  JSON data to be validated
+     * @return {Boolean}
+     */
+    TaskOptionValidator.prototype.validateContextSkipped = function (schema, data) {
+        if (this._ajv.validate(schema, data)) {
```
yyscamper: It's better to hide the internal `_ajv` instance in the inherited validator

lib/utils/task-option-validator.js
```diff
@@ -0,0 +1,93 @@
...
+            return true;
+        }
+
+        var errors = _.filter(this._ajv.errors, function(error) {
+            return !ContextPattern.test(error.data + '');
+        });
```
yyscamper: @benbp : this is your requested feature: support skipping context without modifying the schema file itself.

spec/lib/utils/task-option-validator-spec.js
```diff
@@ -0,0 +1,60 @@
...
+        });
+
+        it('should throw validation error with incorrect data format', function () {
+            expect(function () {
+                taskOptionValidator.validateContextSkipped(testSchema1,{ repo : 'abc' });
+            }).to.throw(Error, 'data.repo should match format "uri"');
+        });
+
+        it('should throw validation error with task option not rendered', function () {
+            var task = require('../../../lib/task-data/tasks/install-centos.js');
```
yyscamper: Suggest removing such dependency, you can directly the option data to here. Imagine in further, the install-centos file name is changed, it's not good design if we need change the file name here as well.

spec/lib/utils/task-option-validator-spec.js
```diff
@@ -0,0 +1,60 @@
...
+    before(function() {
+        helper.setupInjector([
+            helper.require('/lib/utils/task-option-validator')
+        ]);
+        taskOptionValidator = helper.injector.get('TaskOption.Validator');
+    });
+
+    describe('register', function() {
+        it('should load all JSON schemas under /lib/task-data/schemas', function() {
+            var schemaId = 'rackhd/schemas/v1/tasks/install-os-common';
```
yyscamper: Suggest removing the dependency here. You can stub the FileLoader and let it return some example schema files.

lib/utils/task-option-validator.js
WangWinson: The convention you mentioned should apply to local variable, right ? I was intend to make these a static variable at class level.

lib/utils/task-option-validator.js
WangWinson: The 'previous promise' is `tap` function which will automatically return exactly the same result from the previous promise chain.
Is this case:
the `tap` function is equal to 
```javascript
.then(function (files) {
    // do somthing
    return files
})
```

lib/utils/task-option-validator.js
WangWinson: Yes, I originally think the `_ajv` instance could be hidden in the super class. But it's useful in any sub class which need to implement some specific function like this. 
So , I would like to make the `_ajv` more like a `protected` property which could be using by sub class.

spec/lib/utils/task-option-validator-spec.js
WangWinson: Sure. I'll add some test option data directly

lib/utils/task-option-validator.js
WangWinson: Yes, it filters the error instead of modifying the schema itself.

lib/utils/task-option-validator.js
yyscamper: I checked this naming convention documentation, seems we don't need to distinguish local variable or class static variable: http://javascript.crockford.com/code.html#names

lib/utils/task-option-validator.js
yyscamper: The official document has the return, it confuses me. So that's OK now.

lib/utils/task-option-validator.js
yyscamper: My concern is that if we don't hide the `_ajv` instance, then the on-core's JSON-Schema validator seems not so necessary, because it behaves just like renaming work.

lib/utils/task-option-validator.js
benbp: @WangWinson if you want it to be like a class static variable, you can also do:

`TaskOptionValidator.ContextPattern = <...>`

and reference it the same way, e.g.

```
> function A() {}
undefined
> A.foo = 'bar'
'bar'
> A.foo
'bar'
> var a = new A();
undefined
> a.constructor.foo  // another way to do it dynamically
'bar'
>
```

WangWinson: Thanks @benbp . 
Looks like there is no necessary to expose these var outside. So I would update according to @yyscamper  's suggestion.

lib/utils/task-option-validator.js
WangWinson: This will be improved a little later. As discussed, there will be a `validatePatternSkipped` function added to base class, then call that function here.

WangWinson:
Note: CI will fail , because it need to wait for on-core PR merged first.

yyscamper:
Jenkins: test this please

lib/utils/task-option-validator.js
yyscamper: Sound good

anhou:
:+1:

yyscamper:
:+1:

PR #247, opened 2016-06-24
increase default waiting time for ssh validation to cover dns server unreachable case 

iceiilin:
Fix ODR767 Fail to boot strap "Redhat Linux+KVM" on C6320 server

It happened when DNS address configured in node's OS is unreachable. It would take longer time when RackHD tries to login via SSH, when node is waiting for the DNS server to reply the IP of client's domain name. The solution is to increase the ssh client waiting time from 10s to 20s.

@RackHD/corecommitters  @cgx027


yyscamper:
:+1:

yyscamper:
@VulpesArtificem

anhou:
:+1:

PR #249, opened 2016-06-27
Create alert for obm status change

pengz1:
Create node alert when pollers status changed.
"accessible" alert will be issued if pollers status changed from inaccessible to accessible;
"inaccessible" alert will be issue if pollers status changed from accessible to inaccessible;


pengz1:
@cgx027 @RackHD/corecommitters

spec/lib/jobs/ipmi-job-spec.js
```diff
@@ -5,7 +5,8 @@
```
anhou: please keep 0644 file mode

lib/jobs/ipmi-job.js
```diff
@@ -127,34 +129,62 @@ function ipmiJobFactory(
...
+                        delete data.password;
+                    }
+                    return self._publishIpmiCommandResult(self.routingKey, cmd, data);
+                }).then(function() {
+                    return waterline.workitems.findOne({ id: data.workItemId });
+                })
                 .then(function(workitem) {
-                    return waterline.workitems.setFailed(null, workitem);
+                    var workitemOld = workitem;
+                    return Promise.resolve()
```
anhou: it could be below to avoid, promise.resolve() and one more level .then.
return waterline.workitems.setSucceeded(null, workitem)
    .then (function () { 
         xxxxx
    });

lib/jobs/ipmi-job.js
```diff
@@ -127,34 +129,62 @@ function ipmiJobFactory(
...
+                        data: data,
+                        error: err
+                    });
+                    return waterline.workitems.findOne({ id: data.workItemId })
+                        .then(function(workitem) {
+                            var workitemOld = workitem;
+                            return Promise.resolve()
+                                .then(function(){
+                                    //failure count ascends from 1
+                                    if (workitemOld.failureCount === 1) {
```
anhou: so from succeed  to fail , orkitemOld.failureCount has been increased before this place?

lib/jobs/ipmi-job.js
```diff
@@ -127,34 +129,62 @@ function ipmiJobFactory(
...
+                }).then(function() {
+                    return waterline.workitems.findOne({ id: data.workItemId });
+                })
                 .then(function(workitem) {
-                    return waterline.workitems.setFailed(null, workitem);
+                    var workitemOld = workitem;
+                    return Promise.resolve()
+                        .then(function(){
+                            if (workitemOld.failureCount > 0) {
+                                return protocolEvents.publishNodeAlert(workitemOld.node,
```
benbp: Might be better to have this in a `.tap`, since we don't want `setSucceeded` to be skipped even if the alert publish fails.

benbp: Which would warrant its own `.catch` handler w/ failure logging though.

pengz1: Good point, I will update later.
Thanks Ben.

PR #252, opened 2016-06-27
Issue103 add/merge http client/job

zhangy42:
@benbp 

Added the base job that is originally forgot to add.


benbp:
@zhangy42 Is the intent here just to add the base task, since the other changes have already gone in? If so, can you rebase against master so we just have the commit for the base task.

zhangy42:
Let me create another branch, and copy in the files again. When I was doing the rebase it raised too many conflicts. It does not worth the time resolving the conflicts.

PR #255, opened 2016-06-28
Add alert for SNMP accesse status changes

pengz1:
Create alert for SNMP access status changes. This PR is similar to PR:
https://github.com/RackHD/on-tasks/pull/249


pengz1:
@RackHD/corecommitters @cgx027

PR #256, opened 2016-06-28
Create Linux diagnostic task

jlongever:
@RackHD/corecommitters @uppalk1 @tannoa2 @zyoung51 @keedya 

Dependency on PR https://github.com/RackHD/on-http/pull/345

### DIFF
--- a/lib/task-data/tasks/run-emc-diag.js
+++ b/lib/task-data/tasks/run-emc-diag.js
@@ -1,0 +1,26 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Run EMC Diag',
+    injectableName: 'Task.Os.Run.Emc.Diag',
+    implementsTask: 'Task.Base.Os.Install',
+    options: {
+        kernelFile: 'vmlinuz-4.1.15-diag',
+        initrdFile: 'ramfs.lzma',
+        kernelUri: '{{ api.server }}/common/{{ options.kernelFile }}',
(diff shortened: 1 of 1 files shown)


lib/task-data/tasks/run-emc-diag.js
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = {
+    friendlyName: 'Run EMC Diag',
+    injectableName: 'Task.Os.Run.Emc.Diag',
+    implementsTask: 'Task.Base.Os.Install',
+    options: {
+        kernelFile: 'vmlinuz-4.1.15-diag',
+        initrdFile: 'ramfs.lzma',
+        kernelUri: '{{ api.server }}/common/{{ options.kernelFile }}',
+        initrdUri: '{{ api.server }}/common/{{ options.initrdFile }}',
```
anhou: looks a little duplication here. Maybe remove 'kernelUri' and 'initrdUri' then add a field like 'directory: common' or 'owner: common' or some other more reasonable name, and more, add 'uri: {{api.server}}' ' is more  flexible?

jlongever: @anhou  This follows the same pattern as `bootstrap-ubuntu.js`.  https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/bootstrap-ubuntu.js

Should we change all of these task definitions?  Is this a gating issue for you?

anhou: if you think this is fine, it's just nice to have. but not a gating from me.

jlongever: Thanks.

PR #250, opened 2016-06-27
change the default value of 'domain' to null

lanchongyizu:
**Fix ODR-751** Setting 'domain' option is effectively required in esx-ks

The solution is changing the default value of 'domain' to null.

@RackHD/corecommitters  @panpan0000

PR #257, opened 2016-06-29
enable task-schema upfront validation

yyscamper:
1. check task options (skip context value) when creating a task
2. do full schema validation before run the job
3. remove some install-os job validation code as they have been covered by task-schem.
4. remove some null or empty option in install-os options list to pass task schema validation.
5. install-os: change the domain to be optional (requirement from: ODR-751)
6. fix the typo of obm-control task schema.

depends on: https://github.com/RackHD/on-core/pull/171

@RackHD/corecommitters @WangWinson @iceiilin @cgx027 @pengz1

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -52,7 +52,6 @@ function installOsJobFactory(
         self.options = _.assign(self.options, context.repoOptions);
 
         this._validateOptions();
-        this._convertOptions();
         this._encryptPassword();
         this._provideUserCredentials();
     }
@@ -62,28 +61,14 @@ function installOsJobFactory(
     /**
      * @memberOf InstallOsJob
      *
-     * Validate the input options.
+     * Validate the input options, only implement those validation that task-schema cannot covers
      */
     InstallOsJob.prototype._validateOptions = function() {
         assert.string(this.context.target);
-        assert.string(this.options.completionUri);
-        assert.string(this.options.profile);
-
-        // Some of the install task (such as coreos) still hard coded the repo in
-        // the profile/kickstart file, So we cannot assert repo&version here
-        //
-        // TODO: If all install tasks use the same options format,
-        // then uncomment following  lines:
-        // assert.string(this.options.repo);
-        // assert.string(this.options.version);
-        // assert.string(this.options.rootPassword);
-        // assert.string(this.options.hostname);
-        // assert.string(this.options.domain);
 
         if (this.options.networkDevices) {
             _.forEach(this.options.networkDevices, function(dev) {
                 assert.string(dev.device);
-                var vlanIds;
                 if (dev.ipv4) {
                     assert.isIP(dev.ipv4.ipAddr, 4);
                     assert.isIP(dev.ipv4.gateway, 4);
@@ -97,26 +82,6 @@ function installOsJobFactory(
                         }
                         /* jshint ignore:end */
                     });
-
-                    //support vlanId with 'deprecated' message
-                    if(dev.ipv4.vlanId){
-                        dev.ipv4.vlanIds = dev.ipv4.vlanIds || dev.ipv4.vlanId;
-                        logger.deprecate("'vlanIds' is recommended instead of 'vlanId'");
-                    }
-
-                    if(dev.ipv4.vlanIds) {
-                        vlanIds = dev.ipv4.vlanIds;
-                        assert.array(vlanIds);
-                        _.forEach(vlanIds, function(vlanId, index) {
-                            vlanId = +vlanId;
-                            if(parseInt(vlanId) === vlanId && 0 <= vlanId && vlanId <= 4095) {
-                                //string vlanID will be convert to number
-                                vlanIds[index] = parseInt(+vlanId);
-                            } else {
-                                throw new RangeError('VlanId must be between 0 and 4095.');
-                            }
-                        });
-                    }
                 }
 
                 if (dev.ipv6) {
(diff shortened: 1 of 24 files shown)


lib/task-graph.js
```diff
@@ -220,25 +220,43 @@ function taskGraphFactory(
...
-        );
-
-        // If the graph has specifically defined options for a task, don't bother
-        // with whether they exist as a required option or not in the base definition.
-        if (_.has(self.definition.options, label)) {
-            allOptions = allOptions.concat(_.keys(self.definition.options[label]));
+        // If there is JSON schema defined for the task, then pass all task-specified option and
+        // defaults options to the task, let the schema to determine whether the additional options
+        // is allowed or not.
+        if (definition.hasOwnProperty('schemaRef')) {
```
yyscamper: @benbp : This is the code that we had discussed before. The code is to pass all options that in `defaults` into all tasks. Previously we only pass those options only if that option in the option list of task definition. But now when the Task Schema is enabled, the `options` in the task definition doesn't reflect the whole supported options, because I have to remove those `null` or `[]` placeholder option as they violate the schema.

The will be a temporary workaround, I think the best solution is to find the whole supported option name from the schema. This is a challenge right now, as the schema may contain sub-schema. I plan to resolve this challenge later.

lib/task-data/tasks/install-centos.js
```diff
@@ -14,20 +14,10 @@ module.exports = {
         installScriptUri: '{{api.templates}}/{{options.installScript}}',
         hostname: 'localhost',
         comport: 'ttyS0',
-        domain: 'rackhd',
         completionUri: 'renasar-ansible.pub',
```
yyscamper: If the task schema is enabled, the original `options` property in task definition become the place to set the default value for some of option. So we should not expect the original `options` property will give you the full supported option list for a specific task, we should always turn to lookup the schema for all supported options. This is the next enhancement plan for task schema.

yyscamper:
The unit-test error is expected as it depends on: https://github.com/RackHD/on-core/pull/171

lib/task-data/tasks/install-ubuntu.js
```diff
@@ -14,17 +14,10 @@ module.exports = {
         installScriptUri: '{{api.templates}}/{{options.installScript}}',
         hostname: 'localhost',
         comport: 'ttyS0',
-        domain: 'rackhd',
```
WangWinson: Any reason to remove `domain` option here ?

lib/jobs/install-os.js
```diff
@@ -62,28 +61,14 @@ function installOsJobFactory(
...
-        // assert.string(this.options.rootPassword);
-        // assert.string(this.options.hostname);
-        // assert.string(this.options.domain);
 
         if (this.options.networkDevices) {
             _.forEach(this.options.networkDevices, function(dev) {
                 assert.string(dev.device);
-                var vlanIds;
                 if (dev.ipv4) {
                     assert.isIP(dev.ipv4.ipAddr, 4);
```
WangWinson: So, the `ipv4`, `ipv6` `format` in ajv didn't cover this ?

lib/task-data/tasks/install-ubuntu.js
yyscamper: I forgot to mention this change, you can check this PR (https://github.com/RackHD/on-tasks/pull/250), it is a requirement from ODR-751

yyscamper: And that PR forgot to change the corresponding schema, this is a fix for it.

lib/jobs/install-os.js
yyscamper: Yes, ajv only has 'ipv4' format support. I keep the `ipv4` here is to align with the code about `ipv6`.

WangWinson:
:+1:

benbp:
👍

yyscamper:
jenkins: test this please

yyscamper:
jenkins: test this please

brianparry:
test this please

brianparry:
Not sure why the failure didn't get posted back by JenkinsRHD, but here it is:

Regression

Tests_build.spec.Graph Library should validate all existing graphs not requiring user input for null values (from Mocha Tests)

Error Message

cannot find the task schema with id rackhd/schemas/v1/tasks/obm-control
Stacktrace

Error: cannot find the task schema with id rackhd/schemas/v1/tasks/obm-control
    at new Task (node_modules/on-tasks/lib/task.js:75:23)
    at Function.create (node_modules/on-tasks/lib/task.js:463:20)
    at node_modules/on-tasks/lib/task-graph.js:282:25
    at tryCatcher (node_modules/on-core/node_modules/bluebird/js/main/util.js:26:23)
    at MappingPromiseArray._promiseFulfilled (node_modules/on-core/node_modules/bluebird/js/main/map.js:56:38)
    at MappingPromiseArray.init [as _init$] (node_modules/on-core/node_modules/bluebird/js/main/promise_array.js:92:18)
    at MappingPromiseArray.init (node_modules/on-core/node_modules/bluebird/js/main/map.js:29:23)
    at Async._drainQueue (node_modules/on-core/node_modules/bluebird/js/main/async.js:128:12)
    at Async._drainQueues (node_modules/on-core/node_modules/bluebird/js/main/async.js:133:10)
    at Async.drainQueues (node_modules/on-core/node_modules/bluebird/js/main/async.js:15:14)

yyscamper:
@brianparry : The unit-test was caused by the downstream job of on-tasks, but the on-tasks itself is OK. Actually, the related PR of on-taskgraph (https://github.com/RackHD/on-taskgraph/pull/134) is just to fix this unit-test error. So there is inter-dependency between this one and the one of on-taskgraph.

I have tested in local environment with both PRs ready, the unit-test succeed.

yyscamper:
Jenkins: test this please

PR #260, opened 2016-06-30
fix issue 282 about dash in serail number when generating enclosure

iceiilin:
When generating enclosure ID, previously the job only views number and string as valid serial number. In https://github.com/RackHD/RackHD/issues/282, the reporter uses a platform with "-" in this field.

Change the regex to adjust this character.
Move const value into the factory closure so that it won't be global value to potentially impact other codes.


yyscamper:
resolves https://github.com/RackHD/RackHD/issues/282

yyscamper:
:+1:

iceiilin:
Add cases to increase coverage

iceiilin:
test this please

anhou:
resolves https://github.com/RackHD/RackHD/issues/282

anhou:
:+1:

PR #261, opened 2016-06-30
added polling and subgraph recovery

VulpesArtificem:
fix for https://github.com/RackHD/RackHD/issues/277
Also adds a cleanup step to cancel a subgraph when the run-graph tasks are cancelled 
as per https://www.pivotaltracker.com/story/show/117106547
requires https://github.com/RackHD/on-core/pull/175
@RackHD/corecommitters @pscharla

### DIFF
--- a/lib/jobs/run-graph.js
+++ b/lib/jobs/run-graph.js
@@ -12,13 +12,14 @@ di.annotate(runGraphJobFactory, new di.Provide('Job.Graph.Run'));
         'Protocol.TaskGraphRunner',
         'TaskGraph.TaskGraph',
         'TaskGraph.Store',
-        'Services.Waterline',
         'Constants',
         'Logger',
         'Assert',
         'uuid',
         'Util',
-        'JobUtils.WorkflowTool'
+        'JobUtils.WorkflowTool',
+        '_',
+        'Errors'
     )
 );
 function runGraphJobFactory(
@@ -26,13 +27,14 @@ function runGraphJobFactory(
     taskGraphProtocol,
     TaskGraph,
     taskGraphStore,
-    waterline,
     Constants,
     Logger,
     assert,
     uuid,
     util,
-    workflowTool
+    workflowTool,
+    _,
+    Errors
 ) {
     var logger = Logger.initialize(runGraphJobFactory);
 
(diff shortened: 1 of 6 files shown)


benbp:
test this please

lib/jobs/run-graph.js
```diff
@@ -43,57 +47,97 @@ function runGraphJobFactory(
...
-
-        this._subscribeGraphFinished(function(status) {
-            if (status === Constants.Task.States.Succeeded) {
-                self._done();
-            } else {
-                self._done(new Error("Graph " + self.graphId + " failed with status " + status));
+        self.subscribeOwnGraph().then(function(ownGraph) {
+            self.subgraphPoll = self.setSubgraphPoll();
+            if (!ownGraph) {
+            return workflowTool.runGraph(
```
benbp: indentation

lib/jobs/run-graph.js
```diff
@@ -43,57 +47,97 @@ function runGraphJobFactory(
...
+        } else if (_.contains(Constants.Task.FailedStates, subGraphStatus)) {
+            this._done(new Error(
+                "Graph " + this.graphId + " failed with status " + subGraphStatus)
+            );
+        }
+    };
+
+    RunGraphJob.prototype.subscribeOwnGraph = function() {
+        var self = this;
+        return waterline.graphobjects.findOneMongo({"context._parent.taskId": self.taskId})
```
benbp: Can you call this behind the store layer, that we don't have mongo-specific code in the job code.

lib/jobs/run-graph.js
```diff
@@ -43,57 +47,97 @@ function runGraphJobFactory(
...
+            if(ownGraph) { self.graphId = ownGraph.context.graphId; }
+            self._subscribeGraphFinished(self.finishWithState);
+            return ownGraph;
         });
     };
 
+   RunGraphJob.prototype.setSubgraphPoll = function() {
+        var self = this;
+        return setInterval(function() {
+            waterline.graphobjects.findOneMongo({"context._parent.taskId": self.taskId})
```
benbp: Same comment as above re: store

yyscamper:
@VulpesArtificem : I think you are writing some complex logic about task-graph running into the job code, but I think it's better belong to the task-graph.js. For example, create an API  "runSubGraph" in task-graph.js, it returns promise, the promise is fulfilled when subgraph successes, rejected when subgraph fails.
@benbp : how do you think?

yyscamper:
@VulpesArtificem @benbp: correct my last comment, it should belong to task-graph-scheduler not task-graph.js If it is task-graph scheduler, then this will introduce dependence on on-taskgraph, this is not what we want. Alternative is to move some task-graph-scheduler code to on-core, but this requires more code arch change, I don't want to push so much work for you and for this PR.

Anyway, I am OK with putting the complex code here, but I think this is a defect of code arch, we can record it and improve it in future.

VulpesArtificem:
@yyscamper @benbp 
I think it's appropriate for this logic to go in the jobs; a "runGraphJob" is always going to be at odds with the general architecture because it necessarily employs graph management logic. I think it makes sense for runSkuGraphJob to inherit because it is conceptually a subtype of runGraphJob and graph management methods are unique to the runGraphJob type and need access to the job object.

brianparry:
test this please

yyscamper:
@VulpesArtificem : I checked the Jenkins log, the failure probably due to the on-tasks is relative too old, a rebase should fix it.

benbp:
merged the dependent PR. 
test this please

pscharla:
@RackHD/corecommitters Need reviews for this PR

pscharla:
@yyscamper @anhou Asked @brianparry to review this one in slack channel, but please take a look as well. Need this one in to hopefully resolve P1 ODR.

DavidjohnBlodgett:
👍

lib/jobs/run-graph.js
```diff
@@ -43,57 +45,97 @@ function runGraphJobFactory(
...
+        var self = this;
+        return taskGraphStore.findChildGraph(self.taskId)
+        .then(function(ownGraph) {
+            if(ownGraph) { self.graphId = ownGraph.context.graphId; }
+            self._subscribeGraphFinished(self.finishWithState);
+            return ownGraph;
         });
     };
 
+   RunGraphJob.prototype.setSubgraphPoll = function() {
```
brianparry: @VulpesArtificem Could you explain the reasoning behind polling for finished graphs in addition to subscribing to graph finished events?  Is this here to protect against missed events?

lib/jobs/run-graph.js
```diff
@@ -43,57 +45,97 @@ function runGraphJobFactory(
...
-            self.taskId
-        ).catch(function(error) {
-            logger.error("Error starting graph for task.", {
-                taskId: self.taskId,
-                graphName: self.options.graphName,
-                error: error
-            });
-            self._done(error);
+    RunGraphJob.prototype.finishWithState = function(subGraphStatus) {
+        if (subGraphStatus === Constants.Task.States.Succeeded) {
```
brianparry: @VulpesArtificem If `self._done` is called, shouldn't the interval timer also be cancelled?  It seems like there might be a small window for a race condition here where the timer expires between handling a graph finished event and calling the `_cleanup` method.  I think this would result in calling `_done` twice.

brianparry:
@VulpesArtificem Your logic looks sound to me.  I just have a couple of questions to make sure I'm understanding properly.

lib/jobs/run-graph.js
VulpesArtificem: yes, it's in case we subscribe after our graph has already finished so we don't end up waiting forever

lib/jobs/run-graph.js
VulpesArtificem: derp, yes, i'll move that up to right here so we cancel it in time

lib/jobs/run-graph.js
```diff
@@ -43,57 +45,97 @@ function runGraphJobFactory(
...
+            if (this.subgraphPoll) {clearInterval(this.subgraphPoll);}
```
yyscamper: This kind of code isn't recommended, please change it to following:
```javascript
if (this.subgraphPoll) {
    clearInterval(this.subgraphPoll);
}
```

yyscamper:
:+1: after changing the code format that I outlined.

brianparry:
@VulpesArtificem Looks good to me.  :+1: once the tests are fixed.

pscharla:
test this please

pscharla:
@RackHD/corecommitters I think this PR is now ready to go. Please merge if you are ready as it is gating a P1 ODR. Thanks!

PR #264, opened 2016-07-05
Create task-annotation-generator tool script

WangWinson:
The tool script leverage `TaskOption.Validator` to `register` all task schemas,
then get all resolved schema to generate a `task-doc-data.json`, which will be using for task doc build script (later updated by @cgx027 )

@RackHD/corecommitters @cgx027 @iceiilin

### DIFF
--- a/task-annotation-generator.js
+++ b/task-annotation-generator.js
@@ -1,0 +1,140 @@
+// Copyright 2016, EMC
+
+'use strict';
+
+var fs = require('fs');
+
+function TaskAnnotation(validator) {
+    this.validator = validator;
+}
+
+TaskAnnotation.prototype.mergeSchema = function (obj) {
+    var self = this;
+    if (obj.allOf && obj.allOf instanceof Array) {
+        var all = {};
+        _.forEach(obj.allOf, function (item) {
+            if(item.allOf) {
+                item = self.mergeSchema(item);
+            }
+            _.merge(all, item);
+        });
+        delete obj.allOf;
+        _.merge(obj, all);
+    }
+
+    _.forOwn(obj, function(val) {
+        if (val instanceof Object) {
+            self.mergeSchema(val);
+        }
+    });
+
+    return obj;
+};
+
+TaskAnnotation.prototype.generateDocData = function (obj, dataTemplate) {
+    var self = this;
+    var data = {
+        type: obj.type,
+        description: obj.description,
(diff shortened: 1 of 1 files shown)


brianparry:
test this please

iceiilin:
+1

task-annotation-generator.js
```diff
@@ -0,0 +1,140 @@
...
+                name: data.name + '_' + name,
+                title: data.title + '.' + name,
+                group: data.group,
+                groupTitle: data.groupTitle
+            };
+            subItems = subItems.concat(self.generateDocData(option, subTemp));
+        }
+
+        var fieldTemp = {
+            group: 'g1', // TODO: find out gourp usage
```
yyscamper: typo: gourp -> group

task-annotation-generator.js
```diff
@@ -0,0 +1,140 @@
...
+
+'use strict';
+
+var fs = require('fs');
+
+function TaskAnnotation(validator) {
+    this.validator = validator;
+}
+
+TaskAnnotation.prototype.mergeSchema = function (obj) {
```
yyscamper: Could you please add some comment for each function to help other people to understand the design?

task-annotation-generator.js
WangWinson: oh, you catch me again.

task-annotation-generator.js
WangWinson: Sure. I can do it

yyscamper:
How about adding some unit-test? though this file is not used by RackHD service code, but it contains some complex logic and some unit-test will prevent the task-annotation from being broken down.

WangWinson:
@yyscamper 
Actually, this is a tool script, which should be a part of build script, instead of a part of deliver code.
I know I can add some unit test for it , since it's written in JS.

There is another story will need to enhance this script to make the doc task oriented instead of task schema oriented. Also, another PR for build script in on-http will have dependency on this PR. So, how about I make up some unit test in the later PR.

yyscamper:
I agree this is just a kind of build script, previously we don't enforce unit-test for such kind of script, for example like some of scripts in on-http/data/templates, I also agree if your script is not written javascript, I will not give the suggestion about unit-testing ^^. So it's not a gating for me. It will be much better if you can still write some unit-test after your enhance story finished.
:+1:

PR #259, opened 2016-06-30
add NTP configuration in esx-ks

lanchongyizu:
**Fix ODR-750** NTP configuration in esx-ks

The solution is adding ntpSevers options to support NTP configuration, so that the users could use the private NTP servers if needed.

@RackHD/corecommitters @panpan0000

PR #267, opened 2016-07-07
add NTP configuration in esx-ks

lanchongyizu:
**Reopen #259 and resolve the conflict with #257.**

Fix ODR-750 NTP configuration in esx-ks

The solution is adding ntpSevers options to support NTP configuration, so that the users could use the private NTP servers if needed.

@RackHD/corecommitters @panpan0000


yyscamper:
:+1:

iceiilin:
+1

PR #274, opened 2016-07-08
add upsert to node task

zhangy42:
@RackHD/corecommitters 
@benbp @pscharla 

Original Story:
https://www.pivotaltracker.com/story/show/121049511


yyscamper:
@zhangy42 : Why invent a new job? I think the existing APIs `POST /nodes` and `PATCH /nodes` do the same work.

zhangy42:
test this please

zhangy42:
@yyscamper: @benbp is looking for something can be automated to modify a node. Please check him for the the detail.

zhangy42:
test this please

jlongever:
the downstream job for on-taskgraph is failing unit-tests with this PR applied.

jlongever:
test this please

yyscamper:
@zhangy42, @benbp : Whether the requirement is only for nodes and only for upsert? If other db items are also required, or delete/patch is also required, I think we need a more general job to cover all those required db items.

Meanwhile, I remembered you had created a job to do REST API request, maybe we can wrap this job to send RackHD self API request. My concern is that we had created `node-api-service` in on-http, if we want to mimic the same operation in on-tasks side, we may have duplicated code in both repositories.

benbp:
@yyscamper agreed, I think it makes sense to generalize (i.e. a base "CRUD" job that we can build off), but at the same time, I worry a majority of feature functionality we _could_ add might go unused, so I encouraged @zhangy42 to start with just a simple "update node" job, with the thought that we can pull out and abstract logic in the future if we notice a need for more flexible database modifications in tasks.

The API request method is an interesting idea, though we don't necessarily make any guarantees that the on-taskgraph process can talk to the on-http process over HTTP (only over AMQP and our database), and there may be additional considerations when we add authentication on by default? 

Basically, I think you raise some good points but I worry the implications might be too complex as a starting point, rather than letting our software designs grow over time when we have more feedback about how our system is being used for these types of tasks (document updates). I think most of the code duplication you mention would really just be calls to model methods, no?

benbp:
👍  changes look good on their own, pending discussion with @yyscamper

yyscamper:
@zhangy42 @benbp : thanks for you explanation, obviously the Rest Api Job is not good for this kind of job, I didn't think much when I gave the proposal.
I also agree with @benbp, to create a generic and powerful job to cover a small requirement is a kind of over-design, I agree we can do this if there is new requirement for other database models.

lib/jobs/upsert-node-job.js
```diff
@@ -0,0 +1,41 @@
...
+    function UpsertNodeAttribute(options, context, taskId) {
+        UpsertNodeAttribute.super_.call(this, logger, options, context, taskId);
+        this.nodeId = this.options.target;
+        this.data = this.options.data;
+    }
+    util.inherits(UpsertNodeAttribute, BaseJob);
+    
+    UpsertNodeAttribute.prototype._run = function run() {
+        var self = this;
+        return waterline.nodes.updateByIdentifier(self.nodeId, self.data)
```
yyscamper: Upsert = update + insert, but your code only does update, so you either need to change the job name or change your code to support both update & insert.

lib/jobs/upsert-node-job.js
```diff
@@ -0,0 +1,41 @@
...
+    )
+);
+
+function UpsertNodeAttributeFactory(BaseJob, util, waterline, Logger) {
+    var logger = Logger.initialize(UpsertNodeAttributeFactory);
+
+    function UpsertNodeAttribute(options, context, taskId) {
+        UpsertNodeAttribute.super_.call(this, logger, options, context, taskId);
+        this.nodeId = this.options.target;
+        this.data = this.options.data;
```
yyscamper: The `data` may contains arbitrary attributes, but we'd better only allow those allowed attributes to be picked.

spec/lib/jobs/upsert-node-job-spec.js
```diff
@@ -0,0 +1,50 @@
...
+            key1:{
+                ip: "123.456.789.012",
+                port: "12345",
+                key2: {
+                    key3: "value"
+                }
+            }
+        } 
+    },
+    sandbox = sinon.sandbox.create(),
```
yyscamper: var sandbox

spec/lib/jobs/upsert-node-job-spec.js
```diff
@@ -0,0 +1,50 @@
...
+        
+        NodeUpsertor = helper.injector.get('Job.Upsert.Node.Attribute');
+        job = new NodeUpsertor(options, context, taskId);    
+        waterline.nodes = {
+            updateByIdentifier: sandbox.stub().resolves()
+        };
+    });
+
+    it('Should modify node collection via waterline', function(){
+        // console.log(waterline.nodes.updateByIdentifier);
```
yyscamper: remove this line

lib/jobs/upsert-node-job.js
```diff
@@ -0,0 +1,41 @@
...
+    'Logger'
+    )
+);
+
+function UpsertNodeAttributeFactory(BaseJob, util, waterline, Logger) {
+    var logger = Logger.initialize(UpsertNodeAttributeFactory);
+
+    function UpsertNodeAttribute(options, context, taskId) {
+        UpsertNodeAttribute.super_.call(this, logger, options, context, taskId);
+        this.nodeId = this.options.target;
```
yyscamper: A suggestion: The context will still pass a `target` to each job ( see https://github.com/RackHD/on-http/blob/master/lib/services/workflow-api-service.js#L77), I think the options.target can be default to `context.target` if no `options.target`, of course this change need remove the `target` from requiredOptions list in base-task.

spec/lib/jobs/upsert-node-job-spec.js
```diff
@@ -0,0 +1,50 @@
...
+            helper.di.simpleWrapper(waterline, 'Services.Waterline')
+        ]);
+        
+        NodeUpsertor = helper.injector.get('Job.Upsert.Node.Attribute');
+        job = new NodeUpsertor(options, context, taskId);    
+        waterline.nodes = {
+            updateByIdentifier: sandbox.stub().resolves()
+        };
+    });
+
```
yyscamper: it's better to restore the sandbox after unit-test.

lib/jobs/upsert-node-job.js
zhangy42: Actually I tested on my own development environment, it can do both update and insert. 

The test code I used to test the feature is as follows.
[update_node_workflow_tasks.zip](https://github.com/RackHD/on-tasks/files/370247/update_node_workflow_tasks.zip)

lib/jobs/upsert-node-job.js
zhangy42: Since this is a use case issue, @benbp , which keys do we allow users to insert/update? Just existing keys for a node in node collections? Are we going to define a valid list of keys?

spec/lib/jobs/upsert-node-job-spec.js
zhangy42: There is var in line 9, and all the variable are separated by comma. For grammar I don't need var here, but do you mean I should change it for another line of variable definition for style reason?

lib/jobs/upsert-node-job.js
zhangy42: This is a use case isssue, this is an interesting feature, we can execute logic with parameter from url. @benbp how do you think?

benbp: @zhangy42 I think @yyscamper's suggestion is good. This way it will be easier to add the task to existing workflows run against a specific node (and invoked via the `/api/2.0/nodes/<id>/workflows` api) without having to manually specify an ID in the workflow options. This is how we handle workflows that run against nodes in most other cases.

spec/lib/jobs/upsert-node-job-spec.js
yyscamper: ohh, got it. I made a mistake, you are right.
I'd prefer add each variable declarative statement in a line if the variable is assigned some complex initial value, this looks more clear comparing with distinguish `,` and `;` . This may be my own preference, so not a must change for your pull request.

```javascript
var job, nodeUpsertor, foo=1, bar='abc';
var taskId = uuid.v4();
var options = {
     ...
};
var sandbox = sinon.sandbox.create();
var waterline = {};
```

lib/jobs/upsert-node-job.js
yyscamper: @zhangy42 : Check the code about `updateByIdnetifier` (https://github.com/RackHD/on-core/blob/master/lib/common/model.js#L311), you will see it will first call `needByIdentifier`, its success depends on the node has been existing.  This is the function in API to create a node: https://github.com/RackHD/on-http/blob/master/lib/services/nodes-api-service.js#L243

If it is to create the node, its nodeId should be generated by RackHD not from user's input. So nodeId is required to update node, but it should be discard to insert a node.

zhangy42: I understand the requirement as follows, if I am wrong please correct me @benbp 

1. add attribute to a node under node collections.
1a: if the attribute exists, then overwrite the data.
1b: if the attribute does not exist, then insert the attribute and data

2. if the node does not exists in Node Collections, then let it go to error.

zhangy42:
test this please

yyscamper:
@zhangy42 : I think I finally understand your design, this job is to upsert attributes of existing node, not to upsert node. 

The latest code is :+1: for me.

benbp:
test this plesae

pscharla:
test this please

pscharla:
@zhangy42 Looks like on-tasks is passing, but two subproject builds are failing:
http://rackhdci.lss.emc.com/job/on-taskgraph/1276/
http://rackhdci.lss.emc.com/job/on-http/2130/

benbp:
test this please

pscharla:
test this please

yyscamper:
@zhangy42 : There is some new code change (in on-tasks repo) about Task-Schema has been merged, so you need to rebase your pull request so that those subproject build can succeed.

lib/jobs/upsert-node-job.js
brianparry: @zhangy42 I think @yyscamper is correct.  updateByIdentifier should throw a NotFoundError if the node does not exist.  There are a few ways to do an upsert.  One is to use a native mongo update with the `upsert` option.  Another options is to create an upsert by chaining a waterline update and create together.True upserts also require an index on the primary key to prevent duplicate records from getting inserted.

zhangy42:
Okay, I will do it!

zhangy42:
Let me try another PR and see whether it works, it looks like after rebase this PR is having whole bunch of changes that make it hard to debug. Closing this PR for this moment.

PR #276, opened 2016-07-12
Change Graph.InstallEsx to Graph.InstallESXi

anhou:
resolve RackHD/RackHD#218
taksgraph change https://github.com/RackHD/on-taskgraph/pull/139

@RackHD/corecommitters

### DIFF
--- a/lib/task-data/tasks/install-esx.js
+++ b/lib/task-data/tasks/install-esx.js
@@ -4,7 +4,7 @@
 
 module.exports = {
     friendlyName: 'Install Esx',
-    injectableName: 'Task.Os.Install.Esx',
+    injectableName: 'Task.Os.Install.ESXi',
     implementsTask: 'Task.Base.Os.Install',
     schemaRef: 'rackhd/schemas/v1/tasks/install-esxi',
     options: {


lib/task-data/tasks/install-esx.js
```diff
@@ -4,7 +4,7 @@
 
 module.exports = {
     friendlyName: 'Install Esx',
-    injectableName: 'Task.Os.Install.Esx',
+    injectableName: 'Task.Os.Install.ESXi',
```
yyscamper: I think other places need also be changed:
-  The filename.
- the kickstart/boot-config file name.
- the osType.
- the properties.os.hypervisor.

anhou: they're not intended to be included in this PR, these PRs are only for user's interfaces,  not for code details. suggest to refine them during developing bit by bit.

iceiilin:
:+1:

lib/task-data/tasks/install-esx.js
yyscamper: OK, make sense.

yyscamper:
:+1:

PR #277, opened 2016-07-12
Update Task Annotation from task oriented instead of schema

WangWinson:
The update mainly focus on 
- Task oriented annotation
- Add task options value as default to schema
- Refine `description` field

@RackHD/corecommitters @iceiilin @cgx027 @pengz1   

Screen shot of the Task Doc page:
image
image

### DIFF
--- a/task-annotation-generator.js
+++ b/task-annotation-generator.js
@@ -26,11 +26,12 @@ TaskAnnotation.run = function (tasks, validator) {
         var tn = new TaskAnnotation(task);
         var schemaResolved = validator.getSchemaResolved(task.schemaRef);
         var schemaMerged = tn.mergeSchema(schemaResolved);
+        tn.addDefault(schemaMerged);
         return tn.generateDocData(schemaMerged, {
-            url: '/' + task.injectableName,
+            // url: '/' + task.injectableName,
             name: 'option',
             title: 'option',
-            group: task.injectableName,
+            group: task.injectableName.replace(/\./g, '_'),
             groupTitle: task.friendlyName
         });
     })
@@ -79,6 +80,20 @@ TaskAnnotation.prototype.mergeSchema = function (obj) {
 };
 
 /**
+ * add default value from task.options to schema
+ * @param  {Object} obj - JSON schema object
+ */
+TaskAnnotation.prototype.addDefault = function (obj) {
+    var self = this;
+    _.forOwn(self.task.options, function (value, name) {
+        var op = _.get(obj, 'properties.' + name);
+        if (op && op instanceof Object) {
+            op.default = value;
+        }
+    });
+};
+
+/**
  * Parse the JSON schema and generate doc data recusively.
  * The doc data will be rendered in apiDoc template
  *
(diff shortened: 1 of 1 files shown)


yyscamper:
jenkins: test this please

yyscamper:
@WangWinson : could you please attach some screenshot about the view of task annotation?

task-annotation-generator.js
```diff
@@ -4,11 +4,43 @@
...
+    .map(function (task) {
+        var tn = new TaskAnnotation(task);
+        var schemaResolved = validator.getSchemaResolved(task.schemaRef);
+        var schemaMerged = tn.mergeSchema(schemaResolved);
+        tn.addDefault(schemaMerged);
+        return tn.generateDocData(schemaMerged, {
+            // url: '/' + task.injectableName,
+            name: 'option',
+            title: 'option',
+            group: task.injectableName.replace(/\./g, '_'),
```
yyscamper: Why replace `.` with `_` for injectableName?

task-annotation-generator.js
```diff
@@ -48,6 +80,20 @@ TaskAnnotation.prototype.mergeSchema = function (obj) {
...
 
 /**
+ * add default value from task.options to schema
+ * @param  {Object} obj - JSON schema object
+ */
+TaskAnnotation.prototype.addDefault = function (obj) {
+    var self = this;
+    _.forOwn(self.task.options, function (value, name) {
+        var op = _.get(obj, 'properties.' + name);
+        if (op) {
```
yyscamper: Some default value may be 0 or null or empty string (null or empty string is in rare, but 0 is common) , but this judgement will prevent those value from being presented in task annotation.

task-annotation-generator.js
```diff
@@ -160,8 +225,11 @@ if (require.main === module) {
...
+    var tasks = helper.requireGlob('/lib/task-data/tasks/*.js');
 
-    var taskAnnotation = new TaskAnnotation(validator);
-    taskAnnotation.run();
-    console.log('========= task_doc_data.json generated =======');
+    TaskAnnotation.run(tasks, validator)
+    .then(function (docData) {
+        fs.writeFileSync('task_doc_data.json', JSON.stringify(docData));
+        console.log('========= task_doc_data.json generated =======');
+    });
```
yyscamper: Suggest add a `catch` to capture the script error:
```javascript
.catch(function(err) {
    console.error('xxxxxx');
    process.exit(-1);
});
```

task-annotation-generator.js
WangWinson: The `injectableName` here would finally be using as an anchor in HTML.
For example, 
```HTML
<a href="#api-Task.Os.Install.Ubuntu" />
``` 
The anchor originally has scrolling animation, but it broke when I use the anchor link as above.
It works when I change to 
```HTML
<a href="#api-Task_Os_Install_Ubuntu" />
```
But I didn't dig into the root cause

task-annotation-generator.js
WangWinson: Good Catch !

task-annotation-generator.js
WangWinson: Good suggestion.

yyscamper:
:+1:

anhou:
:+1:

iceiilin:
:+1:

PR #278, opened 2016-07-12
Add net imgae and interface options to Ubuntu install Task

keedya:
depends on :https://github.com/RackHD/on-http/pull/364

@RackHD/corecommitters @jlongever @zyoung51 @uppalk1 @tannoa2


jlongever:
test this please

lib/task-data/tasks/install-ubuntu.js
```diff
@@ -17,7 +17,15 @@ module.exports = {
         completionUri: 'renasar-ansible.pub',
         version: 'trusty',
         repo: '{{api.server}}/ubuntu',
-        rootPassword: "RackHDRocks!"
+        netImage: '{{api.server}}/ubuntu/install/filesystem.squashfs',
+        rootPassword: "RackHDRocks!",
+        rootSshKey: null,
```
yyscamper: All `null` and `[]` should be removed from options, because these value will violate the task-schema. 

If task-schema is enabled (there is `schemaRef` in the task definition), only those option that has valid default value can be put in the `options` list, if user want to check the full supported options, user should turn to lookup the schema.

Since you add a new option `interface`, you need to update the task-schema. ( this is the schema for ubuntu: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-os-general.json,  you can copy centos to add a ubuntu specific option: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-centos.json).

yyscamper: The task-schema is relative new feature and now only have only been enabled for majority of install os tasks and obm control tasks, and we still lack a document to describe how to use and define the task-schema.  So all your reference document is the code itself at present. Of course, you are welcome if you have any question about task-schema, either code or unit-test.

amymullins:
@yyscamper at what point do you anticipate documentation for the parts of the feature that have been merged and also what point will task-schema be "enabled" as you mention above?  I'm curious about the "rollout" plan so this can be considered for future dev.

yyscamper:
@amymullins : 
Actually the task-schema has already enabled for some tasks, such as install-os and obm-control. The remaining tasks in my queue to officially announce the enabling task-schema are:
- Change the task-schema to be managed using namespace, just like schemas for redfish API and API 2.0.
- Add document for task-schema design.
- Add some tutorial about how to create the task-schema and unit-test.

The task-schema design progress has some impact by the newly emerged but higher priority @ scale issue. But we will keep developing after issue resolved.

At present, my practice is: if the code change only touches the task without schema already being defined, it's not a gating for me if the code doesn't create the schema for it. But if the code change touches the task that already have  task-schema been defined, then the code must change the schema accordingly if necessary, because the task may not work well if it don't have a good schema file.

amymullins:
@yyscamper thanks for the detail!  If there are tasks that have the schema enabled and will gate a PR, can we get that part announced/documented and should make sure it has the unit+functional tests (this may already, I hadn't checked yet) ?  That way, people are aware of the changes and can code to the schema correctly.

yyscamper:
@amymullins : unit-test has been present for all schemas we have created until now, but we don't have functional tests, we can do it later.
For @keedya 's case, it does reflect a problem that we lack the unit-testing for the default value in options that I wasn't aware of before. I have created story to track this problem: https://www.pivotaltracker.com/projects/1492892/stories/126194769. If this story is done, those invalid default value will be caught by unit-test.

PR #280, opened 2016-07-13
Ubuntu installation

keedya:
- Change base url into an option set by user
- Remove netImage and adding kargs(allowing net image to be optional use case: install ubuntu from mounted ISO)
depends on : https://github.com/RackHD/on-http/pull/364

@RackHD/corecommitters @brianparry @zyoung51 @jlongever @tannoa2

### DIFF
--- a/lib/jobs/clean-work-items.js
+++ b/lib/jobs/clean-work-items.js
@@ -62,7 +62,7 @@ function cleanWorkItemsJobFactory(BaseJob, waterline, Logger, util, uuid, Promis
         }
         var expiry = new Date(Date.now() - self.options.leaseAdjust);
         return waterline.workitems.findExpired(expiry).then(function (workItems) {
-            return waterline.workitems.setFailed(null, workItems);
+            return waterline.workitems.setFailed(null, null, workItems);
         }).then(function () {
             return Promise.delay(self.options.queuePollInterval);
         }).then(function () {
--- a/lib/jobs/ipmi-job.js
+++ b/lib/jobs/ipmi-job.js
@@ -15,7 +15,8 @@ di.annotate(ipmiJobFactory, new di.Inject(
     'Assert',
     'Promise',
     '_',
-    'Services.Waterline'
+    'Services.Waterline',
+    'JobUtils.PollerHelper'
 ));
 
 function ipmiJobFactory(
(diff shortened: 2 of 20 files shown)


yyscamper:
@keedya : As I said in the google group (https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/rackhd/Tx_fd9PQvpQ), please update the schema file accordingly to ensure the integrity of task-schema and task definition.

keedya:
@yyscamper, this PR is to make sure we fix all the errors in Ubuntu installation workflow. I suggest to leave this effort to a different effort in a different story. is that ok with you?

yyscamper:
@keedya : The task schema change is easy because you just need to add the new options rather than creating a new schema.  If move the effort to another story, that story's priority may become low and thus may cause the code divergence for a long time. So I still recommend you to update the schema in this PR.

You can copy The centos schema (https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-centos.json#L9), create UbuntuSpecificOptions to contain all your newly add options.

yyscamper:
@keedya : @WangWinson told me that the failure of Ubuntu installation you encountered is because that  we had not tested the ubuntu ISO, our installation testing was using the ubuntu mirror, such as http://archive.ubuntu.com/ubuntu/ .

Thanks for your fix.

keedya:
@yyscamper code updated.
@RackHD/corecommitters @brianparry @jlongever @zyoung51 @tannoa2

keedya:
@yyscamper there was errors that are not related to the ubuntu ISO.
Previous code had 2 major failures and a missed configuration:
1- doesn't setup users correctly and fails ssh validation
2- if encountered partitionned disk, preceed will fail
3- dhcp server setup needed to be documented and changes needed for our vagrant setup(used for our jenkins regression tests)

yyscamper:
@keedya : Obviously we had missed some test items (#1/#2) even using the ubuntu mirrors. @WangWinson had noticed the problem about #3 but forgot to update the document.

Meanwhile, thanks for your fix of task-schema, that's very good.  The only problem is that you forget tomention the `kargs` in schema. That will not impact the ubuntu workflow because we defaultly allow the additionalProperties, but the task-annotation will not show this option to user. Please complement this if your time is avaiable.

WangWinson:
For 3, there is a story https://www.pivotaltracker.com/n/projects/1492892/stories/120504923 to track

PR #282, opened 2016-07-14
Replace references to 1.1 API with current

brianparry:
@RackHD/corecommitters @keedya @dalebremner @BillyAbildgaard

### DIFF
--- a/lib/jobs/secure-erase-drive.js
+++ b/lib/jobs/secure-erase-drive.js
@@ -372,7 +372,7 @@ function secureEraseJobFactory(
     output will be:
     [
         {
-            downloadUrl: "/api/1.1/templates/secure_erase.py",
+            downloadUrl: "/api/current/templates/secure_erase.py",
             cmd: "sudo python secure_erase.py -d '{\"diskName\": \"/dev/sda\",
                 \"virtualDisk\": \"/c0/v0\",\"deviceIds\": [23],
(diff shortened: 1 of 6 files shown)


keedya:
:+1:

dalebremner:
+1

lib/jobs/secure-erase-drive.js
```diff
@@ -407,7 +407,7 @@ function secureEraseJobFactory(
         }, []);
 
         // add script's url in the first command
-        this.commands[0].downloadUrl = '/api/1.1/templates/secure_erase.py';
+        this.commands[0].downloadUrl = '/api/current/templates/secure_erase.py';
```
benbp: I think it would be easier to use some of the built-in API render variables we have, so that we only have to change this in one place in the future, e.g.

```
downloadUrl = '{{ api.templates }}/secure_erase.py'
```

See https://github.com/RackHD/on-tasks/blob/master/lib/task.js#L120

yyscamper: Agree

keedya: @benbp : {{ api.templates }} will duplicate the base url  and it won't work in this case

api.templates = self.renderContext.api.server + '/api/current/templates

benbp: Ah, bummer. Ok well 👍

PR #284, opened 2016-07-20
Add southbound API to pass notification to RackHD (refer to ODR682)

cgx027:
This is the PR to support https://github.com/RackHD/on-http/pull/377

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -46,6 +46,7 @@ function installOsJobFactory(
 
         self.nodeId = self.context.target;
         self.profile = self.options.profile;
+        self.taskId = taskId;
 
         //OS repository analyze job may pass some options via shared context
         //The value from shared context will override the value in self options.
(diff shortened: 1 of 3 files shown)


yyscamper:
@tannoa2 This is the notification API that you want.

iceiilin:
Looks great to me

yyscamper:
Please change the wait-completion-uri job as well: https://github.com/RackHD/on-tasks/blob/master/lib/jobs/wait-completion-uri.js#L57

lib/jobs/install-os.js
```diff
@@ -236,6 +237,18 @@ function installOsJobFactory(
                     }
                 }
             });
+
+            self._subscribeTaskNotification(function(data) {
+                assert.object(data);
+                if(
+                    data.nodeId === self.nodeId
+                    && data.taskId === self.taskId
+                    && data.type === 'osInstall'
```
yyscamper: I have some concern to judge the `type` here.
- the taskId is enough to tell this is install os job;
- this makes  the code cannot be shared by other job that also wants notification, for example, for wait-completion-uri job, to define another type "waitCompletionUri"? I think it's not worth.

cgx027:
Remove WIP flag.

WangWinson:
:+1:

yyscamper:
:+1:

yyscamper:
jenkins: test this please

PR #286, opened 2016-07-20
Add PDU Node relations using OBM services data

keedya:
- Adds pdu relation to node only if this particular PDU is added to the obm services list

### DIFF
--- a/lib/jobs/pdu-node-relations.js
+++ b/lib/jobs/pdu-node-relations.js
@@ -1,0 +1,199 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = pduRelationsJobFactory;
+di.annotate(pduRelationsJobFactory, new di.Provide('Job.Catalog.PduRelations'));
+di.annotate(pduRelationsJobFactory, new di.Inject(
+    'Job.Base',
+    'Services.Waterline',
+    'Logger',
+    'Util',
+    'Constants',
+    'Promise',
+    '_'
+));
+
+function pduRelationsJobFactory(
+    BaseJob,
+    waterline,
+    Logger,
+    util,
+    Constants,
+    Promise,
+    _
+) {
+
+    var logger = Logger.initialize(pduRelationsJobFactory);
+
+    /**
+     *
+     * @param {Object} [options]
+     * @constructor
+     */
+    function PduRelationsJob(options, context, taskId) {
+        PduRelationsJob.super_.call(this, logger, options, context, taskId);
+
+        this.nodeId = context.target || options.nodeId;
+    }
+
+    util.inherits(PduRelationsJob, BaseJob);
+
+    /**
+     * Globals
+     */
+
+    var NODE_RELATION_TYPE = 'poweredBy';
+    var PDU_RELATION_TYPE = 'powers';
+
+    /**
+     * @memberOf PduRelationsJob
+     */
+    PduRelationsJob.prototype._run = function run() {
+        var self = this;
+        return waterline.nodes.needByIdentifier(self.nodeId)
+        .then(function (node) {
+            if (node.type !== Constants.NodeTypes.Pdu) {
+                throw new Error('Task runs only on PDU nodes');
+            }
+            var pduHost = _adjustIpStrFormat(node.snmpSettings.host);
+            return self._searchObmServices(pduHost, node);
+        })
+        .then(function () {
+            self._done();
+        })
+        .catch(function (err) {
+            self._done(err);
+        });
+
+    };
+
+    PduRelationsJob.prototype._searchObmServices =
+        function _searchObmServices(pduHost, pduNode) {
+            var self = this;
+            return Promise.try(function () {
+                return waterline.obms.find();
+            })
+            .filter(function (obm) {
+                return (_validateIpAddressFormat(obm.config.host));
+            })
+            .map(function (result) {
+                /* Compare hosts
+                 * If PDu host and the pdu obm service host are equal
+                 * Then create the relation between node and discoverd pdu
+                */
+                var obmServiceHost = _adjustIpStrFormat(result.config.host);
+                if (pduHost === obmServiceHost) {
+                    //find the node
+                    return self.updatePoweredNodeRelations(result.node)
+                   .then(function () {
+                       return self.updatePduNodeRelations(pduNode, result.node);
+                   });
+                }
+            });
+        };
+
+    PduRelationsJob.prototype.updatePoweredNodeRelations =
+        function _updatePoweredNodeRelations(poweredNodeId) {
+            var self = this;
+            var targetNode;
+            return waterline.nodes.findByIdentifier(poweredNodeId)
+            .then(function (node) {
+                if (!node) {
+                    throw new Error('Could not find node with identifier ' + node);
+                }
+                targetNode = node;
+                return node.relations;
+            })
+            .filter(function (entry) {
+                //return entries except the one with relation type = poweredBy
+                return (entry.relationType !== NODE_RELATION_TYPE);
+            })
+            .then(function (relations) {
+                //update PDU relation type powers
+                var targets = _getTargets(targetNode, NODE_RELATION_TYPE);
+                if (targets.indexOf(self.nodeId) < 0) {
+                    targets.push(self.nodeId);
+                }
+                relations.push({
+                    relationType: NODE_RELATION_TYPE,
+                    targets: targets
+                });
+                return waterline.nodes.updateByIdentifier(
+                    poweredNodeId,
+                    { relations: relations });
+            });
+        };
+
+    PduRelationsJob.prototype.updatePduNodeRelations =
+        function _updatePduNodeRelations(pduNode, poweredNodeId) {
+            return Promise.try(function () {
+                return pduNode.relations;
+            })
+            .filter(function (entry) {
+                //return entries except the one with relation type = powers
+                return (entry.relationType !== PDU_RELATION_TYPE);
+            })
+            .then(function (pduRelations) {
+                //update PDU relation type powers
+                var targets = _getTargets(pduNode, PDU_RELATION_TYPE);
+                if (targets.indexOf(poweredNodeId) < 0) {
+                    targets.push(poweredNodeId);
+                }
+                pduRelations.push({
+                    relationType: PDU_RELATION_TYPE,
+                    targets: targets
+                });
+                return waterline.nodes.updateByIdentifier(
+                    pduNode.id,
+                    { relations: pduRelations });
+            });
+        };
+
+    /**
+     * verify ip address is 4 dec nbs separated by '.'
+     * returns a booleen
+     */
+    function _validateIpAddressFormat(ipString) {
+        var regex = /^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$/;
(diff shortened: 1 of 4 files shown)


keedya:
@RackHD/corecommitters @dalebremner @BillyAbildgaard @srinia6 @brianparry @zyoung51

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     *
+     * @param {Object} [options]
+     * @constructor
+     */
+    function PduRelationsJob(options, context, taskId) {
+        PduRelationsJob.super_.call(this, logger, options, context, taskId);
+
+        this.nodeId = context.target || options.nodeId;
+        assert.isMongoId(this.nodeId);
```
yyscamper: Don't do this, The RackHD will support multi backends database,  then the nodeId will be set as uuid not MongoId.

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+     */
+
+    var RELATION_TYPE = 'poweredBy';
+
+    /**
+     * @memberOf PduRelationsJob
+     */
+    PduRelationsJob.prototype._run = function run() {
+        var self = this;
+        return waterline.nodes.findByIdentifier(self.nodeId)
```
yyscamper: Change to `needByIdentifier`. `needByIdentifier` will throw error if the node doesn't exist, however `findByIdentifier` will return an empty.

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+        .then(function (node) {
+            if (node.type !== 'pdu') {
```
yyscamper: Change to use Constant rather than direct string: https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L316

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+
+            });
+        };
+
+    /**
+     * verify ip address is 4 dec nbs separated by '.'
+     * returns a booleen
+     */
+    function _validateIpAddressFormat(ipString) {
+        var regex = /^(?:[0-9]{1,3}\.){3}[0-9]{1,3}$/;
```
yyscamper: There is already a Constant for ipAddress: https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L215

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+            var self = this;
+            return Promise.try(function () {
+                return waterline.obms.find();
+            })
+            .filter(function (obm) {
+                return (_validateIpAddressFormat(obm.config.host));
+            })
+            .map(function (result) {
+                //compare hosts
+                var obmServiceHost = _adjustIpStrFormat(result.config.host);
```
yyscamper: Could you please add comment why need do this?

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+        .catch(function (err) {
+            self._done(err);
+        });
+
+    };
+
+    PduRelationsJob.prototype._searchObmServices =
+        function _searchObmServices(pduHost) {
+            var self = this;
+            return Promise.try(function () {
```
yyscamper: You don't need Promise.try as `waterline.obms.find()` has already returned a promise.

lib/jobs/pdu-node-relations.js
keedya: I do need it to use the filters in promises

lib/jobs/pdu-node-relations.js
keedya: I do need it to use the filters in promises

lib/jobs/pdu-node-relations.js
keedya: @yyscamper they both work do you agree?

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+            })
+            .map(function (result) {
+                //compare hosts
+                var obmServiceHost = _adjustIpStrFormat(result.config.host);
+                if (pduHost === obmServiceHost) {
+                    //find the node
+                    return waterline.nodes.findByIdentifier(result.node)
+                   .then(function (node) {
+                       if (!node) {
+                           throw new Error('Could not find node with identifier ' + result.node);
```
yyscamper: A little suggestion, if you change previous `findByIdentifier` to `needByIdentifier`, you don't need this throw.

lib/jobs/pdu-node-relations.js
anhou: there is a check for returned value in needByIdentifier, it's 'node' here, if don't use needByIdentifier , need a check before 'node.type' != 'pdu',  otherwise, if there's no 'node' found by 'nodeId'  there will be error like 'TypeError: Cannot read property 'type' of undefined'

keedya:
@yyscamper thanks for your review. I addressed your comments. I will use needByIdentifier in the future. In this task, I wrote unit test around findByIdentifier. I will prefer to leave it in this task unless you think otherwise?
Thanks :)

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,142 @@
...
+                       }
+                       return node.relations;
+                   })
+                   .filter(function (entry) {
+                       return (entry.targets.indexOf(self.nodeId) < 0);
+                   })
+                   .then(function (relations) {
+                       relations.push({
+                           relationType: RELATION_TYPE,
```
yyscamper: If you add `poweredBy` relation to node, then the counterpart `powers` relation should be added into the pdu node. But you forgot the latter relation.

lib/jobs/pdu-node-relations.js
yyscamper: Ahh, I didn't realize this problem. the promise that returned from waterline isn't powerful than bluebird promise.

lib/jobs/pdu-node-relations.js
yyscamper: @keedya : Either is OK. 'findByIdentifer' + Output Check = 'needByIdentifier'

lib/jobs/pdu-node-relations.js
yyscamper: You can reference how the enclosure relation is handled: https://github.com/RackHD/on-tasks/blob/master/lib/jobs/generate-enclosure.js

anhou:
@keedya  I suggest replacing  findByIdentifier to needByIdentifier in this PR, it's easy to just  replace name findByIdentifier to needByIdentifer in batches in file, and unittest cases wouldn't be affected because the mock method is the same.

anhou:
:+1: with comments updated

lib/jobs/pdu-node-relations.js
yyscamper: I think you are copying the design of switch-node-relation job, but it still doesn't resolve the bi-direction relations, so it still need some fix.

The relation is always bi-direction, if you relates A with B by "poweredBy" and update the A's document, then you should still relates B with A by "powers" and update B's document. 

I admit this is a little complicated, but this is exactly the design with enclosure relation. Either build or remove relation is not easy.

I just come up an idea, copy the current `obm` design, detach the relations to a new kind of resource, this will simplify a lot code logic to either build or remove relations.

keedya: @yyscamper I am working on the changes. I will update the PR shortly
Thanks :)

yyscamper:
@keedya : either needByIdentifier or findByIdentifier is OK for me, in your case, you can save some code to use needByIdentfier. If you choose findByIdentifer, don't forget the empty check.

keedya:
Address @yyscamper and @anhou comments

keedya:
Adding @benbp

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,191 @@
...
+                });
+                return waterline.nodes.updateByIdentifier(
+                    poweredNodeId,
+                    { relations: relations });
+            });
+        };
+
+    PduRelationsJob.prototype.updatePduNodeRelations =
+        function _updatePduNodeRelations(pduNode, poweredNodeId) {
+            Promise.try(function () {
```
yyscamper: This promise need be return

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,191 @@
...
+                //return entries except the one with relation type = powers
+                return (entry.relationType !== PDU_RELATION_TYPE);
+            })
+            .then(function (pduRelations) {
+                //update PDU relation type powers
+                var targets = _getTargets(pduNode, PDU_RELATION_TYPE);
+                if (targets.indexOf(poweredNodeId) < 0) {
+                    targets.push(poweredNodeId);
+                }
+                pduRelations.push({
```
yyscamper: If the pduRelations is empty, you can push a new `powers` relation record. But if the pudRelations is not empty, that means you don't need to create a new relation record, just update the existing relation record and insert the new nodeId to targets is enough.

lib/jobs/pdu-node-relations.js
```diff
@@ -0,0 +1,191 @@
...
+                    relationType: PDU_RELATION_TYPE,
+                    targets: targets
+                });
+                return waterline.nodes.updateByIdentifier(
```
yyscamper: If the relation has already been created and no change, we should prevent from updating the node document. But I see your code logic is whenever the relation is existing or not, create a new one and always update the node document.

yyscamper: The same for function "updatePoweredNodeRelations "

yyscamper:
@keedya : Please make sure at least following test cases have been passed using your code:
- [ ] 1 PDU + 1 NODE: no relations before, then after job running, both nodes have been created correct relation.
- [ ]  1 PDU + 2 NODEs: no relations before, then after job running, 3 nodes have been created correct relations. The PDU should have only one power relation with the `targets` contains both nodes.
- [ ] 1 PDU + 1 NODE:  The power relation has existed, after job running, no change for both nodes' document.
- [ ] 1 PDU + 1 NODE:   The PDU has been created relation with another node before, after job running, the PDU node relation will be updated, but shouldn't create another new power relation.

yyscamper:
@keedya : Whether you only want to run this job in pdu discovery? If that's true, we don't need to consider too much cases because no power relations existing before your job runs.
If so, please ignore my above 2 comments, the only comment you need fix is let the promise be returned.

:+1: after fix the promise not being returned.

keedya:
@yuscamper the logic is to use the create a new one and add existing targets. Same logic. I also return a promise within promise.try()

lib/jobs/pdu-node-relations.js
yyscamper: Just like you do in line76

yyscamper:
@keedya : The promise that need to be returned is the promise.try itself, not the one within promise.try. Below function's return value is `undefined` with a side effect of a floating promise, the promise will be run in background.
```javascript
function() {
   Promise.try(...);
}
```

keedya:
@yyscamper all set!!

yyscamper:
Thanks, :+1:

PR #287, opened 2016-07-20
[SB Redfish] Set Redfish system identifiers

jlongever:
These changes are to support active discovery of new Redfish System resources.

- Build MAC identifiers list using  System EthernetInterfaces resource members.
- Update unit-tests.

@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @zyoung51

### DIFF
--- a/lib/jobs/redfish-discovery.js
+++ b/lib/jobs/redfish-discovery.js
@@ -192,6 +192,7 @@ function RedfishDiscoveryJobFactory(
                     config: config,
                     service: 'redfish-obm-service'
                 };
+                
                 return waterline.obms.upsertByNode(n.id, obm)
                 .then(function() {
                     return n;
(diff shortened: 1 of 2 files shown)


jlongever:
test this please

anhou:
:+1:

anhou:
test this please

lib/jobs/redfish-discovery.js
```diff
@@ -245,33 +246,67 @@ function RedfishDiscoveryJobFactory(
...
-            return self.upsertRelations(node, relations)
-            .then(function(nodeRecord) {
-                return Promise.all([
-                    waterline.obms.upsertByNode(nodeRecord.id, obm),
-                    nodeRecord
-                ]);
+            return self.redfish.clientRequest(config.root)
+            .then(function(res) {
+                var ethernet = _.get(res.body, 'EthernetInterfaces');
+                if(ethernet) {
```
yyscamper: The `[]` is also a truth value, to enhance the judging criteria can prevent some extra and useless request from being sent if the `ethernet` is empty array.

lib/jobs/redfish-discovery.js
```diff
@@ -245,33 +246,67 @@ function RedfishDiscoveryJobFactory(
...
+                    identifiers: identifiers
+                };
+                
+                var relations = [{
+                    relationType: 'enclosedBy',
+                    targets: targetList
+                }];
+
+                var obm = {
+                    config: config,
```
yyscamper: A lot properties in the `config`, are they all necessary for redfish-obm-service? If not all, please only pick those necessary properties before upsert the obm.

lib/jobs/redfish-discovery.js
yyscamper: From the unit-testing code, the `EthernetInterfaces` is object rather than array, if so, ignore my above comment.

yyscamper:
:+1: after addressing the comment of picking only necessary properties of redfish-obm-service.

yyscamper:
Meanwhile, @jlongever , do you have ever consider to refactor the code to use RxJS? With so many powerful operators in RxJS, I think the code logic can be more simple. The long & nesting promise chains is a little hard to understand.

This is not a requirement for this PR, just raise some discussion.

anhou:
test this please

lib/jobs/redfish-discovery.js
jlongever: Yes they are necessary.

lib/jobs/redfish-discovery.js
jlongever: Yes, Redfish spec defines EthernetInterfaces as an object type,   it should never be an array.

PR #297, opened 2016-08-01
modify wait-completion-uri to make use of task notification

cgx027:
Modify wait-completion-uri to make use of task notification.

The old fashion completion uri will still work, but should be removed after all OSes are modified to use task notification instead of completion URL.


yyscamper:
:+1:

WangWinson:
:+1:

cgx027:
@yyscamper @RackHD/corecommitters @lanchongyizu

cgx027:
Close this PR due to design change.

PR #299, opened 2016-08-01
[PDU] Check SKU config for type/mibs

jlongever:
- Add SKU config checks to define the PDU type.
- Add SKU config checks to define MIB list for Sinetica PDUs. 
- Update unit-tests

@RackHD/corecommitters @uppalk1 @tannoa2 @keedya @zyoung51


anhou:
:+1:

PR #300, opened 2016-08-02
Update context in compose job

jlongever:
- This allows newly composed systems be passed to other related tasks under the compose workflow (i.e. create pollers).

@RackHD/corecommitters @uppalk1 @zyoung51 @tannoa2 @keedya


anhou:
:+1:

PR #301, opened 2016-08-02
Collect element power telemetry data

jlongever:
- Adds element.power command support to gather element Power telemetry. 

@RackHD/corecommitters @zyoung51 @tannoa2 @keedya @uppalk1


tannoa2:
+1

PR #307, opened 2016-08-04
coreos install: change completionUri to SSH key

codenrhoden:
Like most of the other OS install tasks, use renesar-ansible.pub
as the completion URI. That way we can indicate that the CoreOS
install is finished when it has actually written the OS to disk
instead of when it downloads the cloud-config file (which happens
at the very beginning, not the end).

### DIFF
--- a/lib/task-data/tasks/install-coreos.js
+++ b/lib/task-data/tasks/install-coreos.js
@@ -15,7 +15,7 @@ module.exports = {
         comport: 'ttyS0',
         hostname: 'coreos-node',
         installDisk: '/dev/sda',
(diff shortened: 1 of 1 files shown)


codenrhoden:
This PR goes together with https://github.com/RackHD/on-http/pull/404

The idea is to have the OS install task not go to "succeeded" state until the OS has actually written to disk. Do this by doing a wget on the SSH key instead of the cloud config file, as that can happen at the very end instead of at the very beginning.

heckj:
👍

yyscamper:
:+1:

yyscamper:
jenkins: test this please

PR #312, opened 2016-08-09
Add support to handle array type when catalogging elements

tannoa2:
@jlongever @uppalk1 @zyoung51 @RackHD/corecommitters @amymullins


lib/jobs/emc-redfish-catalog.js
```diff
@@ -251,8 +251,14 @@ function EmcRedfishCatalogJobFactory(
             var newElement = {};
             _.forEach(element,function(value,key){
                 if(key === "Processors" || key === "Vpd" || key === "Dimms" || key === "Drives") {
-                    newElement [key]=   self.updatingComponents(element,key);
-                } else {
+                    if(value instanceof Array){
+                        newElement [key]=   self.updatingComponentsArray(element,key);
```
benbp: Kind of some weird formatting here, even if the old code was already that way.

benbp:
👍

yyscamper:
:+1: with same format suggestion of Ben

PR #313, opened 2016-08-09
Decrypt obm settings when testing OBM configs in create-X-obm-setting…

benbp:
…s jobs

When we migrated OBM decryption over to the model layer, we forgot about this piece of code.

Requires https://github.com/RackHD/on-core/pull/192

Edit: updated for @yyscamper's comments.

@RackHD/corecommitters @hohene @johren

### DIFF
--- a/spec/lib/jobs/create-obm-settings-spec.js
+++ b/spec/lib/jobs/create-obm-settings-spec.js
@@ -1,5 +1,4 @@
-// Copyright 2015, EMC, Inc.
-/* jshint node:true */
+// Copyright 2015-2016, EMC, Inc.
 
 'use strict';
 
@@ -27,6 +26,9 @@ describe(require('path').basename(__filename), function () {
         waterline.obms = {
             upsertByNode: sinon.stub().resolves()
         };
+
+        var encryption = helper.injector.get('Services.Encryption');
+        return encryption.start();
     });
 
     describe('Base', function () {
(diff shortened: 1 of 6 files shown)


heckj:
👍

benbp:
test this please

spec/lib/jobs/create-obm-settings-spec.js
```diff
@@ -65,12 +69,45 @@ describe(require('path').basename(__filename), function () {
...
+
+            var ObmService = helper.injector.get('Task.Services.OBM');
+            self.sandbox.stub(ObmService, 'create').returns({
+                powerStatus: Promise.method(function() {})
+            });
+
+            var jobOptions = {
+                service: 'ipmi-obm-service',
+                config: {
+                    alias: 'test',
```
yyscamper: Why add `alias` here? I think this may confuse user if user reads the unit-test to study how to use this function.

yyscamper:
I'd support move the decryption to on-core's model, I am in favor to hide the encryption & decryption in model level and being transparent to job and other modules.

Meanwhile, I checked the function `liveTestIpmiConfig`, it created ObmService with input obmConfig, but if we use this function [waterline.obms.findByNode](https://github.com/RackHD/on-core/blob/master/lib/models/obms.js#L123) with the `reveal` set to true, we can not only get ride of the helpers.revealSecret but also verify whether the obm setting is upserted correctly.

benbp:
Makes sense @yyscamper. I'll do a re-work today.

spec/lib/jobs/create-obm-settings-spec.js
benbp: Whoops!

benbp:
@yyscamper updated, I added a reveal option to the `models.obm.upsertByNode` method (instead of findByNode, this way we avoid an extra database call) in https://github.com/RackHD/on-core/pull/192, and also removed the now unused revealSecrets function in the job-helper module.

yyscamper:
@benbp : It's good for me, :+1:

PR #314, opened 2016-08-11
Add task upsert node

zhangy42:
@RackHD/corecommitters 
@benbp @yyscamper @pscharla 

This is a cleaner PR that helps debug the Jenkins issues after rebase.

Original PR is here https://github.com/RackHD/on-tasks/pull/274.


zhangy42:
test this please

zhangy42:
I don't under stand the error message, any help?

yyscamper:
@zhangy42: don't worry, all other PRs have same Jenkins failure, so it's not related with your change.

yyscamper:
jenkins: test this please

yyscamper:
jenkins: test this please

changev:
jenkins: test this please

pscharla:
test this please

brianparry:
@zhangy42 As mentioned by @yyscamper in the original PR, I'd suggest changing the name from upsert to update to avoid confusion with Mongo upsert.

yyscamper:
@brianparry : I think the PR title confused both me and you, I finally understand why @zhangy42 choosed the `upsert` not `update`.  The job's injectableName is `Upsert.Node.Attribute` not `Upsert.Node`. So if standing in the point of node attributes, the operation is surely can be classified as `upsert` because it can update existing node attributes and also insert new attributes. But if in the point of node, the operation can only be classified as `update` because it is to update node document.

brianparry:
@yyscamper @zhangy42 The problem is that this definition of upsert is inconsistent with how the term is typically used.  Mongo defines upsert as "update or insert" a document.  Postgres is similar but substitutes table rows for documents.  Modifying or adding a property to an existing document is still just an update.

lib/jobs/upsert-node-job.js
```diff
@@ -0,0 +1,41 @@
...
+    function UpsertNodeAttribute(options, context, taskId) {
+        UpsertNodeAttribute.super_.call(this, logger, options, context, taskId);
+        this.nodeId = this.options.target || context.target;
+        this.data = this.options.data;
+    }
+    util.inherits(UpsertNodeAttribute, BaseJob);
+    
+    UpsertNodeAttribute.prototype._run = function run() {
+        var self = this;
+        return waterline.nodes.updateByIdentifier(self.nodeId, self.data)
```
zyoung51: This updates the node data without validating against the model.  Allowing schema violations will create failures in schema based backends (PostgreSQL) or may create unintended schema violations when the data is loaded back out of Mongo (ObjectID vs. string, etc.)

zhangy42:
test this please

lib/jobs/upsert-node-job.js
zhangy42: This script is more of a helper tool, we assume that the user of this tool will take the responsibility of all the possible consequences of the update/insert/upsert. 

The original story is here: 
https://www.pivotaltracker.com/story/show/121049511

zhangy42:
@yyscamper : You are right regarding the name. But since multiple people are confused with the name, I am changing the name anyway. Thank you @brianparry @yyscamper for pointing out!

zhangy42:
Per Joel's suggestion for letting jenkins work, I will close and reopen the PR.

lib/jobs/upsert-node-job.js
zyoung51: That seems like a big ask of the user to take on the responsibility of ensuring our database model stays consistent.  

We are going through a huge effort to add schema validation around tasks.  Providing an unvalidated entry point to modifying the nodes collection seems counter to that ongoing effort.  

We just removed this type of raw patching methodology in favor of the OBM API in 2.0, so I'm not sure why we are regressing back to this methodology.  The use case linked in PT should be resolved with a standard API entry point (ie: OBM APIs) and the poller creation API.  Assuming you don't want the user to use the OBM API (and/or the REST API task to do it), maybe you want something like this https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/create-ipmi-obm-settings.js that is specific (and can be validated with a schema) to what you referenced in PT and includes methods to validate the user supplied settings function before committing them to the database.

@yyscamper @RackHD/corecommitters thoughts on this?

panpan0000:
test this please

zhangy42:
Test this please

changev:
test this please

zhangy42:
@zyoung51 This feature is proposed by @pscharla and @benbp, and original ticket is at  https://www.pivotaltracker.com/story/show/121049511

Please ask either Paul or Ben for why they want the feature the way it currently is.

benbp:
@yyscamper I think part of the issue is still there are still, by design, some properties of the node document that are intended to be user-modified, such as tags, properties, and name, and it can be really useful to update these within a workflow (specifically in more interesting orchestration use cases). Ideally we'd just use the API, but we don't have a guarantee the workflow engine can access the API (and probably don't want to break that boundary anyways). Perhaps we can restrict the upsert operation to certain specific properties only, so that we have both protection from modifying internal system attributes, and flexibility for users wanting to do interesting things with workflows?

yyscamper:
@benbp : I think this for a while, there is sure some benefit to define a task to update node document, for example, user may want to auto-generate the node name or tag during the discovery.

I think it's good to restrict the upsert operation only to certain specific properties.

For obms. Actually, we have already a job for obm creation (https://github.com/RackHD/on-tasks/blob/master/lib/jobs/create-obm-settings.js), so I think it's better to move the obm from this job.

For sku, this is the same case just like obms.

For tags. We still have defined a job for tag generation (https://github.com/RackHD/on-tasks/blob/master/lib/jobs/generate-tag.js), of course that's depends on the tag rules are created beforehand.  Even if to support add tags, we should use the special API [waterline.nodes.addTags](https://github.com/RackHD/on-core/blob/2dde0f5eed7cf98bdec92edc99ffde45f5320e34/lib/models/node.js#L120) rather than the general `waterline.nodes.updateByIdentifier`.

For relations. We don't have a general job for it, but we do have some jobs for some particular relation, such as [enclosure-relation](https://github.com/RackHD/on-tasks/blob/master/lib/jobs/generate-enclosure.js), [pdu-relation](https://github.com/RackHD/on-tasks/blob/master/lib/jobs/pdu-node-relations.js) and [switch-relation](https://github.com/RackHD/on-tasks/blob/master/lib/jobs/switch-node-relations.js).

For sshSetting, I know the ssh-validate job has taken the action to update sshSetting into node document, see https://github.com/RackHD/on-tasks/blob/master/lib/jobs/validate-ssh.js.

To filter obms/tags/relations/sshSetting out, I think this job only need handle some very simple properties, such as `name`, `identifier`, `bootSetting`.

heckj:
I think you did some rebasing that wasn't entirely clean... at least by the list of commits in this PR. There's a bunch of merge commits that can definitely be collapsed out - I suspect you need to take another whack at the rebase

PengTian0:
test this please

zhangy42:
Let me first close this PR and reopen another one.

PR #315, opened 2016-08-15
Issue#155: Use ipmitool rather than /usr/bin/ipmitool

yyscamper:
Related issue: https://github.com/RackHD/RackHD/issues/155

In some system, the ipmitool may not always be installed in /usr/bin, the fix is to check the existence of some required commend line tools before on-tasks starts, so that user can be aware of such issue at very early stage.

For the new change, RackHD will use `ipmitool` rather than `usr/bin/ipmitool`, this let user to make sure adding tool in the $PATH.

At present, I only add `ipmitool` as required tools, not including the `snmp` because I see most people doesn't have the chance to try the snmp related functionality. If you have concern, It's OK for me to include `snmp` as well.

@RackHD/corecommitters @heckj @codenrhoden @jgolio

### DIFF
--- a/lib/utils/job-utils/ipmitool.js
+++ b/lib/utils/job-utils/ipmitool.js
@@ -25,48 +25,40 @@ function ipmitoolFactory(Promise) {
     */
     Ipmitool.prototype.runCommand = function(host, user, password, command) {
         return new Promise(function (resolve, reject) {
-            fs.exists('/usr/bin/ipmitool', function(exists) {
-                if (!exists) {
-                    reject("ipmitool isn't hosted on the local machine" +
-                                        " at /usr/bin/ipmitool");
-                    return;
-                }
-
-                var options = { timeout: 60000 };
-                if (host && user && password && command) {
-                    child_process.exec( // jshint ignore:line
-                            '/usr/bin/ipmitool -I lanplus -U '+user+
-                                ' -H '+host+' -P '+password+" " + command,
-                            options,
-                            function(error, stdout, stderr) {
-                                if (error) {
-                                    error.stderr = stderr.replace('-P '+password, '');
-                                    reject(error);
-                                } else {
-                                    resolve(stdout);
-                                }
-                    });
+            var options = { timeout: 60000 };
+            if (host && user && password && command) {
+                child_process.exec( // jshint ignore:line
+                        'ipmitool -I lanplus -U '+user+
+                            ' -H '+host+' -P '+password+" " + command,
+                        options,
+                        function(error, stdout, stderr) {
+                            if (error) {
+                                error.stderr = stderr.replace('-P '+password, '');
+                                reject(error);
+                            } else {
+                                resolve(stdout);
+                            }
+                });
+            } else {
+                if (!host && command) {
+                    child_process.exec('ipmitool -I lanplus ' + command, // jshint ignore:line
+                        options,
+                        function(error, stdout, stderr) {
+                            if (error) {
+                                error.stderr = stderr;
+                                reject(error);
+                            } else {
+                                resolve(stdout);
+                            }
+                        });
+                } else if (!user) {
+                    reject("user not defined");
+                } else if (!password) {
+                    reject("password not defined");
                 } else {
-                    if (!host && command) {
-                        child_process.exec('/usr/bin/ipmitool -I lanplus ' + command, // jshint ignore:line
-                            options,
-                            function(error, stdout, stderr) {
-                                if (error) {
-                                    error.stderr = stderr;
-                                    reject(error);
-                                } else {
-                                    resolve(stdout);
-                                }
-                            });
-                    } else if (!user) {
-                        reject("user not defined");
-                    } else if (!password) {
-                        reject("password not defined");
-                    } else {
-                        reject("command not defined");
-                    }
+                    reject("command not defined");
                 }
-            });
+            }
         });
     };
 


lib/task-service.js
```diff
@@ -4,21 +4,54 @@
...
+) {
+    var logger = Logger.initialize(taskServiceFactory);
+    var requiredCmdTools = ['ipmitool'];
+
     function TaskService() {}
 
     TaskService.prototype.start = function() {
-        return validator.register();
+        var self = this;
+        return Promise.resolve()
```
zyoung51: Unnecessary, checkToolsExist already returns a promise.

lib/task-service.js
```diff
@@ -4,21 +4,54 @@
...
+        .then(function() {
+            return self.checkToolsExist();
+        })
+        .then(function() {
```
zyoung51: Should this be a tap?  Do you want to return the value that came out of checkToolsExist or the value coming out of the validator register?

lib/task-service.js
```diff
@@ -4,21 +4,54 @@
...
 
-function factory(validator) {
+function taskServiceFactory(
+    Logger,
+    Promise,
+    ChildProcess,
+    validator
+) {
+    var logger = Logger.initialize(taskServiceFactory);
+    var requiredCmdTools = ['ipmitool'];
```
zyoung51: Will this eventually include other tools like snmputil?  Should this be a constant?

lib/task-service.js
yyscamper: Actually, that's designed purposely. As I see some function returns Promise, such as `validator.register` and `checkToolsExist`, but some function doesn't return Promise, such as `validator.reset`. If I directly chain promise after the function, I need to take special care for the return value of each promise. But if I put these functions into nesting promise,s I don't need to care their return values.

Actually  this doesn't bring too much benefit since only a few sub-functions in `start`/`stop` at present. It brings flexibility only if more and more sub-functions are put here in future.

So I'm OK to change this or not. If you insist on the Promise.resolve() is redundant, I'm OK to remove it.

lib/task-service.js
yyscamper: Actually I don't want to return nothing more to the caller except the fulfilled/rejected state. So changeingall `then` to `tap` should be more suitable here.

lib/task-service.js
yyscamper: As I said in the PR comment, I purposely only adding the ipmitool and I see the snmp and racadm  tool aren't used too often, especially the racadm tool is only used for Dell server and a few tasks, it's even not easy to install it. So if preventing user to use RackHD if not installing some not required tool (for user's special user case) doesn't make sense.

But I'm OK to include more if you think that's necessary.

As for why not define it as constant, since I thought the `requiredCmdTools` vary with different customers, so maybe it's better to be a configuration item or even get some input from skupack (like the racadm tool for Dell Servers).

Do you have any good idea to handle this?

lib/task-service.js
brianparry: @yyscamper Bluebird's `Promise.try()` is better to this purpose than `Promise.resolve()`

lib/task-service.js
brianparry: @yyscamper I think this check is logically misplaced in `taskService.start`.  Tasks and workflows are dependent on tools, not the service itself.  Rather than preventing the service from starting if tools are missing, I think it would be better to check for existence of all tools needed to run a graph prior to starting it.

yyscamper: @brianparry : good idea. So seems it's better to annotate some required tools for each job or task or graph, when that graph or task is triggered, we can first scan the existence of those tools and fail immediately if any tool is missing, does this match your thinking?

lib/task-service.js
yyscamper: Thx

panpan0000:
I'm JenkinsRHD :-)
Jenkins build on-tasks/1424 failed.
because on-taskgraph #1308 Result was FAILURE
in  on-taskgraph #1308:

Checking out Revision cc4376ab8066bd20abebbb176a97e3260f046fbb (origin/pr/156/merge)
 > git checkout -f cc4376ab8066bd20abebbb176a97e3260f046fbb
FATAL: Could not checkout cc4376ab8066bd20abebbb176a97e3260f046fbb

codenrhoden:
No opinion on which tools should be required and which shouldn't. That's a tough one since it does very by usecase quite a bit.  `racadm` is hardly used, as is the `amt` tooling.

But this looks good to me. 👍   I like that it just uses normal $PATH resolution for `ipmitool` now. That's great.

yyscamper:
@brianparry @zyoung51 @codenrhoden : In my latest change, I have completely remove the checkToolExist and only change the `/usr/bin/ipmitool` to `ipmitool`.

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

panpan0000:
test this please

yyscamper:
test this please

yyscamper:
@panpan0000 was doing some Jenkins experiment on my PR, it was lucky for me ^^

cgx027:
Looks good to me. :+1:

changev:
test this please

yyscamper:
@brianparry  @zyoung51 : What's your comment on the latest change?

anhou:
:+1:

PR #317, opened 2016-08-22
Updated the sel poller functionality to comply with ORFS-250

tannoa2:
The new implementation:
1- will take action input from the user through the sku config.
In the skupack  config.json file, the alerts are defined as shown below:

image

2- Updated the ipmitool command being used (from "-c sel list" to "sel get <ID>")so user get additional information : 

The old sel info looked like this: 
{"logId": "68",
"date": "08/04/2016",
"time": "15:14:11",
"sensorType": "OS",
"sensorNumber": "Boot",
"event": "C: boot completed",
"value": "Asserted"}

The new sel info info looks: 
{ 'SEL Record ID': '0042',
'Record Type': '02',
Timestamp: '07/20/2016 12:37:38',
'Generator ID': '0020',
'EvM Revision': '04',
'Sensor Type': 'Power Supply (0x08)',
'Sensor Number': '6d',
'Event Type': 'Generic Discrete',
'Event Direction': 'Assertion Event',
'Event Data (RAW)': '01ffff',
Description: 'State Asserted',
'Sensor ID': 'PSU Alert (0x6d)',
'Entity ID': '208.1 (Unknown (0xD0))' },

3-Added a raw decoding capabilities of the SEL in order to capture addition information that is specified in the spec but not provided by ipmitool ("Event Type Code" and "Sensor Type Code").

4-The sel poller will report only the new new SEL events and ignore the existing onces

### DIFF
--- a/lib/jobs/ipmi-job.js
+++ b/lib/jobs/ipmi-job.js
@@ -184,19 +184,117 @@ function ipmiJobFactory(
     };
 
     /**
-     * Collect SEL entries list from IPMI
+     * gets the SEL information using the "sel get" ipmi command
      * @memberOf IpmiJob
      *
      * @param data
-     * @param count
+     * @param stringifiedEntries
      */
-    IpmiJob.prototype.collectIpmiSel = function(data, count) {
-        count = count || 25;
-        return ipmitool.sel(data.host, data.user, data.password, count)
-        .then(function (sel) {
-            return parser.parseSelData(sel);
-        });
+    IpmiJob.prototype.getSelEntries = function(data, stringifiedEntries) {
+        return ipmitool.sel(data.host, data.user, data.password,  stringifiedEntries)
     };
+
+    /**
+     * gets the raw information of a single sel entry
+     * @memberOf IpmiJob
+     *
+     * @param data
+     * @param recordId
+     */
+    IpmiJob.prototype.getRawSelEntry = function(data, recordId) {
+        return ipmitool.readRawSel(data.host, data.user, data.password, recordId)
+    };
+
+    /**
+     * Collect SEL entries list from IPMI and add two properties ("Sensor Type Code" and "Event Type Code") to each sel entry
+     * @memberOf IpmiJob
+     *
+     * @param data
+     */
+    IpmiJob.prototype.collectIpmiSel = function(data) {
+        var self = this;
+        var stringifiedEntries = "";
+        var lastUpdatedSelDate;
+        var currentSelDate;
+        var lastReportedSelEntryID;
+        var totalEntries;
+        var lastUpdatedSelDate;
+        var sel;
+
+        return Promise.all([
+           self.collectIpmiSelInformation(data),
+           waterline.workitems.findOne({ id: data.workItemId })
+        ])
+        .spread (function(selInfo, workObj){
+            totalEntries = parseInt(selInfo.Entries);
+            var tmpDate = selInfo["Last Add Time"].split(/\/|:| /);
+            tmpDate =tmpDate[0] +" "+ tmpDate[1] +", "+ tmpDate[2] +" "+ tmpDate[3] +":"+ tmpDate[4]+":"+ tmpDate[5] ;
+            currentSelDate = new Date(tmpDate);
+            lastReportedSelEntryID = _.get(workObj, 'config.lastReportedSelEntryID', 0);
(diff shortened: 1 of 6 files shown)


tannoa2:
@RackHD/corecommitters @zyoung51 @jlongever @uppalk1 @keedya @amymullins

lib/jobs/ipmi-job.js
```diff
@@ -184,20 +184,117 @@ function ipmiJobFactory(
...
+        var sel;
+
+        return Promise.all([
+           self.collectIpmiSelInformation(data),
+           waterline.workitems.findOne({ id: data.workItemId })
+        ])
+        .spread (function(selInfo, workObj){
+            totalEntries = parseInt(selInfo.Entries);
+            var tmpDate = selInfo["Last Add Time"].split(/\/|:| /);
+            tmpDate =tmpDate[0] +" "+ tmpDate[1] +", "+ tmpDate[2] +" "+ tmpDate[3] +":"+ tmpDate[4]+":"+ tmpDate[5] ;
```
benbp: If you're interested, you can probably write this a little easier like:

```
tmpDate = "%s %s, %s %s:%s:%s".format.apply(null, selInfo["Last Add Time"].split(/\/|:| /));
```

Because we've added `format` to the String prototype here: https://github.com/RackHD/on-core/blob/master/lib/extensions.js#L39

lib/jobs/ipmi-job.js
```diff
@@ -184,20 +184,117 @@ function ipmiJobFactory(
...
+            var decodedSel = {
+                "Event Type Code" : "",
+                "Sensor Type Code" : ""
+            };
+            rawArr.forEach(function (value, index) {
+                var values = [];
+                var rawSelRecord;
+                var mask = 1 << 7 //a mask to clear the 7th bit
+
+                rawArr[index] = value.replace(new RegExp('\r?\n','g'), '');
```
benbp: If you define this RegExp as it's own variable outside the for loop (e.g. `var regex = new RegExp(...)`), and pass it in like `.replace(regex, '')`, then we'll only compile the regex once rather than for every iteration. Probably an inconsequential performance improvement though.

lib/jobs/ipmi-job.js
```diff
@@ -184,20 +184,117 @@ function ipmiJobFactory(
...
+        var stringifiedEntries = "";
+        var lastUpdatedSelDate;
+        var currentSelDate;
+        var lastReportedSelEntryID;
+        var totalEntries;
+        var lastUpdatedSelDate;
+        var sel;
+
+        return Promise.all([
+           self.collectIpmiSelInformation(data),
```
benbp: I could be reading this wrong, but it seems like we are replacing the sel list call with a max count of 25 with a call to get the entirety of the list, and then adding an additional call to sel get. Part of the reason we had the max count before was because we had load problems on both our system and the BMC retrieving the whole sel log. Do you anticipate any scale issues with reverting to what I think could be more data and an additional ipmi call per poll?

tannoa2: HI @benbp,
It doesn't get the entire SEL. It gets the range starting from the last SEL record that was read (lastReportedSelEntryID, and it is saved in the database) into the ed of the SEL.  
I tested this new implementation by reading a filled-up SEL log and it works.
I ran into an issue where it was exceeding the max buffer size of the process , but increased its size in the ipmitool file (the default 200K to 400K):
https://github.com/tannoa2/on-tasks/blob/bca88dff4b0499ea89566e070f0b2715c394cc1a/lib/utils/job-utils/ipmitool.js#L35

But Again there is will be a rare case of reading the whole SEL, we are only reading the newly reported entries as specifed by the spec :
"OnRack shall ignore all existing SEL events and only interrogate new SEL events from the time the IPMI poller starts. Each event shall only be reported on once"

benbp: I saw that the `sel get` call was querying a limited range of entries, but my reading of this code was that the initial call to `sel` from `collectIpmiSelInformation` was not limited to any range, but I could be totally reading the code wrong :)

tannoa2: For the initial reading of the SEL we read all of the entries, correct.

benbp: Ahh ok, sorry I missed the changes to `Ipmitool.prototype.sel`, thanks!

benbp:
👍

keedya:
:+1:

PR #320, opened 2016-08-24
Update generate-tag task to take an array of node IDs

benbp:
EDIT: requires https://github.com/RackHD/on-taskgraph/pull/162 and https://github.com/RackHD/on-http/pull/442

This is an update to the generate-tag task and job to support processing an array of nodeIds within a single workflow instance, rather than tying one workflow instance to one nodeId. The upstream impact of this is that issuing POST calls to the `/api/<version>/tags` API no longer triggers N workflows at once where N == the number of nodes in the database, but rather a single workflow (see ODR-821 for internal reporting).

This is partially a quick and easy patch that glosses over issues with the workflow engine itself (system slowdowns when batch running hundreds of workflows at the same time with a small number of task runner/scheduler instances), and while this issue happens to expose this problem, I don't think that the fix should be tied to it, as the solution to our performance problems is likely going to warrant a much more significant investigation and/or overhaul (and I welcome all the shaming @zyoung51 and @jlongever can throw at me 😄).

@yyscamper I've added a JSON schema for the generate-tag task, which is I believe the first schema that uses the oneOf field for top level properties. As such, I had to make some minor modifications to the schema-ut-helper test library. Please let me know if I'm doing things in a different way than you intended or if there is a better way to do them (alternatively, we could just deprecate the singular nodeId option altogether and not need the oneOf option, but I didn't want to potentially break any downstream users of the existing workflow outside of our internal API logic).

@RackHD/corecommitters @pscharla @heckj

### DIFF
--- a/lib/jobs/generate-tag.js
+++ b/lib/jobs/generate-tag.js
@@ -41,8 +41,8 @@ function generateTagJobFactory(
     function GenerateTagJob(options, context, taskId) {
         GenerateTagJob.super_.call(this, logger, options, context, taskId);
 
-        this.nodeId = context.target || options.nodeId;
-        assert.isMongoId(this.nodeId, 'context.target || options.nodeId');
+        this.nodeIds = options.nodeIds;
+        assert.arrayOfString(this.nodeIds, 'options.nodeIds');
     }
 
     util.inherits(GenerateTagJob, BaseJob);
@@ -66,35 +66,42 @@ function generateTagJobFactory(
                 })
                 .uniq()
                 .value();
-            return [ Promise.map(catalogTypes, function(type) {
-                return waterline.catalogs.findMostRecent({
-                    node: self.nodeId,
-                    source: type
-                });
-            }), tags ];
-        }).spread(function(catalogs, tags) {
-            catalogs = _(catalogs)
-                .flattenDeep()
-                .compact()
-                .transform(function (catalogs, catalog) {
-                    catalogs[catalog.source] = catalog.data;
-                }, {})
-                .value();
 
-            var matches = _.filter( matchTags(catalogs, tags), function(match) {
-                return match.tag.rules.length && !match.errors.length;
+            return Promise.map(self.nodeIds, function(nodeId) {
+                return Promise.map(catalogTypes, function(type) {
+                    return waterline.catalogs.findMostRecent({
+                        node: nodeId,
+                        source: type
+                    });
+                }).then(processCatalogs.bind(null, tags, nodeId));
             });
-
-            var tagNames = _(matches).pluck('tag.name').value();
-            if( !_.isEmpty(tagNames)) {
-                return waterline.nodes.addTags(self.nodeId, tagNames);
-            }
         }).then(function() {
             self._done();
         }).catch(function(err) {
             self._done(err);
         });
     };
+
+    function processCatalogs(tags, nodeId, catalogs) {
+        catalogs = _(catalogs)
+            .flattenDeep()
+            .compact()
+            .transform(function (catalogs, catalog) {
+                catalogs[catalog.source] = catalog.data;
+            }, {})
+            .value();
+
+        var matches = _.filter( matchTags(catalogs, tags), function(match) {
+            return match.tag.rules.length && !match.errors.length;
+        });
+
+        var tagNames = _(matches).pluck('tag.name').value();
+
+        if (!_.isEmpty(tagNames)) {
+            return waterline.nodes.addTags(nodeId, tagNames);
+        }
+        return Promise.resolve();
+    }
 
     function matchTags(catalogs, tags) {
         return _.map(tags, function (tag) {
--- a/lib/task-data/base-tasks/generate-tag.js
+++ b/lib/task-data/base-tasks/generate-tag.js
@@ -7,7 +7,7 @@ module.exports = {
     injectableName: 'Task.Base.Catalog.GenerateTag',
     runJob: 'Job.Catalog.GenerateTag',
     requiredOptions: [
-        'nodeId'
+        "nodeIds"
     ],
     requiredProperties: {},
     properties: {}
--- a/lib/task-data/schemas/generate-tag.json
+++ b/lib/task-data/schemas/generate-tag.json
@@ -1,0 +1,22 @@
+{
+    "$schema": "rackhd-task-schema.json",
+    "copyright": "Copyright 2016, EMC, Inc.",
+    "title": "Generate Tag Task",
+    "description": "The schema for generate-tag",
+    "describeJob": "Job.Catalog.GenerateTag",
+    "allOf": [
+        { "$ref": "common-task-options.json#/definitions/Options" },
+        {
+            "properties": {
+                "nodeIds": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    },
+                    "description": "An array of nodeIds to process tags against"
+                }
+            },
+            "required": ["nodeIds"]
+        }
+    ]
+}
--- a/lib/task-data/tasks/generate-tag.js
+++ b/lib/task-data/tasks/generate-tag.js
@@ -6,8 +6,9 @@ module.exports = {
     friendlyName: 'Generate Tag',
     injectableName: 'Task.Catalog.GenerateTag',
     implementsTask: 'Task.Base.Catalog.GenerateTag',
+    schemaRef: 'generate-tag.json',
     options: {
-        nodeId: null
+        nodeIds: []
     },
     properties: {}
 };
--- a/spec/lib/task-data/schemas/generate-tag-spec.js
+++ b/spec/lib/task-data/schemas/generate-tag-spec.js
@@ -1,0 +1,33 @@
+// Copyright 2016, EMC, Inc.
+/* jshint node: true */
+
+'use strict';
+
+describe(require('path').basename(__filename), function() {
+    var schemaFileName = 'generate-tag.json';
+
+    var canonical = {
+        "nodeIds": []
+    };
+
+    var negativeSetParam = {
+        "nodeIds": [ "test", [ "test", 1234 ] ]
+    };
+
+    var positiveSetParam = {
+        "nodeIds": [
+            [ ],
+            [ "579b7de7a39b92ee0da4f26a" ],
+            [ "579b7de7a39b92ee0da4f26a", "579b7e47dc1c6c180e39d458" ],
+            [ "579b7de7a39b92ee0da4f26a", "579b7e47dc1c6c180e39d458", "579b7de7a39b92ee0da4f26a" ]
+        ]
+    };
+
+    var negativeUnsetParam = [ "nodeIds" ];
+
+    var positiveUnsetParam = [];
+
+    var SchemaUtHelper = require('./schema-ut-helper');
+    new SchemaUtHelper(schemaFileName, canonical, null, null).batchTest(
+        positiveSetParam, negativeSetParam, positiveUnsetParam, negativeUnsetParam);
+});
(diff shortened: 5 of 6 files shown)


lib/jobs/generate-tag.js
```diff
@@ -41,8 +41,13 @@ function generateTagJobFactory(
...
         GenerateTagJob.super_.call(this, logger, options, context, taskId);
 
-        this.nodeId = context.target || options.nodeId;
-        assert.isMongoId(this.nodeId, 'context.target || options.nodeId');
+        if (options.nodes) {
+            this.nodes = options.nodes;
+            assert.arrayOfString(this.nodes, 'options.nodes');
+        } else {
+            this.nodeId = context.target || options.nodeId;
+            assert.isMongoId(this.nodeId, 'context.target || options.nodeId');
```
heckj: I recognize we're changing this, but should this be pinned to a mongoId or is UUID now more appropriate with the migration of node ID's to UUIDs? (or am I a clueless monkey here?)

heckj:
question about the assert being used (which may be a stupid one), but otherwise 👍

lib/jobs/generate-tag.js
benbp: Whoops, no this warrants a change (these have been mostly updated piecemeal rather than all at once). Good catch.

benbp: Changed it to just `assert.string`.

lib/jobs/generate-tag.js
```diff
@@ -66,36 +71,53 @@ function generateTagJobFactory(
...
+        catalogs = _(catalogs)
+            .flattenDeep()
+            .compact()
+            .transform(function (catalogs, catalog) {
+                catalogs[catalog.source] = catalog.data;
+            }, {})
+            .value();
+
+        var matches = _.filter( matchTags(catalogs, tags), function(match) {
+            return match.tag.rules.length && !match.errors.length;
```
anhou: don't return promise, is it right or intended?

lib/task-data/schemas/generate-tag.json
```diff
@@ -0,0 +1,41 @@
...
+                            "type": "string",
+                            "description": "An individual nodeId to process tags against"
+                        },
+                        "nodes": {
+                            "type": "null"
+                        }
+                    },
+                    "required": ["nodeId"]
+                },
+                {
```
anhou: How about only use arrays of nodes, don't need support nodeId option. I think for usage perspective, the array of nodes could include one nodeId when array of nodes length is 1, and only contain one nodeId. it will be not complex from usage or codes implementation.

panpan0000:
test this please

lib/jobs/generate-tag.js
yyscamper: this is not the out-most return, just within the callback of `filter`, so it doesn't need be promise.

lib/jobs/generate-tag.js
```diff
@@ -41,8 +41,13 @@ function generateTagJobFactory(
     function GenerateTagJob(options, context, taskId) {
         GenerateTagJob.super_.call(this, logger, options, context, taskId);
 
-        this.nodeId = context.target || options.nodeId;
-        assert.isMongoId(this.nodeId, 'context.target || options.nodeId');
+        if (options.nodes) {
+            this.nodes = options.nodes;
+            assert.arrayOfString(this.nodes, 'options.nodes');
+        } else {
+            this.nodeId = context.target || options.nodeId;
```
yyscamper: Actually the logic can be combined, so that the `_run` only handles the multiple nodes case. Even in future, we want to get rid of the singular nodeId, it will be easier:
```javascript
if (options.nodes) {
    this.nodes = options.nodes;
} else {
    this.nodes = [context.target || options.nodeId];
} 
assert.arrayOfString(this.nodes, 'Nodes from options');
``

File: lib/task-data/schemas/generate-tag.json [hunk 5]
```diff
@@ -0,0 +1,41 @@
+{
+    "$schema": "rackhd-task-schema.json",
+    "copyright": "Copyright 2016, EMC, Inc.",
+    "title": "Generate Tag Task",
+    "description": "The schema for generate-tag",
+    "describeJob": "Job.Catalog.GenerateTag",
+    "allOf": [
+        { "$ref": "common-task-options.json#/definitions/Options" },
+        {
+            "oneOf": [
```
yyscamper: The sub-schemas within `oneOf` doesn't need to be so complex, unlike `anyOf`, `oneOf` is to assert it conforms to ONLY one sub-schema, it will fail if conforms to two or more sub-schemas. So it can be simplified to:
```json
"allof": [
   { "$ref": "common-task-options.json#/definitions/Options" },
   {
        "properties": {
            "nodeId": {
                "type": "string",
                "description": "An individual nodeId to process tags against"
            },
            "nodes": {
                "type": "array",
                "description": "An array of nodeIds to process tags against",
                "items": {
                    "type": "string"
                }
            }
        },
        "oneOf": [
            { "required": ["nodeId"] },
            { "required": ["nodes"] }
        ]
   }
]
```

lib/task-data/tasks/generate-tag.js
```diff
@@ -6,8 +6,10 @@ module.exports = {
     friendlyName: 'Generate Tag',
     injectableName: 'Task.Catalog.GenerateTag',
     implementsTask: 'Task.Base.Catalog.GenerateTag',
+    schemaRef: 'generate-tag.json',
     options: {
-        nodeId: null
+        nodeId: null,
+        nodes: null
```
yyscamper: If you agree to change to my proposed schemas, then both `null` options need be removed here as the `null` violates the schema.

lib/task-data/schemas/generate-tag.json
yyscamper: There is similar example, that is [create-megaraid.json](https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/create-megaraid.json#L80), the difference is it uses `anyOf` for `required`, but here should be `oneOf`

spec/lib/task-data/schemas/schema-ut-helper.js
```diff
@@ -136,19 +136,20 @@ function validateSchemaDefinition(schema, jobNames) {
  * validation, this depends the the canonical data is specified
  */
 function SchemaUnitTestHelper(schemaFileName, canonicalData, skipTaskSchemaDefValidation,
-                              skipCommonOptionsValidation) {
+                              skipCommonOptionsValidation, supportNullOptions) {
```
yyscamper: if you agree to my proposed schema change, then this change isn't necessary.

yyscamper:
@benbp : I see you change the schema-ut-helper.js but I don't see your unit-test file for generate-tag.json schema, I guess  you forgot to commit it.

lib/jobs/generate-tag.js
```diff
@@ -66,36 +71,53 @@ function generateTagJobFactory(
...
+                    return Promise.map(catalogTypes, function(type) {
+                        return waterline.catalogs.findMostRecent({
+                            node: nodeId,
+                            source: type
+                        });
+                    }).then(processCatalogs.bind(null, tags, nodeId));
+                });
+            }
+
+            return Promise.map(catalogTypes, function(type) {
```
yyscamper: If you agree to assign the singular nodeId into self.nodes, then we can completely remove this code block.

yyscamper:
It's happy there is new people to add new task schema, @benbp , you are the third, @cgx027 is the second. ^^

lib/jobs/generate-tag.js
anhou: you're right :)

lib/task-data/schemas/generate-tag.json
benbp: Oh awesome, thanks this is what I was looking for!

benbp:
@yyscamper woohoo third place! Yeah, so I agree that it would be simpler to just support a `nodes` option, but I was worried about whether anyone was manually invoking the graph today (probably unlikely, but you never know). I also was thinking of some use cases where someone might have a graph running against a node, or with a shared `nodeId` option across tasks, that they wanted to include tag creation within instead of using the API. In those cases, it could be useful to have the nodeId option. That said, maybe you're right and it's not worth it, so I'm happy to change it.

lib/jobs/generate-tag.js
benbp: Yeah, I only named it nodes because I was worried about potential spelling mistakes since nodeId and nodeIds are so similar. But if we switch it to not include a `nodeId` option as you suggest, I might call the option `nodeIds` only.

benbp:
Updated to just use `options.nodeIds`, which also means I can remove the updates to the schema-ut-helper since I'm not using `oneOf` anymore.

lib/task-data/tasks/generate-tag.js
```diff
@@ -6,8 +6,9 @@ module.exports = {
     friendlyName: 'Generate Tag',
     injectableName: 'Task.Catalog.GenerateTag',
     implementsTask: 'Task.Base.Catalog.GenerateTag',
+    schemaRef: 'generate-tag.json',
     options: {
-        nodeId: null
+        nodeIds: null
```
yyscamper: This will fail the unit-testing as the `null` violates the schema.

lib/task-data/base-tasks/generate-tag.js
```diff
@@ -6,9 +6,7 @@ module.exports = {
     friendlyName: 'Generate Tag',
     injectableName: 'Task.Base.Catalog.GenerateTag',
     runJob: 'Job.Catalog.GenerateTag',
-    requiredOptions: [
-        'nodeId'
-    ],
+    requiredOptions: [],
```
yyscamper: Add `nodeIds` as requiredOpitons.

changev:
jenkins: depends on https://github.com/RackHD/on-taskgraph/pull/162 https://github.com/RackHD/on-http/pull/442

anhou:
👍  with test pass

spec/lib/task-data/schemas/generate-tag-spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+describe(require('path').basename(__filename), function() {
+    var schemaFileName = 'generate-tag.json';
+
+    var canonical = {
+        "nodeIds": []
+    };
+
+    var negativeSetParam = {
+        "nodeIds": [ "test", 1234 ]
```
yyscamper: There is a little tricky here, you need use the Array's Array here, as the testing will create a test case for each element in the array, rather than use the array as a whole. See the code: https://github.com/RackHD/on-tasks/blob/master/spec/lib/task-data/schemas/schema-ut-helper.js#L281,
```
"nodeIds": [  ["test", 123] ]
```

brianparry:
Looks good to me :+1:

lib/task-data/base-tasks/generate-tag.js
benbp: Oh okay, I had this as non-required so I could have `nodeIds` as null in the definition, while still letting the JSON schema work, but maybe I am misunderstanding the right way to do this.

spec/lib/task-data/schemas/generate-tag-spec.js
benbp: Will do.

benbp:
test this please

yyscamper:
:+1:

changev:
test this please 

for dependency

pscharla:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

pscharla:
@RackHD/corecommitters I got confirmation from @panpan0000 that we can ignore the Jenkins Dependency PR check for now.

PR #322, opened 2016-08-25
This is a fix that applies to issue 337 and required for Fix Bootstrap payload for esxi fails when IPv4 is set as static

keedya:
Fix Bootstrap payload for esxi fails when IPv4 is set as static

Resolves https://github.com/RackHD/RackHD/issues/337:
- esxSwitchName is missing from networkDevices schema

### DIFF
--- a/lib/task-data/schemas/install-os-types.json
+++ b/lib/task-data/schemas/install-os-types.json
@@ -162,7 +162,7 @@
                 }
             },
             "required": ["device"],
-            "additionalProperties": false
+            "additionalProperties": true
         },
         "NetworkDevices": {
             "description": "The network configuration for each NIC",


keedya:
@RackHD/corecommitters @brianparry @zyoung51 @jlongever @yyscamper @srinia6 @amymullins @BillyAbildgaard @dalebremner

srinia6:
+1

benbp:
Out of curiousity, how does restricting additional properties fix an issue? Is it because passing in other magic properties trigger other functionality we don't want?

👍

keedya:
@benbp this fix actually does the opposite. Allow passing other properties for net configure. In this case I needed to pass esxiswitchName

benbp:
Oh, silly me, you changed it to true. Thanks!

keedya:
@benbp this fix actually does the opposite. Allow passing other properties for net configure. In this case I needed to pass esxiswitchName

PR #325, opened 2016-08-30
SNMP poller alert

zyoung51:
- Add alerts for metrics
- Add sensor metric poller
- Change alerts to support regex based keyname matches and validation
  object matches for values
- Modify generate-sku to treat absent catalogs as do-not-care instead
  of an error


panpan0000:
test this please

amymullins:
test this please

PR #323, opened 2016-08-26
Revert "Updated the sel poller functionality to comply with ORFS-250"

benbp:
Reverts RackHD/on-tasks#317

Apparently the sel get command introduced by this PR doesn't seem to work on intel nodes, so we might want a revert for now until we figure out a solution.

@RackHD/corecommitters @gpaulos @pscharla @tannoa2

PR #326, opened 2016-08-30
[ORFS-250] Add SEL entry alerting

jlongever:
@RackHD/corecommitters @tannoa2 @uppalk1 @zyoung51 @keedya 

This reverts the revert PR https://github.com/RackHD/on-tasks/pull/323 to add SEL entry alerting support added by PR https://github.com/RackHD/on-tasks/pull/317.  

Change made to define a new work-item poller that collects each new SEL entry (`sel get n..n`) as opposed to replacing the existing SEL list (`sel list -c 25`) poller.

### DIFF
--- a/lib/jobs/ipmi-job.js
+++ b/lib/jobs/ipmi-job.js
@@ -237,14 +237,23 @@ function ipmiJobFactory(
         var currentSelDate;
         var lastReportedSelEntryID;
         var totalEntries;
-        var sel;
 
         return Promise.all([
            self.collectIpmiSelInformation(data),
            waterline.workitems.findOne({ id: data.workItemId })
-        ])
-        .spread (function(selInfo, workObj){
+        ]).spread (function(selInfo, workObj){
             totalEntries = parseInt(selInfo.Entries);
+            /*  Parses SEL info output:     
+                "SEL Information
+                 Version          : 1.5 (v1.5, v2 compliant)
+                 Entries          : 187
+                 Free Space       : 13008 bytes
+                 Percent Used     : 18%
+                 Last Add Time    : 01/01/1970 01:56:22
+                 Last Del Time    : Not Available
+                 Overflow         : false
+                 Supported Cmds   : 'Delete' 'Reserve'"
+            */
             var tmpDate = selInfo["Last Add Time"].split(/\/|:| /);
             tmpDate = tmpDate[0] +" "+ tmpDate[1] +", "+ tmpDate[2] +
                 " "+ tmpDate[3] +":"+ tmpDate[4]+":"+ tmpDate[5];
@@ -273,33 +282,28 @@ function ipmiJobFactory(
             return parser.parseSelDataEntries(unparsedSel);
         })
         .then(function(parsedSel) {
+            // we get the raw sel info for each sel entry in order to decode additional info
             var rawArr = [];
-            var rawData;
-            // we get the raw sel info for each sel entry in order to decode additional informaiton
-            sel = parsedSel;
-            if(sel) {
-                _.forEach(parsedSel, function(entry) {
-                    rawData = self.getRawSelEntry(data, entry["SEL Record ID"])
-                    .then(function(data){
-                        return data;
-                    });
+            return Promise.each(parsedSel, function(entry) {
+                return self.getRawSelEntry(data, entry["SEL Record ID"])
+                .then(function(rawData) {
                     rawArr.push(rawData);
                 });
-            }
-            return Promise.all(rawArr);
+            }).then(function() {
+                return [ parsedSel, rawArr ];
+            });
         })
-        .then(function(rawArr) {
+        .spread(function(parsedSel, rawArr) {
             // we add two properties ("Sensor Type Code" and "Event Type Code") to each sel entry.
             var decodedSel = {
                 "Event Type Code" : "",
                 "Sensor Type Code" : ""
             };
             var regex = new RegExp('\r?\n','g');
-            rawArr.forEach(function (value, index) {
+            _.forEach(rawArr, function(value, index) {
                 var values = [];
                 var rawSelRecord;
                 var mask = 1 << 7; // a mask to clear the 7th bit
-
                 rawArr[index] = value.replace(regex, '');
                 values = rawArr[index].split(" ");
                 rawSelRecord = values.slice(3,19);
@@ -308,9 +312,9 @@ function ipmiJobFactory(
                 decodedSel["Event Type Code"] = 
                     ("0" +(parseInt(rawSelRecord[12],16) & ~mask)).toString(16).slice(-2);
                 decodedSel["Sensor Type Code"] = rawSelRecord[10];
-                sel[index] = _.merge(sel[index], decodedSel);
+                parsedSel[index] = _.merge(parsedSel[index], decodedSel);
             });
-            return sel;
+            return parsedSel;
         });
     };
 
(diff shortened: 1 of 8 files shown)


lib/utils/job-utils/ipmitool.js
```diff
@@ -163,16 +162,44 @@ function ipmitoolFactory(Promise) {
...
+     * Returns a promise with the results or errors of invoking sel raw
+     *
+     * @param host
+     * @param user
+     * @param password
+     * @param Sel Entries
+     */
+    Ipmitool.prototype.readRawSel = function(host, user, password, selEntry) {
+        var tmp = selEntry.split("") ;
+        var command = "raw 0x0a 0x43 0x00 0x00 " +
```
yyscamper: Please add some comment for these constants, otherwise other developer will not be aware of what "0x0a" and "0x43" means.

lib/jobs/ipmi-job.js
```diff
@@ -199,6 +224,98 @@ function ipmiJobFactory(
...
+        var totalEntries;
+        var sel;
+
+        return Promise.all([
+           self.collectIpmiSelInformation(data),
+           waterline.workitems.findOne({ id: data.workItemId })
+        ])
+        .spread (function(selInfo, workObj){
+            totalEntries = parseInt(selInfo.Entries);
+            var tmpDate = selInfo["Last Add Time"].split(/\/|:| /);
```
yyscamper: Could you please add some example data here so that other developer will feel much easier to understand the parser code?

lib/jobs/ipmi-job.js
```diff
@@ -199,6 +224,98 @@ function ipmiJobFactory(
...
+            }
+            return [];
+        })
+        .then(function(unparsedSel) {
+            return parser.parseSelDataEntries(unparsedSel);
+        })
+        .then(function(parsedSel) {
+            var rawArr = [];
+            var rawData;
+            // we get the raw sel info for each sel entry in order to decode additional informaiton
```
yyscamper: typo: informaiton => information

lib/jobs/ipmi-job.js
```diff
@@ -199,6 +224,98 @@ function ipmiJobFactory(
...
+            if(sel) {
+                _.forEach(parsedSel, function(entry) {
+                    rawData = self.getRawSelEntry(data, entry["SEL Record ID"])
+                    .then(function(data){
+                        return data;
+                    });
+                    rawArr.push(rawData);
+                });
+            }
+            return Promise.all(rawArr);
```
yyscamper: For your awareness, The IPMI has max session defined (the actual value is varied by different implementation), if concurrent ipmi session exceeds the threshold, the subsequent commands will fail. So if there is too many SEL entries, some IPMI commands may fail. The retry can improve, but not a completely fix.

The fix may be complex (If only change Promise.all to Promise.each will bring performance lost), other module may have same potential problem. So this comment is only to keep you being aware of such potential problem. 

Any idea on this?

lib/jobs/ipmi-job.js
```diff
@@ -199,6 +224,98 @@ function ipmiJobFactory(
...
+                var rawSelRecord;
+                var mask = 1 << 7; // a mask to clear the 7th bit
+
+                rawArr[index] = value.replace(regex, '');
+                values = rawArr[index].split(" ");
+                rawSelRecord = values.slice(3,19);
+
+                //This is from table 32 of the IPMI spec v4
+                decodedSel["Event Type Code"] = 
+                    ("0" +(parseInt(rawSelRecord[12],16) & ~mask)).toString(16).slice(-2);
```
yyscamper: What's the purpose for adding the leading 0 `"0" + `?

lib/utils/job-utils/ipmi-parser.js
```diff
@@ -55,6 +55,40 @@ function parseSensorsDataFactory(assert, _) {
         return sel;
     }
 
+    function parseSelDataEntries(selData) {
```
yyscamper: Please add some sample data here to helper other developer understand the code.

brianparry:
:+1:

yyscamper:
@jlongever : I have a concern is about the performance. There may be a lot SEL entries for a node, if for each SEL entry to trigger an IPMI command to get detail sensor information, the whole process will be very time consuming, especially in the scale environment.
This is not a question for block this PR, but this will sure become a problem in future.

lib/jobs/ipmi-job.js
jlongever: The `getRawSelEntry()` is actually running synchronously inside the `_.forEach()`;  the `return Promise.all(rawArr)` is probably unnecessary here.   Not sure what the performance impact is here,  would need to profile it at scale.
FYI @tannoa2

jlongever:
@yyscamper Agreed.  If it's any consolation the code only collects new entries add and not the entire SEL.

lib/jobs/ipmi-job.js
brianparry: @jlongever @yyscamper  Couldn't we use `Promise.map` to avoid this problem?
```
rawArr = Promise.map(parsedSel, function(entry) {
    return self.getRawSelEntry(data, entry["SEL Record ID"]);
});
```

brianparry: Nevermind, this will have the same problem.

tannoa2: As Joel mentioned, this will only run against the new sel entries only

lib/utils/job-utils/ipmitool.js
tannoa2: Hi Felix, it should be clear from the name of the function

lib/jobs/ipmi-job.js
tannoa2: In order to make sure to always return a 2 digits hex value . A 7 would be returned as 07

lib/jobs/ipmi-job.js
jlongever: @yyscamper @brianparry @tannoa2 I went ahead and simplified this a little it and removed the redundant `Promise.all()`.

lib/jobs/ipmi-job.js
jlongever: @tannoa2 should probably pad it out based on the string size in case it returns 2 digits and you don't end up with `"0" + "6f" === "06f`.  

Pushed up a commit to do this.

jlongever:
@panpan0000 @changev Any reason why the Jenkins status shows failed but the last PR Jenkins build passed? http://rackhdci.lss.emc.com/job/on-tasks/1550/ ??

@amymullins

lib/utils/job-utils/ipmitool.js
yyscamper: @tannoa2 : I don't think it is self-explanation enough, not every developer is familiar with IPMI and let alone the raw command.  Even add something like "NetFn=0x0a, Command=0x43 is Get SEL Entry" will provide a lot clue if other developer want to lookup the detail command format.

jlongever:
test this please

lib/utils/job-utils/ipmitool.js
jlongever: @yyscamper If it helps,  I put a reference to the IPMIv2 spec in the comments: https://github.com/RackHD/on-tasks/pull/326/commits/25e180a0bf4510ae10416827985840548f8b4036#diff-a7628525a00f26023738040d65c193e5R189

I would hope any developer messing around with IPMI code would have some basic understanding of what it is they are touching.

panpan0000:
@jlongever , this is because @changev's dependent-PR sandbox Jenkins jobs encounter "PR flood" last night, and meet "job executor competition" even deadlock..
so @changev 's Jenkins jobs timeout aborted.
and no chance to write the current result status ( Production Jenkins Test result ) back to PR .

@changev is improving it now.
Thanks for reminder.

PR #335, opened 2016-09-13
Remove length and pattern check in set bmc credentials

keedya:
- variable assigned to this task are internal variables({{context.password}}),
  therefore checking patern or lenght is invalid in this situation


keedya:
@RackHD/corecommitters 
@yyscamper @brianparry @zyoung51 @jlongever @uppalk1 @tannoa2

benbp:
I'm a little confused by the commit message description. Does it mean we check the length elsewhere, or that it is impossible to check?

benbp:
@keedya tells me that the password itself is set to `{{ context.password }}`, which then gets rendered at runtime.

👍

yyscamper:
The task validation has already considered to skip context during upfront validation (https://github.com/RackHD/on-tasks/blob/master/lib/task.js#L434), this validation will only be done after this task is really scheduled. So at that time, the {{context.password}} has already been rendered and the validation will work well.

As for why I limit the password & username length, because this is defined in IPMI spec. If you want to know the detail, I can give you the particular section in spec.

So I don't think this is a problem due to task schema.
@keedya : could you please explain more for the issue you encountered?

At present, this PR is :-1

yyscamper:
@keedya : this is the IPMI spec: http://www.intel.com/content/www/us/en/servers/ipmi/ipmi-second-gen-interface-spec-v2-rev1-1.html

The username is in Section 22.28 "Set User Name Command", the password is in Section "22.30 Set User Password Command"

panpan0000:
Jenkins shows unit-test failure , please notify me if it's a false alarm, thank you.

keedya:
@yyscamper validation is skipped in the beginning when task is called by id, the variable for password will be {{ context.password }}. This is a end variable that won't change unless it is rendered in a script

keedya:
@yyscamper, this has nothing to do with the real password that we pass to ipmitool. I am aware of the bmc and ipmitool specs

keedya:
@yyscamper  the failure is related to the encryption added in the following commit:
https://github.com/RackHD/on-core/commit/06638621b02727db77c57ad2493640b6b41806ab

I will open a github issue to track it
Closing this PR

PR #340, opened 2016-09-18
Fix issue that task annotation doesn't render data

yyscamper:
resolves https://github.com/RackHD/RackHD/issues/418

@RackHD/corecommitters @WangWinson @heckj @dmskinner


rackhdmirror:
Build finished.

WangWinson:
:+1: 
The change itself looks good. But this seems different from the reason Joe commented in the issue.

yyscamper:
@WangWinson : That's because the input task objects are not correct, so the filter doesn't generate correct data. We should fix the input not the filter.

PR #341, opened 2016-09-19
Added ability to get specified catalog values and map to specified keys

dmskinner:


rackhdmirror:

brianparry:
ok to test

rackhdmirror:
Build finished.

lib/jobs/get-catalog-values-job.js
```diff
@@ -0,0 +1,85 @@
...
+            newObj = {};
+
+        // Get the correct catalog based on source
+        _.forEach(self.options.requestedData, function(catalogInfo){
+            waterline.catalogs.findOne({
+                'node': self.nodeId, 'source': catalogInfo.source
+            })
+                .then(function(catalog){
+                    _.assign(newObj, _.mapValues(catalogInfo.keys, function(val){
+                        return _.get(catalog, val, '');
```
brianparry: Hi @dmskinner Minor comment: would it make more sense to leave the value undefined if it doesn't exist rather than specifying a default?

lib/jobs/get-catalog-values-job.js
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * @memberOf GetCatalogValuesJob
+     */
+    GetCatalogValuesJob.prototype._run = function _run(){
+        var self = this,
+            newObj = {};
+
+        // Get the correct catalog based on source
+        _.forEach(self.options.requestedData, function(catalogInfo){
+            waterline.catalogs.findOne({
```
yyscamper: Since there may be multiple versions of catalog for same source, so please change to `findMostRecent`
You can refer to: https://github.com/RackHD/on-tasks/blob/master/lib/jobs/provide-catalog-data.js#L60

lib/jobs/get-catalog-values-job.js
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * @memberOf GetCatalogValuesJob
+     */
+    GetCatalogValuesJob.prototype._run = function _run(){
+        var self = this,
+            newObj = {};
+
+        // Get the correct catalog based on source
+        _.forEach(self.options.requestedData, function(catalogInfo){
```
yyscamper: The promise array doesn't work as your expectation. Actually the job is done after the first requestData is successful regardless of the subsequent requestedData. Change `_.forEach` to `Promise.each` or `Promise.reduce` will be better:
```javascript
return Promise.reduce(self.options.requestedData, function(catalogInfo) {
    return waterline.catalogs.findMostRecent(.....)
    .then(...)
    .then(...);
}, {})
.then(function(results) {
    //merge the results to context
    self._done();
})
.catch(function(err) {
    self._done(err);
});
```

lib/jobs/get-catalog-values-job.js
```diff
@@ -0,0 +1,85 @@
...
+            waterline.catalogs.findOne({
+                'node': self.nodeId, 'source': catalogInfo.source
+            })
+                .then(function(catalog){
+                    _.assign(newObj, _.mapValues(catalogInfo.keys, function(val){
```
yyscamper: It's not problem, a minor suggestion, use `_.reduce` will be simpler here.

lib/task-data/schemas/get-catalog-values.json
```diff
@@ -0,0 +1,33 @@
...
+    "copyright": "Copyright 2016, Adobe Systems Incorporated",
+    "title": "Get Catalog Values",
+    "description": "The schema for getting requested catalog values",
+    "describeJob": "Job.Get.Catalog.Values",
+    "allOf": [
+        { "$ref": "common-task-options.json#/definitions/Options" },
+        {
+            "type": "object",
+            "properties": {
+                "requestedData": {
```
yyscamper: Add a `description` for this property will help the task annotation.

spec/lib/jobs/get-catalog-values-job-spec.js
```diff
@@ -0,0 +1,122 @@
...
+            },
+            uuid.v4()
+        );
+        sandbox.stub(jobObject, '_subscribeActiveTaskExists').resolves();
+
+        return jobObject.run()
+           .then(function(){
+                expect(jobObject.context.data.fail_case).to.be.empty;
+            })
+           .catch(function(err){
```
yyscamper: If you expect the promise to be fulfilled, then remove this `catch` so that the mocha will fail this test case automatically.

yyscamper: The mocha will also dump the error message, so you don't need to manually print the error object.

yyscamper:
In summary, this job is very similar with the existing job [provide-catalog-data](https://github.com/RackHD/on-tasks/blob/master/lib/jobs/provide-catalog-data.js), and this is the task input example [provide-catalog-ami-bios-version](https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/provide-catalog-ami-bios-version.js), your job can be an advance version of existing job since your PR supports multiple paths.

To avoid duplication, I suggesting modifying existing job or removing existing job after your PR is merged.    Either is OK for me.

lib/jobs/get-catalog-values-job.js
dmskinner: @yyscamper I was hoping you could expound because I'm not actually seeing the same behavior that you are indicating. When I provide multiple objects in the requestedData array, the `_.forEach` successfully iterates over them and returns all the requested data.

Could you also clarify why you are recommending using `Promise.reduce` rather than the map? I might be misunderstanding the documentation for reduce but it doesn't seem to line up with what the job is trying to accomplish. The job will map the located catalog values to the user provided keys.

Given the input as the requstedData array:
```
[
    {
		"source": "ohai",
		"keys"{
			"myDefinedKey": "data.ipaddress",
			"myOtherDefinedKey": "data.cpu.0.cores"
		}
	},
	{
		"source": "lshw",
		"keys": {
			"randomName": "data.handle"
	}
]
```
It would return: 
```
{
	"myDefinedKey": "1.1.1.1",
	"myOtherDefinedKey": "1",
	"randomName": "DMI:0001"
}
```

I am open to suggestions on how to simplify this section because I know it is a bit complex.

lib/jobs/get-catalog-values-job.js
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * This job finds one or more requested values from one or more specified catalogs.
+     * The catalog values are made available in the shared context in an object that 
+     * maps the catalog values to requester provided keys.
+     *
+     * Example requestedData array:
+     * [{
+     *   "source": "ohai",
+     *   "keys": {
+     *     "myIPAddrKeyName": "data.ipaddress"
```
dmskinner: The job was designed to be as flexible as possible to allow access to the any of the values in the catalog, not just the ones within the `data` property. If you don't think there is any value to allowing access to all the catalog properties, I can update the job to not require `data` as part of the JSON path.

dmskinner:
@yyscamper  I spent quite a bit of time looking at extending or modifying the [provide-catalog-data](https://github.com/RackHD/on-tasks/blob/master/lib/jobs/provide-catalog-data.js) job. They are similar, as you mentioned, but the new job has the ability to support multiple paths. I ended up not modifying the provide-catalog-data job because the two jobs are ultimately accomplishing two different things. The new job not only gets multiple values from different locations but maps the values to user provided keys. 

I felt since it was a drastic change to the requested input and the format of the return value that it might be better to add a new job. I also didn't feel like it was my call to request to have the other job removed because it is being used by tasks within, and potentially outside, the project.

lib/jobs/get-catalog-values-job.js
yyscamper: For catalog, we usually only use its data. Except `data` and `source`, other properties are all generated by database, like the database identifier, created time and modified time. Do you have any specific user case that may use the other properties?

Of course, I am OK to keep your design, as it covers more user cases.

lib/jobs/get-catalog-values-job.js
yyscamper: @dmskinner : Actually, this is a problem about how to control the timing sequence using Promise, I will use following example to describe what's the problem:
```javascript
var _ = require('lodash');
var Promise= require('bluebird');

var requestedData = ['a', 'b', 'c'];
var context = {};
_.forEach(requestedData, function(data) {
    Promise.resolve(data)
    .then(function(val) {
        return val + '1';
    })
    .then(function(val) {
        return val + '2';
    })
    .then(function(val) {
        context[data] = val;
        console.log("call _done!");
    })
    .catch(function(err) {
        console.log("call _done(err)!");
    });
});

console.log('context #1:');
console.log(context);

setTimeout(function() {
    console.log('context #2');
    console.log(context);
}, 2000);
```

The output is
```
context #1:
{}
call _done!
call _done!
call _done!
context #2
{ a: 'a12', b: 'b12', c: 'c12' }
```

The problems are:
- The _done is called thrice, actually we only allow it is called after all catalogs are handled.
- The context has no data when the _.forEach finishes, because the javascript code is async.

yyscamper:
@dmskinner : I know some task is currently using the job `provide-catalog-data`, so we need some testing if removing it. It's OK to keep your current job and I can create an action-item to remove `provide-catalog-data` in future.

yyscamper:
So the only must-have change is to fix the promise timing.

lib/jobs/get-catalog-values-job.js
dmskinner: That's interesting because when I run the graph it is actually going through each item in the array just fine. However, it sounds like that is not the desired behavior, so regardless I need to implement it differently. I'll change the `_.forEach` to `Promise.each` so that I'm only calling `_done` once. It is definitely a better solution. Thank you for the feedback.

If you are interested... I was curious so I did some investigating. It seems the reason mine behaves as I expected, excluding the  multiple calls to `_done`, is related  to how the `_.mapValues`  & `_.assign` behave. The  `_.assign` to update the newObj isn't called until all the values are mapped to the keys. 

I added some logging into the the Job:
```
 GetCatalogValuesJob.prototype._run = function _run(){
        var self = this,
            newObj = {};

        // Get the correct catalog based on source
        _.forEach(self.options.requestedData, function(catalogInfo){
            waterline.catalogs.findMostRecent({
                node: self.nodeId, source: catalogInfo.source
            })
                .then(function(catalog){
                    _.assign(newObj, _.mapValues(catalogInfo.keys, function(val){
                        console.log('*** What is my val ' + val);
                        return _.get(catalog, val, null);
                    }));
                })
                .then(function(){
                    self.context.data = newObj;
                    console.log('*** CATALOG *** ' + catalogInfo.source + '\n***** CONTEXT.DATA ***** ' + JSON.stringify(self.context.data));
                    self._done();
                })
                .catch(function(err){
                    self._done(err);
                });
        });
    };
```
And when I run the consuming task it produces the following:
```
job:    Job.Get.Catalog.Values
*** What is my val data.cpu.0.cores
*** What is my val data.ipaddress
*** CATALOG *** ohai
***** CONTEXT.DATA ***** {"cpu_cores":"1","ip":"172.31.128.2"}
*** What is my val data.handle
*** CATALOG *** lshw
***** CONTEXT.DATA ***** {"cpu_cores":"1","ip":"172.31.128.2","test_val":"DMI:0001"}
*** What is my val [data][Chassis Information][Type]
*** CATALOG *** dmi
***** CONTEXT.DATA ***** {"cpu_cores":"1","ip":"172.31.128.2","test_val":"DMI:0001","two_words":"Other"}
[info] [2016-09-28T16:54:08.799Z] [on-taskgraph] [TaskGraph.TaskRunner] [Server] Task finished
```

lib/jobs/get-catalog-values-job.js
dmskinner: I personally don't have any specific use case for utilizing the other properties, but just thought it might be beneficial for someone to be able to access all the information in the property, especially the created/modified times. I am not as familiar with how others use the catalog data, so it may be unnecessary. I'm good with either route.

lib/jobs/get-catalog-values-job.js
yyscamper: @dmskinner : For only one catalog source, the log does show the sequence is OK. The key root cause is the _done is called multiple times. Actually, RackHD will immediately finish the job if `_done` is called, so that the next job will be scheduled to run. Image the next job is just want to use the fetched catalog data, the data may not be ready at that time (just check my example code of `console.log('context after forEach:');`). Of course, if there is enough time before the next job uses the context value, you may not realize there is such problem.
In summary, Make sure the `_done` is called only once is key principle for job.

lib/jobs/get-catalog-values-job.js
```diff
@@ -0,0 +1,88 @@
...
+            return waterline.catalogs.findMostRecent({
+                node: self.nodeId,
+                source: catalogInfo.source
+            })
+                .then(function(catalog){
+                    _.assign(newObj, _.mapValues(catalogInfo.keys, function(val){
+                        return _.get(catalog, val, null);
+                    }));
+                });
+        })
```
yyscamper: Could you please kindly align the code?  It's not easy to understand the code if the code is not well-aligned.

yyscamper:
jenkins: test this please

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

PR #343, opened 2016-09-21
Add ping iDRAC IP feature after update firmware to make sure iDRAC re…

pengz1:
This PR is to fix ODR895 and ODR878.
For ODR878 it is found immediate power cycle after transferring iDRAC firmware to Dell iDRAC will cause BIOS-iDRAC connection failure during update, thus a 5 seconds delay is added between image transferring-done and power cycle the server.
For ODR895 it is found with latest iDRAC firmware, iDRAC IP/password/usename will be lost during iDRAC firmware update. A "ping" like feature is added in Dell updateFirmware function so that after iDRAC start update firmware, RackHD will first try to ping iDRAC. Only after RackHD can ping the iDRAC IP successfully that we will continue following steps.

### DIFF
--- a/lib/utils/job-utils/racadm-tool.js
+++ b/lib/utils/job-utils/racadm-tool.js
@@ -125,9 +125,6 @@ function racadmFactory(
     RacadmTool.prototype.waitConnectBack = function(host, user, password, retryCount, delay) {
         var self = this;
         return self.runCommand(host, user, password, "help")
-        .then(function(){
-            return Promise.resolve();
-        })
         .catch(function() {
             if (retryCount < self.retries) {
                 return Promise.delay(delay)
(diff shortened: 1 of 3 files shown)


rackhdmirror:

pengz1:
@RackHD/corecommitters @cgx027 @linlynn

anhou:
@pengz1  who you want to @ is  @iceiilin , not @linlynn :-)

iceiilin:
:+1:

lib/utils/job-utils/racadm-tool.js
```diff
@@ -235,16 +265,28 @@ function racadmFactory(
...
-            })
-            .then(function(jobId){
-                return self.waitJobDone(host, user, password, jobId, 0, 1000);
-            });
+        
+        rebootFlag = (_.has(config, "forceReboot")) ? config.forceReboot : true;
+        
+        if (rebootFlag) {
+            return self.runCommand(host, user, password, command, 0, 1000)
+                .delay(5000) //Add 5s buffer for iDRAC responding
```
yyscamper: Talked with @pengz1 , this value is enough in our testing.

PR #349, opened 2016-09-28
Upgraded solution for ODR895

pengz1:
Sunny reported in ODR 895 we still have issue for iDRAC firmware update with 2.1.x branch release.

This is an upgraded solution for ODR-895 to bypass errors reported when get racadm job status. This is feasible as we have 10 retries to get racadm job status. If all retries failed, the failure will reported by RackHD


pengz1:
@cgx027 @iceiilin @RackHD/corecommitters

rackhdmirror:

PR #352, opened 2016-10-05
Replace the rawread  ipmitool command and the "sel get" with writeraw.

tannoa2:
This solution solve the issue of the FW on certain HW that can't read the first sel entry.
In addition, it stabilize the pollers by running less ipmi comamnds

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,11 +5,13 @@
 var di = require('di'),
     _ = require('lodash'),
     core = require('on-core')(di);
+var temp = require('temp');
 
 module.exports = {
     injectables: _.flattenDeep([
         core.helper.requireWrapper('ssh2', 'ssh', undefined, __dirname),
         core.helper.requireWrapper('dockerode', 'Docker', undefined, __dirname),
+        core.helper.simpleWrapper(temp, 'temp'),
         core.helper.requireGlob(__dirname + '/lib/*.js'),
         core.helper.requireGlob(__dirname + '/lib/jobs/*.js'),
         core.helper.requireGlob(__dirname + '/lib/utils/**/*.js'),
--- a/lib/jobs/ipmi-job.js
+++ b/lib/jobs/ipmi-job.js
@@ -16,7 +16,10 @@ di.annotate(ipmiJobFactory, new di.Inject(
     'Promise',
     '_',
     'Services.Waterline',
-    'JobUtils.PollerHelper'
+    'JobUtils.PollerHelper',
+    'fs',
+    'temp'
+
 ));
 
 function ipmiJobFactory(
@@ -29,7 +32,10 @@ function ipmiJobFactory(
     Promise,
     _,
     waterline,
-    pollerHelper
+    pollerHelper,
+    fs,
+    temp
+
 ) {
     var logger = Logger.initialize(ipmiJobFactory);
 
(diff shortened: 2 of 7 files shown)


rackhdmirror:

tannoa2:
@jlongever @zyoung51 @brianparry @amymullins @uppalk1 @keedya

tannoa2:
I would like to get some reviews, but please don't merge the PR yet @RackHD/corecommitters

tannoa2:
@stuart-stanley @hohene @larry-dean

index.js
```diff
@@ -5,11 +5,13 @@
 var di = require('di'),
     _ = require('lodash'),
     core = require('on-core')(di);
+var temp = require('temp').track();
```
stuart-stanley: Please remove ".track()". The temp module's track feature is only usable for shorter-run processes to delete all temp files _when the process exits_. In our case, that array of file names (already deleted or otherwise) will OOM the process/machine.

lib/jobs/ipmi-job.js
```diff
@@ -312,42 +228,161 @@ function ipmiJobFactory(
...
+            .then(function(){
+                if (totalEntries !== 0 && selCleared === false ) {
+                    //It skips one poll when the SEL is being cleared
+                    var readFile = Promise.promisify(fs.readFile);
+                    var unlink = Promise.promisify(fs.unlink);
+                    temp = Promise.promisifyAll(temp);
+                    var filePath;
+                    return temp.openAsync('tmp-ipmi-data')
+                        .then(function (info) {
+                            return info.path;
```
stuart-stanley: There needs to be a line inserted before the "return info.path":
fs.close(info.fd);

(the temp module doesn't actually track the file itself and clean up when it goes out of scope, so if we don't do this, we will leak all the file descriptors).

tannoa2:
test this please please please

larry-dean:
+1

stuart-stanley:
+1 for file/file-descriptors bits.

lib/utils/job-utils/ipmitool.js
```diff
@@ -186,17 +186,14 @@ function ipmitoolFactory(Promise) {
...
 
     /**
-     * Returns the raw SEL entry data, see section 31.5 in the IPMIv2 specification
+     * Returns a promise with the results of the user ipmitool command
      *
      * @param host
      * @param user
      * @param password
-     * @param Sel Entries
+     * @param commadn
```
geoff-reid: should be "command"

amymullins:
@RackHD/corecommitters , this is resolution for the P1 ODR -931 and considered a gate for the 2.1.0 release.  can you please help with an urgent review?

lib/jobs/ipmi-job.js
```diff
@@ -312,42 +228,164 @@ function ipmiJobFactory(
...
+                            logger.debug("***number of new SEL entries :" + newEntries.length +
+                                ". Poller: "+data.workItemId+ " nodeID: " + workObject.node );//jshint ignore: line
+                            return newEntries; //returning the new  SEL entries
+                        }).
+                        catch(function(err){
+                            logger.error("Failed to process SEL. Poller: " +data.workItemId +
+                                " nodeID: " + workObject.node + ". Error message: " + err.message);
+                        })
+                        .finally(function () {
+                            if(filePath){
```
brianparry: @tannoa2 If you're using fs methods to cleanup temp files, you can just use `temp.path()` to generate a unique file name.  That won't open the file either so you won't need to close it.

stuart-stanley: Truth sir! However, it introduces a cluster of other cleanup paths to deal with. Really, we should use 'tmp' rather than 'temp', but this is critical path. :)

PR #358, opened 2016-10-17
Validate ssh fix. Avoid trying to ssh when host ip is null

nucklehead:
When the lookup does not have an ip address (the feild is null), the code below will trigger the 'ready' event and not the 'error'.
```
conn.connect(_.defaults({
                    readyTimeout: self.timeout,
                    username: sshSettings.user
                }, sshSettings)
```
This will mark the validate-ssh task as complete which is completely false. 
This PR is to add a check in the ready event. Then the code will continue trying to ssh with the rest of the lookup entries.


nucklehead:
@mtannous

rackhdmirror:

nucklehead:
I believe the coverage went down because of the few lines I added.
Does that mean the test does not cover the case where the ip in the lookup is not null?

keedya:
:+1:

lib/jobs/validate-ssh.js
```diff
@@ -123,6 +123,9 @@ function validationJobFactory(
             };
             conn.on('ready', function() {
                 conn.end();
+                if(sshSettings.host === null){
+                    reject('host ip is not defined in this lookup.');
```
anhou: 'lookup' may be some other words, 'lookup' in RackHD is easy to be understand as lookup documents.

keedya:
@RackHD/corecommitters @brianparry @benbp @tannoa2

brianparry:
ok to test

PR #354, opened 2016-10-11
[DONOT MERGE] POC of Task Schema Redesign

yyscamper:
**DONOT MERGE BEFORE I REMOVE THE POC TAG**

This is the POC PR for task schema redesign.  Please provide further comment.

(1) Job schema is put in baseTask
(2) Support inline task schema

**NOTE**: Comparing with my initial proposal, in the POC, I decided to keep the baseTask and put the job schema at baseTask. so that baseTask can be thought of an interface wrapper of job. User can know all options and how to use the job without touching the job code. Another benefit is this can minimize the code change.

@RackHD/corecommitters @leoyjchang  @amymullins @keedya @zyoung51 @jlongever @iceiilin @pengz1 @cgx027 @lanchongyizu 

jenkins: ignore

PR #360, opened 2016-10-19
Task schema refactor

yyscamper:
# NOTE

This is a pretty big PR, sorry to bring some trouble for reviewer. I intentionally create this big PR to combine all changes about the task-schema refactor design because I don't want to bring confusion for user & developer if both old schema and new schema coexist. 
Actually the core code  is in my POC (see https://github.com/RackHD/on-tasks/pull/354), you can first get the general structure design about new task-schema from POC. The most additional work of this PR comparing with the POC is just to port all existing schema to new one, so you will see most code change is similar.
# Change List

Following is the change list for this PR:
    1. split into common/job/task-specific schemas
    2. add a key 'optionsSchema' in both task and baseTask definition to store schema.
    3. job/task-specific schema is optional
    4. support built-in schema and reference external schema file.
    5. refactor install-os schemas design.
    6. add schema for Windows server installation.
    7. change unit-test accordingly.
    8. change task-annotation accordingly.

@RackHD/corecommitters @iceiilin @pengz1 @cgx027 @lanchongyizu @amymullins 

@tannoa2 : I also added the schema for windows installation, please help review it, the file is: https://github.com/yyscamper/on-tasks/blob/task-schema-refactor/lib/task-data/schemas/install-windows.json

jenkins: depends on https://github.com/RackHD/on-core/pull/217 https://github.com/RackHD/on-taskgraph/pull/176


rackhdmirror:
Build finished.

yyscamper:
The travisci and jenkins failure is expected, since it depends on the change of [on-core](https://github.com/RackHD/on-core/pull/217)

changev:
test this please

rackhdmirror:
Build finished.

rackhdmirror:
Build finished.

changev:
test this please

rackhdmirror:
Build finished.

changev:
test this please

rackhdmirror:
Build finished.

changev:
test this please

rackhdmirror:
Build finished.

amymullins:
This is still a -1 for me.  The complexity of the design/redesign continues to carry a very large development/support, usability, and long term maintenance cost that is highly disproportionate to its business value for API Input Validation.  Several simplification suggestions were provided in previous Core Committer meetings and email threads.  I request we revisit the topic.

yyscamper:
@brianparry : For why I don't choose the inline schema for each option, I have replied you in the on-core PR: https://github.com/RackHD/on-core/pull/217 

As for why I still keep the base-task, I explained using the MVC. The baseTask is designed as the "viewer" layer of job, user don't need to read the job code but can still know how to use the job. Though now the baseTask only contains the job schema, but it can be easily to extend to put some annotation here in future, for example: 
(1) A job may depend on some external tools, such as `ipmitool`, `snmp`, `racadm`, we can annotate the requiredTools in baseTask, so before a job (or even the workflow) starts, we can check whether those tools existence and fail earlier.
(2) Current auto generated taskdoc can only be used to as a simple doc to explain each option, but if for a job, there is some very detail manual guide, we can annotate a doc link or doc content in BaseTask.

Of course, all above bases on my thought on future baseTask, and I agree the baseTask is not the only way for above future features ( if we agree it is a valid feature).  At present, the biggest problem to keep baseTask is it is easy to diverge because the job code and job schema are put in different source folder, developer may forget to change both.

In a whole, I am not against to remove the baseTask, If we all agree it's not necessary to keep baseTask, I can put it into my plan to remove it. Actually, to remove the baseTask (and moving job schema to job code itself) will introduce a lot change including:
 - task.js/task-graph.js
 - Every job code (move job schema to job)
 - Every job unit-test (test job schema)
 - Every built-in task definition (change `implementsTask` to job's `injectableName`)
 - Every workflow inline task definition (change `implementsTask` to job's `injectableName`)
 - Every task definition in on-skupack (change `implementsTask` to job's `injectableName`)
 - Even the property name `implementsTask` in task definition need better change to `runJob`!

So my suggestion is even we decide to remove baseTask, let's keep it in another PR.

PengTian0:
test this please

changev:
test this please

brianparry:
@yyscamper Specifying tooling requirements is an interesting use of base jobs that I hadn't considered.  Although, don't you still have the problem of the job code and job specification living in two different places where one could be changed without the other?  In any case, I'm okay with discussing that separately and making that another PR, @amymullins What are your thoughts here?  Any objection to deferring the removal of base jobs?

amymullins:
@brianparry and @yyscamper I have no objection to the deferral.   Removal of the base jobs was part of the original discussions/requirements, so we probably won't want to defer for too long.

iceiilin:
+1 as long as all tests run smoothly

PR #364, opened 2016-10-21
On tasks changes for secure erase workflow progress

pengz1:
jenkins: depends on https://github.com/RackHD/on-core/pull/221
https://github.com/RackHD/on-taskgraph/pull/177
https://github.com/RackHD/on-http/pull/507

On-tasks changes for secure erase workflow progress. Includes modification of secure erase job/task changes and some common changes for sending progress events.

### DIFF
--- a/lib/task-data/tasks/secure-erase-drive.js
+++ b/lib/task-data/tasks/secure-erase-drive.js
@@ -7,7 +7,8 @@ module.exports = {
     injectableName: 'Task.Drive.SecureErase',
     implementsTask: 'Task.Base.SecureErase',
     options: {
-        eraseSettings: []
+        eraseSettings: [],
+        baseUri: '{{ api.server }}'
     },
     properties: {}
 };
--- a/lib/task-graph.js
+++ b/lib/task-graph.js
@@ -525,8 +525,6 @@ function taskGraphFactory(
     };
 
     TaskGraph.prototype.updateGraphProgress = function(progressData) {
-        assert.string(progressData.graphName); 
-        
         return store.updateGraphProgress(progressData)
         .tap(function(){
             //For graph start and finished, no task progress data is include
--- a/lib/task-graph.js
+++ b/lib/task-graph.js
@@ -524,13 +524,33 @@ function taskGraphFactory(
         }));
     };
 
-    TaskGraph.prototype.updateGraphProgress = function(progressData) {
+    /**
+     * Update tasks dependencies and graph object progress data, send progress AMQP message
+     * @param {Object} an object of progress data that should include graphId and progress data
+     * @returns {Promise}
+     */ 
+    TaskGraph.updateGraphProgress = function(progressData) {
+        var pattern = /(100|\d{1,2})(\.\d{0,})?\%/;
+        assert.uuid(progressData.graphId, "Progress data should include graphId");
+        assert.object(progressData.progress, "Progress data should include progress data");
+        if (_.has(progressData, 'progress.percentage')) {
+            if (!progressData.progress.percentage ||
+                !progressData.progress.percentage.match(pattern)) {
+                progressData.progress.percentage = "Not Available";
+            }
+        }
+        if (_.has(progressData, 'taskProgress.progress.percentage')) {
+            if (!progressData.taskProgress.progress.percentage ||
+                !progressData.taskProgress.progress.percentage.match(pattern)) {
+                progressData.taskProgress.progress.percentage = "Not Available";
+            }
+        }
         return store.updateGraphProgress(progressData)
         .tap(function(){
             //For graph start and finished, no task progress data is include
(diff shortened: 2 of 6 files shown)


rackhdmirror:

changev:
test this please

changev:
test this please

rackhdmirror:
Build finished.

changev:
test this please

rackhdmirror:
Build finished.

changev:
test this please

rackhdmirror:
Build finished.

pengz1:
@cgx027 @iceiilin

pengz1:
@benbp @brianparry

lib/task-data/tasks/secure-erase-drive.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
     injectableName: 'Task.Drive.SecureErase',
     implementsTask: 'Task.Base.SecureErase',
     options: {
-        eraseSettings: []
+        eraseSettings: [],
+        baseUri: '{{ file.server }}'
```
cgx027: In the case where external file server is used, file.server will be the external file server address and the notification will not be sent to correct server. Replace this with api.server sounds more corect. @iceiilin

cgx027: I would also suggest to replace ```baseUri``` with ```notificationUri``` and append the sub-router into the tail, so that the API uri will be limited in RackHD knowledge scope. If the notification API changed in the API in the future, people will not need to modify the secure erase. python script. Not a  blocking comment anyway.

yyscamper: I agree with @cgx027 's suggestion.
Even more, we can add a new template item like "{{api.notification}}"

lib/task-graph.js
```diff
@@ -522,6 +524,35 @@ function taskGraphFactory(
         }));
     };
 
+    TaskGraph.prototype.updateGraphProgress = function(progressData) {
+        var pattern = /\d{1,3}\.?\d{0,2}\%/;
```
yyscamper: why not restrict the digit number after the decimal point? I think both 12.89% and 12.892% is OK

pengz1: Here I want to set restriction on all progress data format to make it look clean. Two digital after point should be enough.

lib/task-graph.js
```diff
@@ -522,6 +524,35 @@ function taskGraphFactory(
         }));
     };
 
+    TaskGraph.prototype.updateGraphProgress = function(progressData) {
```
yyscamper: This is a potential problem for this prototype function is that there is no reference for `this`.  Seems the input `progressData` has already included the graphId, but you don't check whether the input graphId is matched with this graph instance, so that to call this API, the progress of graph A may be wrongly updated into graph B.

You can either add some assertion logic to make sure the graph id is matched, or you can auto-fill the graphId for `progressData`, so that the `progressData` doesn't need contain the `graphId`.

lib/task-graph.js
```diff
@@ -522,6 +524,35 @@ function taskGraphFactory(
...
+        var pattern = /\d{1,3}\.?\d{0,2}\%/;
+        if (_.has(progressData, 'progress.percentage')) {
+            if (!progressData.progress.percentage.match(pattern)) {
+                progressData.progress.percentage = "Not Available";
+            }
+        }
+        if (_.has(progressData, 'taskProgress.progress.percentage')) {
+            if (!progressData.taskProgress.progress.percentage.match(pattern)) {
+                progressData.taskProgress.progress.percentage = "Not Available";
```
yyscamper: the variable name is a little verbose , `taskProgress` has already showed it is progress, isn't `progressData.taskProgress.percentage` better?

lib/task-graph.js
```diff
@@ -522,6 +524,35 @@ function taskGraphFactory(
         }));
     };
 
+    TaskGraph.prototype.updateGraphProgress = function(progressData) {
+        var pattern = /\d{1,3}\.?\d{0,2}\%/;
+        if (_.has(progressData, 'progress.percentage')) {
+            if (!progressData.progress.percentage.match(pattern)) {
+                progressData.progress.percentage = "Not Available";
```
pengz1: on-core is aligned to Not Available

changev:
test this please

rackhdmirror:
Build finished.

yyscamper:
jenkins: test this please

pengz1:
test this please

yyscamper:
Since the on-core has been merged, so there is no dependency for this PR now, the outstanding dependency problem can be ignored.  

Meanwhile, communicated with @brianparry , he agreed to merge as it is. We will have some follow-up discussion about how to improve the progress reporting.

PR #369, opened 2016-11-03
make ssh private key optional

lanchongyizu:
Fix ODR-972:
Since we use public key as privateKey to ssh into node, which makes ssh client throw
`Error: privateKey value does not contain a (valid) private key`
So it doesn't use user/password at all. 

The solution is making ssh private key optional.

@RackHD/corecommitters @panpan0000 @benbp @iceiilin

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -127,7 +127,7 @@ function installOsJobFactory(
             var rootUser = {
                 name: 'root',
                 password: this.options.rootPassword,
-                privateKey: this.options.rootSshKey
+                publicKey: this.options.rootSshKey
             };
             this.context.users = _.compact((this.context.users || []).concat(rootUser));
         }
--- a/lib/jobs/sftp-job.js
+++ b/lib/jobs/sftp-job.js
@@ -105,16 +105,19 @@ function sftpJobFactory(
                 }
             });
 
-            ssh.connect({
+            var sshConfig = {
                 host: sshSettings.host,
                 port: sshSettings.port || 22,
                 username: sshSettings.user,
                 password: cryptService.decrypt(sshSettings.password),
-                privateKey: cryptService.decrypt(sshSettings.privateKey),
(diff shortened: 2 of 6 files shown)


iceiilin:
:+1:

lanchongyizu:
test this please

lanchongyizu:
test this please

keedya:
Jamie, how can the user upload the private keys?

amymullins:
@lanchongyizu  as discussed in the email thread, we should be removing ssh validation instead of making it optional in the workflow.  There is still a large security concern passing the private key in the payload.

lanchongyizu:
@keedya @amymullins Do you agree with removing private key from RackHD?

lanchongyizu:
The comments from @johren in ODR-972:

> This is my viewpoint on dealing with the SSH keys (both public and private):
> 1. SSH PUBLIC keys should remain in the OS install payload. This option is to allow the client to specify a 
> public key that they want installed on the compute node such that the client (not RackHD) can use their 
> private key to access the node.
> 2. RackHD does not need to have (and should not have) the client's private key.
> 3. If RackHD needs to access the node with an SSH key pair (for inband management/ansible) we should generate a separate key pair for RackHD. RackHD will keep that private key and install 
> the corresponding public key on the node.
> So, coming from this viewpoint, I'm fine with https://github.com/RackHD/on-tasks/pull/369 and
> https://github.com/RackHD/on-http/pull/520 (other than not needing to encrypt/redact the public key) as long as we are not asking the client to upload their own private keys. These private keys should be ones generated by RackHD.

panpan0000:
@lanchongyizu , so the title should be changed to "make ssh **public** key optional in bootstrap payload" ?

lanchongyizu:
@panpan0000 No, the Public Key has already been optional, this PR is to make the Private Key optional so that RackHD works fine no matter whether Private Key exists or not.

PR #371, opened 2016-11-04
fix local catalog schema reference error

yyscamper:
This is a part of fix for https://github.com/RackHD/RackHD/issues/483

Root cause: the local catalog task are wrongly pointed to the linux-catalog job.

@RackHD/corecommitters @iceiilin @cgx027 @pengz1 @manfrednde


iceiilin:
:+1:

PR #372, opened 2016-11-04
Do lookups for host MAC addresses when needed.

jeamland:
This is a direct rip of the lookupHost method in the IPMI OBM service.

### DIFF
--- a/lib/jobs/ipmi-catalog.js
+++ b/lib/jobs/ipmi-catalog.js
@@ -18,8 +18,7 @@ di.annotate(ipmiCatalogJobFactory, new di.Provide('Job.LocalIpmi.Catalog'));
         '_',
         'Services.Encryption',
         'ChildProcess',
-        'Services.Lookup',
-        'Constants'
+        'JobUtils.JobHelpers'
     )
 );
 function ipmiCatalogJobFactory(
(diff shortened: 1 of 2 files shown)


brianparry:
ok to test

lib/jobs/ipmi-catalog.js
```diff
@@ -59,10 +63,20 @@ function ipmiCatalogJobFactory(
...
 
         return waterline.obms.findByNode(self.nodeId, 'ipmi-obm-service', true)
-            .then(function (obmSettings) {
-                assert.object(obmSettings, 'No OBM service available.');
-                assert.ok(!_.isEmpty(obmSettings.config), 'No OBM service config available.');
+            .then(function(settings){
+                assert.object(settings, 'No OBM service available.');
+                assert.ok(!_.isEmpty(settings.config), 'No OBM service config available.');
+
+                if (settings.config.host && Constants.Regex.MacAddress.test(settings.config.host)) {
```
yyscamper: We have already created a util function to handle this conversion, see https://github.com/RackHD/on-tasks/blob/master/lib/utils/job-utils/job-helper.js#L31, please change it to that function to avoid duplicated code.

PR #374, opened 2016-11-16
refactor southbound templates api

lanchongyizu:
It's related with https://github.com/RackHD/on-http/pull/531.

@RackHD/corecommitters @panpan0000 @iceiilin @keedya @johren

### DIFF
--- a/lib/task-data/schemas/linux-command.json
+++ b/lib/task-data/schemas/linux-command.json
@@ -41,9 +41,9 @@
                     "minimum": 0
                 },
                 "downloadUrl": {
-                    "description": "Specify the download URL (relative to RackHD southbound API server) if the command needs to download some data or script from RackHD before execution",
+                    "description": "Specify the download URL if the command needs to download some data or script before execution",
                     "type": "string",
-                    "pattern": "^/[a-zA-Z0-9+-.#?=\\\/]+"
+                    "minLength": 1
                 },
                 "acceptedResponseCodes": {
                     "description": "The acceptable exit codes to mark the command success. Defaultly, exit code 0 means the command success, otherwise failure. If you want to specify other exit codes as success, write them here",
--- a/lib/task-data/tasks/catalog-driveid.js
+++ b/lib/task-data/tasks/catalog-driveid.js
@@ -11,7 +11,7 @@ module.exports = {
         commands: [
             {
                 command: 'sudo node get_driveid.js',
-                downloadUrl: '/api/current/templates/get_driveid.js'
+                downloadUrl: '{{ api.templates }}/get_driveid.js?nodeId={{ task.nodeId }}'
             }
         ]
     },
--- a/lib/task-data/tasks/catalog-smart.js
+++ b/lib/task-data/tasks/catalog-smart.js
@@ -11,7 +11,7 @@ module.exports = {
         commands: [
             {
                 command: 'sudo bash get_smart.sh',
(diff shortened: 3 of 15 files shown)


panpan0000:
:+0: , sounds functional good.

brianparry:
@lanchongyizu I think it would be best to have @anhou or @yyscamper work with you to merge this to make sure all the dependencies are covered.

anhou:
@lanchongyizu  if this PR had dependency with  PR  https://github.com/RackHD/on-http/pull/531, at least the depenency PR test should pass.  
and if it's possible to run regression test locally or in a separated branch, it's better

yyscamper:
@lanchongyizu: Forgot to mention you, this description should be changed since now the downloadUrl is full path not relative path: https://github.com/RackHD/on-tasks/blob/4cfd6a6efe141a83c27095a5eb6cadd7e5602c47/lib/task-data/schemas/linux-command.json#L44

lib/task-data/schemas/linux-command.json
```diff
@@ -43,7 +43,7 @@
                 "downloadUrl": {
                     "description": "Specify the download URL (relative to RackHD southbound API server) if the command needs to download some data or script from RackHD before execution",
                     "type": "string",
-                    "pattern": "^/[a-zA-Z0-9+-.#?=\\\/]+"
+                    "minLength": 1
```
cgx027: I would suggest to keep the ```pattern``` and check ```downloadurl``` against a valid http URL pattern. There are talks/PRs around changing relative URI to fullpath URL and this check will help people catch errors if a legacy relative URI is used.

Ref: https://github.com/RackHD/RackHD/issues/489#issuecomment-262570418

lib/task-data/tasks/catalog-smart.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
         commands: [
             {
                 command: 'sudo bash get_smart.sh',
-                downloadUrl: '/api/current/templates/get_smart.sh'
+                downloadUrl: '{{api.templates}}/get_smart.sh?nodeId={{context.target}}'
```
cgx027: This suggestion has gone beyond this PR, but ```{{ context.target }}``` gives people unclear hint that is actually referring to a node ID. something like ```?nodeId={{context.nodeId}}'``` or ```?nodeId={{ tasks.nodeID}}'``` would make the code more readable and less error-prong. 

Agreed? @yyscamper @anhou . If yes, this would generate a follow up refactor task.

cgx027: Just find the refactor has already been done in master, so please update {{context.target}} to {{nodeId}}

lib/task-data/tasks/catalog-driveid.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
         commands: [
             {
                 command: 'sudo node get_driveid.js',
-                downloadUrl: '/api/current/templates/get_driveid.js'
+                downloadUrl: '{{api.templates}}/get_driveid.js?nodeId={{context.target}}'
```
lanchongyizu: Please check with https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/install-centos.js#L14.
Thanks.

lib/task-data/tasks/catalog-smart.js
lanchongyizu: This doesn't work for task rendering, but only for templates and profiles rendering.
Thanks.

PR #376, opened 2016-12-01
Fix "TypeError: Cannot read property 'subscribe' of undefined"

sunnyqianzhang:
When cancelling tests, there is such error in on-taskgraph.log, which does no help when debugging, thus need remove. The root cause of this issue is that in on-taskgraph/lib/task-runner.js,
    TaskRunner.prototype.cancelTask = function(data) {
        var self = this;
        return Rx.Observable.just(data)
            .map(function(taskData) {
                return self.activeTasks[taskData.taskId];
            })
            .filter(function(task) { return !_.isEmpty(task); })
            .tap(function(task) {
                logger.info('Cancelling task', { taskId: task.instanceId });
            })
            .flatMap(function(task) { return task.cancel();})  ==> In original design, the task.cancel() is "undefined" instead of a promise, so the change is made to let task.cancel()return a promise
            .map(function() { return { taskId: data.taskId }; })
            .finally(function() {
                delete self.activeTasks[data.taskId];
            });
    };

### DIFF
--- a/lib/jobs/base-job.js
+++ b/lib/jobs/base-job.js
@@ -132,7 +132,7 @@ function baseJobFactory(
             loggerObject.error = error;
         }
         this.logger.info("Stopping job.", loggerObject);
-        this._done(error);
+        return this._done(error);
     };
 
     BaseJob.prototype._done = function _done(error) {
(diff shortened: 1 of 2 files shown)


sunnyqianzhang:
test this please

cgx027:
Good finding. :+1: 

Better if this PR can explain why the fix works.

yyscamper:
Good finding. Invite @benbp to take a look.

PR #377, opened 2016-12-16
Add "ignitionScriptUri" to install-coreos schema

tgelter:
This PR is paired with https://github.com/RackHD/on-http/pull/541 in order to support CoreOS Ignition templating support on bare metal via RackHD.

### DIFF
--- a/spec/lib/task-data/schemas/install-coreos-spec.js
+++ b/spec/lib/task-data/schemas/install-coreos-spec.js
@@ -41,8 +41,7 @@ describe(require('path').basename(__filename), function() {
         "version",
         "repo",
         "hostname",
-        "comport",
-        "ignitionScriptUri"
+        "comport"
     ];
 
     var SchemaUtHelper = require('./schema-ut-helper');
--- a/spec/lib/task-data/schemas/install-coreos-spec.js
+++ b/spec/lib/task-data/schemas/install-coreos-spec.js
@@ -16,7 +16,8 @@ describe(require('path').basename(__filename), function() {
         "installDisk": "/dev/sda",
         "comport": "ttyS0",
         "osType": "linux",
-        "rootPassword": "RackHD"
+        "rootPassword": "RackHD",
+        "ignitionScriptUri": "generic.json"
     };
 
     var positiveSetParam = {
--- a/spec/lib/task-data/schemas/install-coreos-spec.js
+++ b/spec/lib/task-data/schemas/install-coreos-spec.js
@@ -17,7 +17,7 @@ describe(require('path').basename(__filename), function() {
         "comport": "ttyS0",
         "osType": "linux",
         "rootPassword": "RackHD",
-        "ignitionScriptUri": "generic.json"
+        "ignitionScriptUri": "http://172.31.128.9090/api/current/templates/ignition.json"
     };
 
     var positiveSetParam = {
@@ -41,7 +41,8 @@ describe(require('path').basename(__filename), function() {
         "version",
         "repo",
         "hostname",
-        "comport"
+        "comport",
+        "ignitionScriptUri"
     ];
 
     var SchemaUtHelper = require('./schema-ut-helper');
--- a/lib/task-data/schemas/install-coreos.json
+++ b/lib/task-data/schemas/install-coreos.json
@@ -4,9 +4,21 @@
     "title": "Install CoreOS",
     "description": "The paramters schema for CoreOS installation",
     "describeJob": "Job.Os.Install",
+    "definitions": {
+        "CoreOsSpecificOptions": {
+            "type": "object",
+            "properties": {
+                "ignitionScriptUri": {
+                    "description": "The URI used to download the CoreOS Ignition configuration file",
+                    "type": "string",
+                    "minLength": 1
+            }
+        }
+    },
     "allOf": [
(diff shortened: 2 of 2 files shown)


brianparry:
ok to test

brianparry:
add to whitelist

spec/lib/task-data/schemas/install-coreos-spec.js
```diff
@@ -16,7 +16,8 @@ describe(require('path').basename(__filename), function() {
         "installDisk": "/dev/sda",
         "comport": "ttyS0",
         "osType": "linux",
-        "rootPassword": "RackHD"
+        "rootPassword": "RackHD",
+        "ignitionScriptUri": "generic.json"
```
yyscamper: where is `generic.json`? I don't see it in both on-tasks PR and on-http PR. I think you forgot to commit it.

tgelter: @yyscamper, I purposefully excluded any ignition script, not wanting to customize CoreOS unless someone overrides `ignitionScriptUri`.
As currently written, Ignition will silently fail, indicating in the log that `generic.json` could not be found. If it's preferable, I can put a dummy ignition script in that does something like creates a `README` or something similar.

anhou: if it's only for unit test, present 'generic.json' is fine, no need for a dummy script

anhou:
👍

yyscamper:
I just realized you just update the schema unit-testing for install-coreos, actually this is not enough. if the `ignitionScriptUri` is never specified, actually template rendering for `install-core.sh` will fail at this line (https://github.com/RackHD/on-http/pull/541/files#diff-76f31d740bfe833fb4ae241050bb6d99R11) due to cannot find the reference of `ignitionScriptUri`, so to download the install-core.sh will always fail before you mentioned silent fail.

To fix this, you need set a default value for `ignitionScriptUri` in task definition (https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/install-coreos.js), or you can change the `ignitionScriptUri` to be optional just like we handle the `rootSshKey` for ESXi, this is the code: https://github.com/RackHD/on-http/blob/master/data/templates/esx-ks#L63

What's more, whenever you add a new option for task, please don't forget to update the task schema (https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-coreos.json). Since the `ignitionScriptUri` option is only specific for CoreOS, so you can copy the Photon-OS schema to create some CoreOS specific option, this is the reference code: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-photon-os.json

tgelter:
Hello @yyscamper,
Thank you very much for the feedback.
I'll be back in the office on Jan 3 and will begin immediately implementing the requested changes.
Thanks,
Tim

tgelter:
Hello @yyscamper, I'm performing some final testing of the changes you requested, however I did notice that the unit testing coverage decreased as a result of my changes, which I'd assume would make these changes ineligible for merging in their current form.
However, since **ignitionScriptUri** is an optional parameter, I wasn't sure that it needed to exist in any section other than **canonical**. Please let me know if that's not the case.
Thanks,
Tim

yyscamper:
@tgelter: I believe the unit-test decrease is due to you didn't rebase your branch with latest upstream. The coverage is always compared with the upstream latest. So you need to sync your upstream and then rebase.

I also notice your PR contains some change that not belongs to you (those task progress code), to rebase the code can fix this as well.

tgelter:
Thank you again @yyscamper!
I've rebased and pushed. Please take another look and let me know if any other changes are needed.

spec/lib/task-data/schemas/install-coreos-spec.js
```diff
@@ -16,7 +16,8 @@ describe(require('path').basename(__filename), function() {
         "installDisk": "/dev/sda",
         "comport": "ttyS0",
         "osType": "linux",
-        "rootPassword": "RackHD"
+        "rootPassword": "RackHD",
+        "ignitionScriptUri": "http://172.31.128.9090/api/current/templates/ignition.json"
```
yyscamper: If `ignitionScriptUri` is an optional parameter, you can add it in `positiveUnsetParam`, then the unit-test will try to test whether it is correctly set as optional. If your schema is wrongly set it as required, then the unit-test will fail.

yyscamper:
jenkins: ok to test

spec/lib/task-data/schemas/install-coreos-spec.js
tgelter: Thank you for this information. Is this something you'd like to see updated prior to merge?

yyscamper: It's a minor suggestion for me, either updated it or not is OK. I have already approved your pull request. 
I will find an appropriate time to merge your pull request.

Thanks for your effort again.

yyscamper:
jenkins: test this please

brianparry:
test this please

tgelter:
@yyscamper / @brianparry, if there are issues with the code, let me know. From what I'm seeing, it looks like Jenkins is having issues -- something to do with a lack of compute nodes?

yyscamper:
jenkins: test this please

changev:
test this please

PR #379, opened 2017-01-10
Split existing events to internal events and external events

sunnyqianzhang:
This code change depend on code changes of on-core in the following PR:
https://github.com/RackHD/on-core/pull/239
@iceiilin @pengz1 @anhou @yyscamper


pengz1:
+1

lib/jobs/base-job.js
```diff
@@ -569,7 +569,7 @@ function baseJobFactory(
      */
     BaseJob.prototype._publishRunRedfishCommand =
         function _publishRunRedfishCommand (uuid, hostData) {
-        return messenger.publish(
+        return messenger.publishInternalEvents(
             Constants.Protocol.Exchanges.Task.Name,
```
pengz1: Those two messenger.publish should be move to on-core/lib/protocol/events, otherwise you need another change here to align data format. 
For this PR I am OK to keep as it is

iceiilin: i have a little doubt about this comment. event.js is used for external event, while the function here is used in job level to communicate among processes inside rackhd. keeping it here is clear in the code structure.

PR #389, opened 2017-01-20
fix and enhance task annotation due to new task schema change

yyscamper:
The new task annotation looks like:

image

All changes include:
1. Fix issue: getFullSchema doesn't handle the $ref for own definitions.
2. Use task injectableName as task annotation's primary key.
3. Sort task annotation via injectableName
4. Add original task definition.
5. Enhance option view

@anhou @iceiilin @pengz1 @brianparry @WangWinson

### DIFF
--- a/lib/task.js
+++ b/lib/task.js
@@ -583,15 +583,30 @@ function factory(
     };
 
     Task.getFullSchema = function getFullSchema(definition) {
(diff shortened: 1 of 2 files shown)


iceiilin:
:+1:

PR #390, opened 2017-01-22
bugfix: move node-uuid dep from devdep to dep

changev:
When a module is installed in other repos dependence , devDependencies packages of this module won't be installed:
* run ```npm install``` in on-http or on-taskgraph, and then can not find node-uuid  in  ```on-http/node_modules/on-tasks/node_modules/```. This caused on-http build failure. 

@iceiilin @panpan0000 @yyscamper

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "ssh2": "0.5.0",
     "url-parse": "~1.0.5",
     "xml2js": "^0.4.4",
-    "temp": "~0.8.3"
+    "temp": "~0.8.3",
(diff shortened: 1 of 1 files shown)


iceiilin:
:+1:

panpan0000:
👍 
maybe we can consider add ```npm run apidoc``` in PR Gate , to ensure this kind of error not missing.
background failure in  http://rackhdci.lss.emc.com/job/BuildRelease/job/Build/job/debian-build/186/console

PR #396, opened 2017-01-27
Added waitOn.either for logical workflow path

zhangy42:
@RackHD/corecommitters 
@pscharla 
@VulpesArtificem 

There are several highlights:

1. Task Request comes from:

https://rackhd.atlassian.net/browse/RAC-2043

2. Since current waitOn can already do:
"trigger task based on both conditions"

This feature is letting workflow engine do:
"trigger task based on one of these conditions ( ie: under waitOn.either) that is fulfilled"

3. There is also dependency clearing part in on-core:

https://github.com/RackHD/on-core/pull/253
This feature need to be enabled by merging these two PRs.

4. functional test is at:
https://github.com/RackHD/RackHD/pull/633


zhangy42:
test this please

zhangy42:
what is task.ucs.catalog?

lib/task-graph.js
```diff
@@ -59,13 +59,57 @@ function taskGraphFactory(
         };
         if (this.context.target) {
             this.logContext.id = this.context.target;
-        }
+       }
```
yyscamper: The previous indent was correct, but you changed it to wrong indent, please revert this change.

lib/task-graph.js
```diff
@@ -88,28 +132,22 @@ function taskGraphFactory(
...
-            // This won't actually get called because we do this same check in
-            // _populateTaskData, but keep this here just to be defensive.
-            if (!_.has(self.tasks, dep)) {
-                throw new Error('Graph does not contain task with ID ' + dep);
-            }
-            // Expand 'finished' to all finished task states
-            if (_.contains(value, Constants.Task.States.Finished)) {
-                value = Constants.Task.FinishedStates;
-                task.waitingOn[dep] = [Constants.Task.States.Finished];
+            if (dep === "either") {
```
yyscamper: The word `either` is a magic value or a keyword in workflow definition, We can either defining some indicator to distinguish it from task labels, such as changing `either` to `$either`, or we need add a validation logic in workflow definition to make sure the `either` is not been used as task label.

spec/lib/task-graph-spec.js
```diff
@@ -520,6 +520,14 @@ describe('Task Graph', function () {
                     .that.equals('finished');
             });
         });
+
+        it('should create task with waitOn.either', function(){
```
yyscamper: Can you create a test case to verify the `either` is worked as the "OR" logic?

lib/task-graph.js
```diff
@@ -149,11 +187,14 @@ function taskGraphFactory(
...
 
             _.forEach(_.keys(taskData.waitOn), function(waitOnTask) {
-                var newWaitOnTaskKey = idMap[waitOnTask];
-                assert.ok(newWaitOnTaskKey, 'Task to wait on does not exist: ' + waitOnTask);
-                var waitOnTaskValue = taskData.waitOn[waitOnTask];
-                delete taskData.waitOn[waitOnTask];
-                taskData.waitOn[newWaitOnTaskKey] = waitOnTaskValue;
+                if (waitOnTask === "either") {
+                    _.forEach(_.keys(taskData.waitOn[waitOnTask]), function(eitherTask){
+                      self._processOneWaitOnTask(taskData.waitOn[waitOnTask], idMap, eitherTask);
```
yyscamper: Our code convention is to use 4 spaces for indent

lib/task-graph.js
```diff
@@ -149,11 +187,14 @@ function taskGraphFactory(
             }
 
             _.forEach(_.keys(taskData.waitOn), function(waitOnTask) {
-                var newWaitOnTaskKey = idMap[waitOnTask];
-                assert.ok(newWaitOnTaskKey, 'Task to wait on does not exist: ' + waitOnTask);
-                var waitOnTaskValue = taskData.waitOn[waitOnTask];
-                delete taskData.waitOn[waitOnTask];
-                taskData.waitOn[newWaitOnTaskKey] = waitOnTaskValue;
+                if (waitOnTask === "either") {
+                    _.forEach(_.keys(taskData.waitOn[waitOnTask]), function(eitherTask){
```
yyscamper: Not a problem here, you can check the `_.forOwn`, it will be more simple if use at here.

spec/lib/task-graph-spec.js
zhangy42: Just a question, I actually have my own test at hand, but I am wondering, how can I merge/upload it to our existing test system, should I use FIT or CIT?

lib/task-graph.js
zhangy42: I agree with that we need to check every "reserved words". But when I asked around, I cannot find any logic validation feature in our codbase, do you know where should I start?

spec/lib/task-graph-spec.js
pscharla: @zhangy42 : I would use the FIT framework as that is path we are using going forward. Meaning CIT tests are now being merged into FIT.  Please contact @hohene if you have questions

zhangy42:
test this please

keedya:
test this please

lib/task-graph.js
yyscamper: The mongodb will not allow user to define variable name that starts with '$', so you could first choose to add the assertion logic, I think this is one of the place that is suitable for you mentioned code: https://github.com/RackHD/on-tasks/blob/master/lib/task-graph.js#L344

spec/lib/task-graph-spec.js
yyscamper: @zhangy42 : If we can do it in unit-test, that will be best. Otherwise, do in in FIT is also acceptable.

yyscamper:
@zhangy42 : you need first rebase your PR, those code about UCS catalog shouldn't belong to your PR.

yyscamper:
@zhangy42 : as for the keyword `either`, it means `one or the other of two:` (see http://www.dictionary.com/browse/either), but in our cases, we many define 3 or even more waitOn conditions. so the keyword need be changed.

From my experience on JSON-Schema, I think the word `anyOf` is suitable here (see https://spacetelescope.github.io/understanding-json-schema/reference/combining.html).

zhangy42:
Too many irrelevant changes and conflicts for rebase, making another banch and another PR.

PR #397, opened 2017-01-27
Add UCS catalog capabilities

tannoa2:


geoff-reid:
:+1: once Jenkins is passing.

PR #398, opened 2017-02-03
Catalog UCS rackmount server's children

tannoa2:
-Fix the URL for the Catalog.
-Catalog the server and its children instead of the server alone.

### DIFF
--- a/lib/jobs/ucs-catalog.js
+++ b/lib/jobs/ucs-catalog.js
@@ -80,19 +80,18 @@ function UcsCatalogJobFactory(
                 var username = self.ucs.settings.username;
                 var password = self.ucs.settings.password;
                 var ucs = self.ucs.settings.ucs;
-                var baseurl= self.ucs.settings.protocol+"://"+self.ucs.settings.host+ ":"+self.ucs.settings.port;/* jshint ignore:line */
-                var url= baseurl + "/catalog/?host=" + ucs+ "&user="+ username + "&password=" + password + "&identifier="+ node.name;/* jshint ignore:line */
+                var url=  "/catalog?host=" + ucs+ "&user="+ username + "&password=" + password + "&identifier="+ node.name;/* jshint ignore:line */
                 return self.ucs.clientRequest(url);
             })
             .then(function(res) {
                 assert.object(res);
                 return res.body;
             })
-            .then(function(data) {
+            .map(function(data) {
                 assert.object(data);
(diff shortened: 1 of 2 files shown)


tannoa2:
@RackHD/corecommitters @RackHD/veyron

lib/jobs/ucs-catalog.js
```diff
@@ -80,19 +80,18 @@ function UcsCatalogJobFactory(
...
-                var url= baseurl + "/catalog/?host=" + ucs+ "&user="+ username + "&password=" + password + "&identifier="+ node.name;/* jshint ignore:line */
+                var url=  "/catalog?host=" + ucs+ "&user="+ username + "&password=" + password + "&identifier="+ node.name;/* jshint ignore:line */
                 return self.ucs.clientRequest(url);
             })
             .then(function(res) {
                 assert.object(res);
                 return res.body;
             })
-            .then(function(data) {
+            .map(function(data) {
```
BillyAbildgaard: isn't the `.map` function only for arrays?  Looks like you then assert data should be an object so I'm not sure if that makes sense

tannoa2: Hi @BillyAbildgaard, correct. We changed the res.body content into an array from the UCS server side, that why we are using the .map

lib/jobs/ucs-catalog.js
```diff
@@ -80,19 +80,18 @@ function UcsCatalogJobFactory(
                 var username = self.ucs.settings.username;
                 var password = self.ucs.settings.password;
                 var ucs = self.ucs.settings.ucs;
-                var baseurl= self.ucs.settings.protocol+"://"+self.ucs.settings.host+ ":"+self.ucs.settings.port;/* jshint ignore:line */
-                var url= baseurl + "/catalog/?host=" + ucs+ "&user="+ username + "&password=" + password + "&identifier="+ node.name;/* jshint ignore:line */
+                var url=  "/catalog?host=" + ucs+ "&user="+ username + "&password=" + password + "&identifier="+ node.name;/* jshint ignore:line */
```
geoff-reid: Why mark this line as jshint ignore?

tannoa2: Too Long for hound's taste

tannoa2: Too Long for hound's taste

PR #403, opened 2017-02-14
Enable discovery of UCS Chassis and Servers

geoff-reid:
@RackHD/corecommitters @tannoa2 @keedya @uppalk1

### DIFF
--- a/lib/jobs/ucs-discovery.js
+++ b/lib/jobs/ucs-discovery.js
@@ -16,7 +16,8 @@ di.annotate(UcsDiscoveryJobFactory, new di.Inject(
     'Services.Waterline',
     'Services.Encryption',
     '_',
-    'JobUtils.UcsTool'
+    'JobUtils.UcsTool',
+    'Constants'
 ));
 
 function UcsDiscoveryJobFactory(
@@ -28,7 +29,8 @@ function UcsDiscoveryJobFactory(
     waterline,
     encryption,
     _,
-    UcsTool
+    UcsTool,
+    Constants
 ) {
     var logger = Logger.initialize(UcsDiscoveryJobFactory);
 
@@ -77,7 +79,7 @@ function UcsDiscoveryJobFactory(
         
         return self.getRoot()
             .then(function(root) {
-                return [ root, self.createRackmounts(root) ];
+                return [ root, self.createRackmounts(root), self.createServers(root) ];
             })
             .spread(function() {
                 self._done();
@@ -140,44 +142,7 @@ function UcsDiscoveryJobFactory(
                         return res.body;
                     })
                     .map(function(data) {
-                        assert.array(data.macs);
-                        assert.string(data.name);
-                        assert.string(data.path);
-
-                        var config = Object.assign({}, self.settings);
-                        obm = {
-                            config: config,
-                            service: 'ucs-obm-service'
-                        };
-                        var node = {
-                            type: 'compute',
-                            name: data.path,
-                            identifiers: data.macs,
-                            obm:[obm],
-                            relations : []
-                        };
-
-                        // Update existing node  or create a new one
-                        return waterline.nodes.needOne({identifiers:data.macs})
-                            .then(function(curNode) {
-                                return waterline.nodes.updateOne(
-                                    { id: curNode.id },
-                                    node
-                                )
-                                .then(function(data){
-                                    return data;
-                                });
-                            })
-                            .catch(function(error) {
-                                if (error.name === 'NotFoundError') {
-                                    return waterline.nodes.create(node);
-                                }
-                                throw error;
-                            })
-                            .then(function(data){
-                                return self.createObms(data.id,obm);
-                            });
-
+                        self.createNode(data, Constants.NodeTypes.Compute);
                     });
             }else{
                 logger.warning('No rackmount servers found Found');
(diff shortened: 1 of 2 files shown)


amymullins:
test this please

geoff-reid:
test this please

geoff-reid:
test this please

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +199,69 @@ function UcsDiscoveryJobFactory(
...
+        var id;
+        if (type === 'compute')
+        {
+            assert.array(data.macs);
+            id = data.macs;
+        } else {
+            id = [config.ucs, data.path];
+        }
+        assert.string(data.name);
+        assert.string(data.path);
```
yyscamper: These two lines are duplicated with line 220-221

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +199,69 @@ function UcsDiscoveryJobFactory(
...
+
+        var self = this;
+        var config = Object.assign({}, self.settings);
+
+        assert.string(data.name);
+        assert.string(data.path);
+
+        var id;
+        if (type === 'compute')
+        {
```
yyscamper: A little style problem, this mark should be put at the end of its last line.

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +199,69 @@ function UcsDiscoveryJobFactory(
...
+        var node = {
+            type: type,
+            name: data.path,
+            identifiers: id,
+            obm:[obm],
+            relations : []
+        };
+
+        // Update existing node  or create a new one
+        return waterline.nodes.needOne({identifiers:data.macs})
```
yyscamper: Please confirm whether these macs will always be stored in fixed order for the same ucs server. For example, if the server has mac1 and mac2, if first time it returns `[mac1, mac2]`, but the second time it is `[mac2, mac1]`. These two array are not identical but this should be considered to be the same server.

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +199,69 @@ function UcsDiscoveryJobFactory(
...
+            .then(function(curNode) {
+                return waterline.nodes.updateOne(
+                    { id: curNode.id },
+                    node
+                )
+                    .then(function(data){
```
yyscamper: indent error

yyscamper: And this `then` block is redundant, since it just returns what it receives, so please remove this then block.

lib/jobs/ucs-discovery.js
```diff
@@ -140,44 +140,57 @@ function UcsDiscoveryJobFactory(
...
+                return;
+            }
+        });
+    };
+
+    /**
+     * @function createServers
+     * @description discovers all the chassis servers in the Cisco UCS
+     */
+    UcsDiscoveryJob.prototype.createServers = function (root) {
```
yyscamper: Too many nesting levels for this function! it's hard to read, please split it into smaller functions to reduce the nesting levels.

lib/jobs/ucs-discovery.js
```diff
@@ -140,44 +140,57 @@ function UcsDiscoveryJobFactory(
...
+        var self = this;
+        var username = this.settings.username;
+        var password = this.settings.password;
+        var ucs = this.settings.ucs;
+        var baseurl = this.settings.protocol+"://"+this.settings.host+ ":"+this.settings.port;
+        var url= baseurl + "/chassis?host=" + ucs+ "&user="+ username + "&password=" + password;/* jshint ignore:line */
+        return Promise.try(function () {
+            if (_.has(root, 'Chassis')) {
```
yyscamper: If you change to below style, actually you can reduce on nesting level:
```
if (!_.has(root, 'Chassis')) {
  return;
}

return self.ucs.clientRequest(url)
...
```

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +199,69 @@ function UcsDiscoveryJobFactory(
         });
     };
 
+    UcsDiscoveryJob.prototype.updateRelations = function(nodeId, relations) {
```
yyscamper: The relations are always mutual relations, such as "encloses" vs. "enclosedBy". If if you change the node which has "encloses" relation, the its peer node need be updated as well.

lib/jobs/ucs-discovery.js
```diff
@@ -140,44 +140,57 @@ function UcsDiscoveryJobFactory(
...
+     * @function createServers
+     * @description discovers all the chassis servers in the Cisco UCS
+     */
+    UcsDiscoveryJob.prototype.createServers = function (root) {
+        var self = this;
+        var username = this.settings.username;
+        var password = this.settings.password;
+        var ucs = this.settings.ucs;
+        var baseurl = this.settings.protocol+"://"+this.settings.host+ ":"+this.settings.port;
+        var url= baseurl + "/chassis?host=" + ucs+ "&user="+ username + "&password=" + password;/* jshint ignore:line */
```
yyscamper: A little style suggestion. Somewhere there is space both before and after the operator `+` , sometimes only after, sometimes neither, you should at least keep it in same style.

lib/jobs/ucs-discovery.js
```diff
@@ -140,44 +140,57 @@ function UcsDiscoveryJobFactory(
...
+                    .map(function(chassisData) {
+                        return self.createNode(chassisData, 'enclosure')
+                            .spread(function (chassisNode) {
+                                var chassisId = chassisNode.id;
+                                var nodeList = [];
+                                return Promise.map(chassisData.members, function (data) {
+                                    return self.createNode(data, 'compute')
+                                    .spread(function(newNode) {
+                                        var relations = [{
+                                                         relationType: 'encloseBy',
```
yyscamper: a typo here `encloseBy` => `enclosuredBy`. see https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L273

lib/jobs/ucs-discovery.js
```diff
@@ -140,44 +140,57 @@ function UcsDiscoveryJobFactory(
...
+        var url= baseurl + "/chassis?host=" + ucs+ "&user="+ username + "&password=" + password;/* jshint ignore:line */
+        return Promise.try(function () {
+            if (_.has(root, 'Chassis')) {
+                return self.ucs.clientRequest(url)
+                    .then(function(res) {
+                        assert.object(res);
+                        return res.body;
+                    })
+                    .map(function(chassisData) {
+                        return self.createNode(chassisData, 'enclosure')
```
yyscamper: Please change node type to use the constant that defined at https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L328

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +199,67 @@ function UcsDiscoveryJobFactory(
...
+
+        var obm = {
+            config: config,
+            service: 'ucs-obm-service'
+        };
+        var node = {
+            type: type,
+            name: data.path,
+            identifiers: id,
+            obm:[obm],
```
brianparry: @geoff-reid This property should be removed.  The obms are added below and live their own collection.

lib/jobs/ucs-discovery.js
yyscamper: After examining the whole code, the cause that leads to tool meany nesting levels is that you mixed the code that creating node and updating node relations together, just split the node creation and node update to a smaller function, this function will look much more simple:

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +199,69 @@ function UcsDiscoveryJobFactory(
...
+                    });
+            })
+            .catch(function(error) {
+                if (error.name === 'NotFoundError') {
+                    return waterline.nodes.create(node);
+                }
+                throw error;
+            })
+            .then(function(data){
+                return [data, waterline.obms.upsertByNode(data.id,obm)];
```
yyscamper: This is not a bug since this function is private and only used by your job, but may not a good design.  This function will return `Promise`, but how should the caller know it should use the `spread` rather than the `then` to get the returned node document? The return value of `waterline.obms.upsertByNode` actually not be used anywhere, so it should be better hidden in this function itself. 

For this comment, either change this or not is OK for me.

spec/lib/jobs/ucs-discovery-spec.js
```diff
@@ -105,18 +110,46 @@ describe('Ucs Discovery Job', function () {
...
+            ucsTool.clientRequest.onCall(2).resolves(serverData);
             ucsJob._run()
             .then(function() {
                  expect(waterline.nodes.updateOne).to.be.called.once;
             });
         });
 
+        it('should not update a node', function() {
+            ucsTool.clientRequest.onCall(0).resolves([]);
+            ucsJob._run()
```
yyscamper: The promise must be returned! Otherwise this unit-test will always success.

lib/jobs/ucs-discovery.js
geoff-reid: This is done with multiple calls to updateRelations in the createServers routine.  The first call updates the compute node, the second call updates the chassis node.

lib/jobs/ucs-discovery.js
geoff-reid: I will make sure the ucs-service always returns the macs in the same order.

lib/jobs/ucs-discovery.js
geoff-reid: The node data is needed by the calling function, both promises need to be returned by the function.

lib/jobs/ucs-discovery.js
```diff
@@ -186,15 +151,114 @@ function UcsDiscoveryJobFactory(
...
+                );
+            })
+            .catch(function(error) {
+                if (error.name === 'NotFoundError') {
+                    return waterline.nodes.create(node);
+                }
+                throw error;
+            })
+            .then(function(data){
+                return [data, waterline.obms.upsertByNode(data.id,obm)];
```
brianparry: @geoff-reid It doesn't look like you ever use the fulfillment value from the obms upsert, so maybe you can just do this instead:
```
.tap(function(data) {
    return waterline.obms.upsertByNode(data.id,obm);
});
```
That will return a promise that fulfills to `data` when the upsert promise fulfills..

brianparry:
test this please

lib/jobs/ucs-discovery.js
yyscamper: @brianparry got my meaning.

geoff-reid:
test this please

geoff-reid:
test this please

PR #404, opened 2017-02-16
workflow level progress

lanchongyizu:
Implement workflow level progress indicator after task level progress.
https://rackhd.atlassian.net/browse/RAC-3923

@RackHD/corecommitters @pengz1 @iceiilin

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -1,4 +1,4 @@
-// Copyright 2015-2017, Dell EMC, Inc.
+// Copyright © 2015-2017 Dell Inc. or its subsidiaries.  All Rights Reserved.
 'use strict';
 
 var di = require('di');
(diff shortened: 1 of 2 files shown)


lib/jobs/install-os.js
```diff
@@ -245,10 +249,8 @@ function installOsJobFactory(
         }).then(function() {
             self._subscribeRequestProfile(function() {
                 var description = "Reboot suceeded, starting kernel download.";
-                return self.updateProgress(description, 1)
-                .then(function(){
-                    return self.profile;
-                });
+                self.updateProgress(description, 1)
+                return self.profile;
```
pengz1: Why remove promise chain here?

lanchongyizu: There is a special feature for publish task/graph progress event is that, it will swallow the error if the publishing itself has some error, so that the event publishing will not impact the existing logic.

PR #407, opened 2017-02-17
[skip ci] Refator OS installation progress

yyscamper:
1. Add the progress milestones definition into task options, now all milestones value are defined in single place, thus avoid the potentially divergence.
2. Prepare the milestone URIs which can be directly referenced by install scripts
3. Refine the milestones definition, add Pre/Post OS configuration milestones.

jenkins: depends on https://github.com/RackHD/on-http/pull/591

@pengz1 @lanchongyizu

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -50,11 +50,11 @@ function installOsJobFactory(
         self.nodeId = self.context.target;
         self.profile = self.options.profile;
         self.taskId = taskId;
-        self.options.totalSteps = self.options.totalSteps || 100;
 
         //OS repository analyze job may pass some options via shared context
         //The value from shared context will override the value in self options.
         self.options = _.assign(self.options, context.repoOptions);
+        insertProgressMilestones(self.options, taskId);
 
         this._validateOptions();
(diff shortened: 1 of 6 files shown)


lanchongyizu:
:+1:

lib/jobs/install-os.js
```diff
@@ -49,18 +49,49 @@ function installOsJobFactory(
...
+    }
+
+    function insertProgressMilestones(options, taskId) {
+        var defaultMilestones = {
+            requestProfile: { value: 1, description: 'Enter ipxe and request OS installation profile' }, //jshint ignore: line
+            enterProfile:   { value: 2, description: 'Enter profile, start to download installer'},
+            startInstaller: { value: 3, description: 'Start installer and prepare installation' },
+            preConfig:      { value: 4, description: 'Enter Pre OS configuration'},
+            postConfig:     { value: 5, description: 'Enter Post OS configuration'},
+            completed:      { value: 6, description: 'Finished OS installation'}
```
pengz1: We should document this for user since not all OSes have such 6 steps.

PR #411, opened 2017-02-23
 Update-provide-catalog-data-job-to-provide-full-catalog

pengz1:
* Secure erase graph needs full catalog of driveId source cached 
 * Provide catalog data job is extended to provide full catalog of a source.
 * A full path include "data." is required after this PR to get catalog from specified source.

An example of provide-quanta-bios-version.js, before this PR, task options are:
    options: {
        'source': 'ami',
        'path': 'systemRomId'
    }

With this PR, task options will be:
    options: {
        'source': 'ami',
        'path': 'data.systemRomId'
    }

### DIFF
--- a/lib/jobs/provide-catalog-data.js
+++ b/lib/jobs/provide-catalog-data.js
@@ -66,7 +66,7 @@ function ProvideCatalogDataJobFactory(
             if (!catalog) {
                 throw new Error("Could not find a catalog entry for " + self.options.source);
             }
-            var value = catalogSearch.getPath(catalog.data, self.options.path);
+            var value = catalogSearch.getPath(catalog, self.options.path);
             if (value === undefined) {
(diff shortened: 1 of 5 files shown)


pengz1:
@iceiilin @sunnyqianzhang @lanchongyizu

yyscamper:
@pengz1 : I just realized we still have another similar job which has implemented the same feature you are trying to get. 
 - Job: https://github.com/RackHD/on-tasks/blob/master/lib/jobs/get-catalog-values-job.js
 - BaseTask: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/base-tasks/get-catalog-values.js
 - Schema: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/get-catalog-values.json

Please evaluate whether it is enough for you.

PR #413, opened 2017-02-23
Fixing the data corruption in the nodes collection

tannoa2:


tannoa2:
@dalebremner @king-jam

PR #415, opened 2017-02-24
 Add node sleep task

pengz1:
* Node sleep task will set node to sleep for a duration

### DIFF
--- a/lib/task-data/tasks/node-sleep.js
+++ b/lib/task-data/tasks/node-sleep.js
@@ -1,0 +1,14 @@
+// Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Node sleep',
+    injectableName: 'Task.Node.Sleep',
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        duration: 10000, //in milliseconds
+        commands: ['sudo sleep $(awk "BEGIN{print {{options.duration}}/1000}")']
+    },
+    properties: {}
+};
--- a/spec/lib/task-data/tasks/node-sleep-spec.js
+++ b/spec/lib/task-data/tasks/node-sleep-spec.js
@@ -1,0 +1,17 @@
+// Copyright 2017, Dell EMC, Inc.
+/* jshint node:true */
+
+'use strict';
+
+describe(require('path').basename(__filename), function () {
+    var base = require('./base-tasks-spec');
+
+    base.before(function (context) {
+        context.taskdefinition = helper.require('/lib/task-data/tasks/node-sleep.js');
+    });
+
+    describe('task-data', function () {
+        base.examples();
+    });
+
+});


pengz1:
@iceiilin @sunnyqianzhang @lanchongyizu

lib/task-data/tasks/node-sleep.js
```diff
@@ -0,0 +1,17 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Node sleep',
+    injectableName: 'Task.Node.Sleep',
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        duration: 10, //in seconds
+        commands: ['sudo sleep {{options.duration}}']
```
yyscamper: Why use the micro kernel sleep? isn't same to delay in RackHD. The bluebird has function `Promise.delay` to do delay in async mode.

lib/task-data/tasks/node-sleep.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Node sleep',
+    injectableName: 'Task.Node.Sleep',
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        duration: 10, //in seconds
```
yyscamper: Usually we prefer the delay in milliseconds.

lib/task-data/tasks/node-sleep.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: Copyright date is not correct.

spec/lib/task-data/tasks/node-sleep-spec.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2015, EMC, Inc.
```
yyscamper: Incorrect copyright date

lib/task-data/tasks/node-sleep.js
pengz1: There is difference between RackHD delay and node delay. In the case a delay after reboot, RackHD delay starts from reboot finished while a node delay can only start after microkernel is ready.

lanchongyizu:
:+1:

lib/task-data/tasks/node-sleep.js
```diff
@@ -0,0 +1,14 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Node sleep',
+    injectableName: 'Task.Node.Sleep',
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        duration: 10000, //in milliseconds
+        commands: ['sudo sleep $(({{options.duration}}/1000))']
```
yyscamper: This doesn't support float number calculation, for float number, you could try:
```
>>> $(echo "scale=3; 500/110" | bc)
4.545
>>> $(awk 'BEGIN{print 500/110 }')
4.54545
```

sunnyqianzhang:
:+1:

lib/task-data/tasks/node-sleep.js
```diff
@@ -0,0 +1,14 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Node sleep',
+    injectableName: 'Task.Node.Sleep',
+    implementsTask: 'Task.Base.Linux.Commands',
+    options: {
+        duration: 10000, //in milliseconds
+        commands: ['sudo sleep $(awk "BEGIN{print {{options.duration}}/1000}")']
```
ytjohn: Not a big deal, but we don't need privilege escalation for sleep.

pengz1: @ytjohn 
This delay is configurable in secure, if you don't need delay, you can configure it to 0. You test reminds me we only have RackHD delay but don't have a real node delay, so I created this task that can also be used for other workflows.
Also after fixing secure erase job bug in:  https://github.com/RackHD/on-tasks/pull/416, You should be able achieve delay with noop task as you wanted.

PR #418, opened 2017-02-27
Refine workflow progress

lanchongyizu:
Refactor workflow progress implementation with progress service design.
1) move updateGraphProgress into Services.GraphProgress
2) remove default progressMilestones from install-os job since we can't get it to calculate progress when task started.

This PR is related with https://rackhd.atlassian.net/browse/RAC-4390

Depends on https://github.com/RackHD/on-core/pull/261 https://github.com/RackHD/on-taskgraph/pull/220 https://github.com/RackHD/on-http/pull/598

@RackHD/corecommitters @pengz1

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -16,8 +16,7 @@ di.annotate(installOsJobFactory, new di.Provide('Job.Os.Install'));
         'Promise',
         'JobUtils.CatalogSearchHelpers',
         'Services.Waterline',
-        'GraphProgress',
-        'Protocol.Events'
+        'Services.GraphProgress'
     )
 );
 
@@ -31,8 +30,7 @@ function installOsJobFactory(
     Promise,
     catalogSearch,
     waterline,
-    GraphProgress,
-    eventsProtocol
+    graphProgressService
 ) {
     var logger = Logger.initialize(installOsJobFactory);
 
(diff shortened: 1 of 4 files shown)


lib/jobs/install-os.js
```diff
@@ -258,49 +256,21 @@ function installOsJobFactory(
...
         var self = this;
         return Promise.resolve().then(function() {
             return self._convertInstallDisk();
         }).then(function() {
             self._subscribeRequestProfile(function() {
-                return self.updateGraphProgress(self.options.progressMilestones.requestProfile)
-                .then(function() {
-                    return self.profile;
-                });
+                if(_.has(self.options, 'progressMilestones.requestProfile')) {
```
yyscamper: This is not enough, if `requestProfile` is set to `null` or something else, the `_.has` will also return `true`.

lib/jobs/install-os.js
```diff
@@ -310,10 +280,14 @@ function installOsJobFactory(
...
-                    return self._done();
-                });
+                if(_.has(self.options, 'progressMilestones.completed')) {
+                    graphProgressService.publishTaskProgress(
+                        self.context.graphId,
+                        self.taskId,
+                        self.options.progressMilestones.completed
+                    );
+                }
+                return self._done();
```
yyscamper: Since `graphProgressService.publishTaskProgress` is a promise, so the `_done` is called before the publishTaskProgress finishes, this means the job is finished with some background code is still in execution.

Please change this to make sure the _done is called when everything is OK.

anhou:
@lanchongyizu Could you help explicitly link the dependency  PR where 'publishTaskProgress' is ?

lib/jobs/install-os.js
```diff
@@ -258,46 +256,24 @@ function installOsJobFactory(
...
      */
     InstallOsJob.prototype._run = function() {
         var self = this;
         return Promise.resolve().then(function() {
             return self._convertInstallDisk();
         }).then(function() {
             self._subscribeRequestProfile(function() {
-                return self.updateGraphProgress(self.options.progressMilestones.requestProfile)
+                return Promise.resolve()
+                .tap(function() {
```
anhou: 'then' is enough, don't need 'tap' here

lib/jobs/install-os.js
```diff
@@ -310,7 +286,17 @@ function installOsJobFactory(
             self._subscribeNodeNotification(self.nodeId, function(data) {
                 assert.object(data);
 
-                return self.updateGraphProgress(self.options.progressMilestones.completed)
+                return Promise.resolve()
+                .tap(function() {
```
anhou: 'then' is enough, don't need 'tap' here

lib/jobs/install-os.js
lanchongyizu: Since publishing progress should not affect the original flow, I think `tap` is better than `then`.
@yyscamper @pengz1 What do you think about that?

lib/jobs/install-os.js
lanchongyizu: Since publishing progress should not affect the original flow, I think `tap` is better than `then`.
@yyscamper @pengz1 What do you think about that?

yyscamper: Both `then` and `tap` are OK at here, but considering the event publish is just an action that "sneaked" into a normal procedure, so I'd prefer the `tap`

PengTian0:
test this please

PR #420, opened 2017-03-07
Add support for Cisco UCS Service Profile Discovery

uppalk1:
*Create base-task, task and job for a new workflow to
         discover UCS Service Profile

### DIFF
--- a/lib/jobs/ucs-service-profile-discovery.js
+++ b/lib/jobs/ucs-service-profile-discovery.js
@@ -1,0 +1,199 @@
+//Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+var di = require('di'),
+    urlParse = require('url-parse');
+
+module.exports = UcsServiceProfileDiscoveryJobFactory;
+di.annotate(UcsServiceProfileDiscoveryJobFactory,
+    new di.Provide('Job.Ucs.Service.Profile.Discovery'));
+di.annotate(UcsServiceProfileDiscoveryJobFactory, new di.Inject(
+    'Job.Base',
+    'Logger',
+    'Promise',
+    'Assert',
+    'Util',
+    'Services.Waterline',
+    'Services.Encryption',
+    '_',
+    'JobUtils.UcsTool',
+    'Constants'
+));
+
+function UcsServiceProfileDiscoveryJobFactory(
+    BaseJob,
+    Logger,
+    Promise,
+    assert,
+    util,
+    waterline,
+    encryption,
+    _,
+    UcsTool,
+    Constants
+) {
+    var logger = Logger.initialize(UcsServiceProfileDiscoveryJobFactory);
+
+    /**
+     * @param {Object} options task options object
+     * @param {Object} context graph context object
+     * @param {String} taskId running task identifier
+     * @constructor
+     */
+    function UcsServiceProfileDiscoveryJob(options, context, taskId) {
+        UcsServiceProfileDiscoveryJob.super_.call(this,
+                                   logger,
+                                   options,
+                                   context,
+                                   taskId);
+
+        assert.object(this.options);
+        assert.string(this.options.uri);
+        assert.string(this.options.username);
+        assert.string(this.options.password);
+        assert.string(this.options.ucs);
+        var parse = urlParse(this.options.uri);
+        var protocol = parse.protocol.replace(':','').trim();
+        this.settings = {
+            uri: parse.href,
+            host: parse.host.split(':')[0],
+            root: parse.pathname,
+            port: parse.port,
+            protocol: protocol,
+            username: this.options.username,
+            password: this.options.password,
+            ucs: this.options.ucs,
+            verifySSL: this.options.verifySSL || false
+        };
+        this.ucs = new UcsTool();
+        this.ucs.settings = this.settings;
+    }
+    
+    util.inherits(UcsServiceProfileDiscoveryJob, BaseJob);
+
+    /**
+     * @memberOf UcsServiceProfileDiscoveryJob
+     */
+    UcsServiceProfileDiscoveryJob.prototype._run = function() {
+        var self = this;
+        
+        return self.getRoot()
+            .then(function(root) {
+                return [ self.createServiceProfile(root) ];
+            })
+            .spread(function() {
+                self._done();
+            })
+            .catch(function(err) {
+                self._done(err);
+            });
+    };
+
+    /**
+     * @function getRoot
+     * @description return all the elements from the Cisco UCS Manager that RackHD cares about
+     */
+    UcsServiceProfileDiscoveryJob.prototype.getRoot = function () {
+        var username = this.settings.username;
+        var password = this.settings.password;
+        var ucs = this.settings.ucs;
+        var url=
+            "/serviceProfile?host=" + ucs+ "&user="+ username + "&password=" + password;
+        return this.ucs.clientRequest(url)
+            .then(function(response) {
+                return response.body;
+            });
+    };
+
+    /**
+     * @function createServiceProfile
+     * @description discovers all the service profiles in the Cisco UCS
+     */
+    UcsServiceProfileDiscoveryJob.prototype.createServiceProfile = function (root) {
+        var self = this;
+        var username = this.settings.username;
+        var password = this.settings.password;
+        var ucs = this.settings.ucs;
+        var baseurl = this.settings.protocol+"://"+this.settings.host+ ":"+this.settings.port;
+        var url=
+            baseurl + "/serviceProfile?host=" + ucs+ "&user="+ username + "&password=" + password;
+        return Promise.try(function () {
+            if (_.has(root, 'ServiceProfile')) {
+                return self.ucs.clientRequest(url)
+                    .then(function(res) {
+                        assert.object(res);
+                        return res.body;
+                    })
+                    .then(function(data) {
+                        assert.object(data);
+                        return data['ServiceProfile']['members'];
+                    })
+                    .map(function(members){
+                        return self.createUpdateNode(members, Constants.NodeTypes.Compute)
+                        .spread(function(newNode){
+                            var relations = [{
+                                relationType: 'associatedTo',
+                                targets: [newNode.associatedServer]
+                            }];
+                            return self.updateRelations(newNode.id, relations)
+                        });
+                    });
+            }else{
+                logger.warning('No Logical Servers found');
+                return;
+            }
+        });
+    };
+
+    UcsServiceProfileDiscoveryJob.prototype.updateRelations = function(nodeId, relations) {
+        // Update existing node with new relations or create one
+        return waterline.nodes.needOneById(nodeId)
+            .then(function(curNode) {
+                relations = _.uniq(relations.concat(curNode.relations), 'relationType');
+                return waterline.nodes.updateOne(
+                    { id: curNode.id },
+                    { relations: relations });
+            });
+    };
+
+    UcsServiceProfileDiscoveryJob.prototype.createUpdateNode = function(data, type) {
+
+        var self = this;
+        var config = Object.assign({}, self.settings);
+
+        assert.string(data.name);
+        assert.string(data.path);
+        var ids;
+        ids = [config.ucs, data.path];
+        var obm = {
+            config: config,
+            service: 'ucs-obm-service'
+        };
+        var node = {
+            type: type,
+            name: data.name,
+            identifiers: ids,
+            relations : []
+        };
+        // Update existing node  or create a new one
+        return waterline.nodes.needOne({identifiers:ids})
+            .then(function(curNode) {
+                return waterline.nodes.updateOne(
+                    { id: curNode.id },
+                    node
+                );
+            })
+            .catch(function(error) {
+                if (error.name === 'NotFoundError') {
+                    return waterline.nodes.create(node);
+                }
+                throw error;
+            })
+            .then(function(data){
+                return [data, waterline.obms.upsertByNode(data.id,obm)];
+            });
+    };
+
+    return UcsServiceProfileDiscoveryJob;
+}
--- a/lib/jobs/ucs-service-profile-discovery.js
+++ b/lib/jobs/ucs-service-profile-discovery.js
@@ -136,7 +136,7 @@ function UcsServiceProfileDiscoveryJobFactory(
                                 relationType: 'associatedTo',
                                 targets: [newNode.associatedServer]
                             }];
-                            return self.updateRelations(newNode.id, relations)
+                            return self.updateRelations(newNode.id, relations);
                         });
                     });
             }else{
--- a/lib/task-data/base-tasks/ucs-service-profile-discovery.js
+++ b/lib/task-data/base-tasks/ucs-service-profile-discovery.js
@@ -1,0 +1,17 @@
+//Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'UCS Service Profile Discovery Job',
+    injectableName: 'Task.Base.UCS.Service.Profile.Discovery',
+    runJob: 'Job.Ucs.Service.Profile.Discovery',
+    requiredOptions: [
+        'uri'
+    ],
+    requiredProperties: {
+    },
+    properties: {
+        catalog: {}
+    }
+};
--- a/lib/task-data/tasks/ucs-service-profile-discovery.js
+++ b/lib/task-data/tasks/ucs-service-profile-discovery.js
@@ -1,0 +1,17 @@
+//Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: "Ucs Service Profile Discovery",
+    injectableName: "Task.Ucs.Service.Profile.Discovery",
+    implementsTask: "Task.Base.UCS.Service.Profile.Discovery",
+    options: {
+        uri: null,
+        ucs: null,
+        username: null,
+        password: null,
+        verifySSL: false
+    },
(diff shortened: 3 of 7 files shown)


lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+
+    /**
+     * @memberOf UcsServiceProfileDiscoveryJob
+     */
+    UcsServiceProfileDiscoveryJob.prototype._run = function() {
+        var self = this;
+        
+        return self.getRoot()
+            .then(function(root) {
+                return [ root, self.createServiceProfile(root) ];
```
yyscamper: Since `root` is not used in following `spread` , so it better to be removed from the array.

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+        var protocol = parse.protocol.replace(':','').trim();
+        this.settings = {
+            uri: parse.href,
+            host: parse.host.split(':')[0],
+            root: parse.pathname,
+            port: parse.port,
+            protocol: protocol,
+            username: this.options.username,
+            password: this.options.password,
+            ucs: this.options.ucs,
```
yyscamper: Could you please teach me what's the difference between the setting of `ucs` and `host`? Because I see you use the `ucs` as the host in the query but rather than `host`
```JavaScript
var host = this.settings.ucs;
var url= "/login?host=" + host+ "&user="+ username + "&password=" + password;
```

My understanding is:
 - `host` is the ucs-service address, RackHD use the ucs-service to talk with UCS servers.
 - `ucs` is the target UCS server, the ucs-service proxies the request to UCS server.

Is correct?

lib/task-data/base-tasks/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,17 @@
...
+    runJob: 'Job.Ucs.Service.Profile.Discovery',
+    requiredOptions: [
+        'uri'
+    ],
+    requiredProperties: {
+    },
+    properties: {
+        catalog: {}
+    }
+};
```
yyscamper: Please add the schema for the new base task.

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+                    .then(function(res) {
+                        assert.object(res);
+                        return res.body;
+                    })
+                    .then(function(data) {
+                        assert.object(data);
+                        return data['ServiceProfile']['members']
+                    })
+                    .map(function(members){
+                        self.createNode(members, Constants.NodeTypes.Compute);
```
yyscamper: This promise has to be returned, otherwise this function will return earlier before all nodes are really created.

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+        var id;
+        id = [config.ucs, data.path];
+        var obm = {
+            config: config,
+            service: 'ucs-obm-service'
+        };
+        var node = {
+            type: type,
+            name: data.name,
+            associatedServer: data.associatedServer,
```
yyscamper: What's this property? I don't see it is defined in node model (https://github.com/RackHD/on-core/blob/master/lib/models/node.js). If it is not necessary, them remove it; if you think this is necessary, please explain the reason and then we can check whether it is suitable to create a new property.

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+
+    UcsServiceProfileDiscoveryJob.prototype.createNode = function(data, type) {
+
+        var self = this;
+        var config = Object.assign({}, self.settings);
+
+        assert.string(data.name);
+        assert.string(data.path);
+        var id;
+        id = [config.ucs, data.path];
```
yyscamper: Please use the plural of `id` or change to other better name.

Because when I see the code `identifer: id` in line 174, I thought the `identifer` need be array, but you wrongly gave it a single id.  So I look back to check the node, and see the `id` is really an array.

My point is choosing a good variable name and don't bring surprise to other people who read the code.

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+            associatedServer: data.associatedServer,
+            identifiers: id,
+            relations : []
+        };
+        // Update existing node  or create a new one
+        return waterline.nodes.needOne({identifiers:data.path})
+            .then(function(curNode) {
+                return waterline.nodes.updateOne(
+                    { id: curNode.id },
+                    node
```
yyscamper: In line 175, the relations is reset to be empty, since this line of code is to update the node document, so whether it is intentionally to reset the relations? And where the relations are built again?

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+                        self.createNode(members, Constants.NodeTypes.Compute);
+                    });
+            }else{
+                logger.warning('No Logical Servers found');
+                return;
+            }
+        });
+    };
+
+    UcsServiceProfileDiscoveryJob.prototype.createNode = function(data, type) {
```
yyscamper: The function name is not accurate, the code is to "update or create" node, not just "create" node.

lib/task-data/base-tasks/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2016, EMC, Inc.
```
yyscamper: Incorrect copyright date

lib/task-data/base-tasks/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,17 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'UCS Service Profile Discovery Job',
+    injectableName: 'Task.Base.UCS.Service.Profile.Discovery',
+    runJob: 'Job.Ucs.Service.Profile.Discovery',
+    requiredOptions: [
+        'uri'
```
yyscamper: From your implementation, the `username`, `password`, `ucs` are also required options.

```javascript
assert.string(this.options.uri);
assert.string(this.options.username);
assert.string(this.options.password);
assert.string(this.options.ucs);
```

lib/task-data/base-tasks/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,17 @@
...
+    ],
+    requiredProperties: {
+    },
+    properties: {
+        catalog: {}
```
yyscamper: What's this property used for? I don't see any reason to define such property and I don't see any code is to create the catalog.

lib/task-data/tasks/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2016, EMC, Inc.
```
yyscamper: Incorrect copyright date

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+            host: parse.host.split(':')[0],
+            root: parse.pathname,
+            port: parse.port,
+            protocol: protocol,
+            username: this.options.username,
+            password: this.options.password,
+            ucs: this.options.ucs,
+            verifySSL: this.options.verifySSL || false
+        };
+        this.ucs = new UcsTool();
```
yyscamper: `this.ucs` and `this.settings.ucs` is a little confused here,  please rename to another distinguish one?

spec/lib/jobs/ucs-service-profile-discovery-spec.js
```diff
@@ -0,0 +1,146 @@
...
+    });
+    
+    beforeEach(function() {
+        var Job = helper.injector.get('Job.Ucs.Service.Profile.Discovery');
+        ucsJob = new Job({
+            uri:'https://1.1.1.1:7080/sys',
+            username:'user',
+            password:'pass',
+            ucs:"1.1.1.1"
+        }, {}, graphId);
```
yyscamper: this is `taskId`, not `graphId`, though this doesn't break the unit-test, but don't bring confusion since unit-test is a kind of doc for the code.

spec/lib/jobs/ucs-service-profile-discovery-spec.js
```diff
@@ -0,0 +1,146 @@
...
+        
+        sandbox.stub(ucsJob.ucs);
```
yyscamper: The correct usage is `sandbox.stub(ucsJob, 'ucs')`

spec/lib/jobs/ucs-service-profile-discovery-spec.js
```diff
@@ -0,0 +1,146 @@
...
+            //needOneById: sandbox.stub().resolves(node)
+        };
+        waterline.obms = {
+            upsertByNode: sandbox.stub().resolves(obm)
+        };
+        Error = helper.injector.get('Errors');
+    });
+    
+    afterEach(function() {
+        sandbox.restore();
```
yyscamper: All stubs are created once in `before`, but all are restored after the first test case finishes. So when the 2nd test case begins to run, actually all stubs are not existing, so what's the purpose to create so many stubs in advance?

Meanwhile, I see a lot stubs are never been used, if so, just remove them.

spec/lib/jobs/ucs-service-profile-discovery-spec.js
```diff
@@ -0,0 +1,146 @@
...
+                });
+        });
+
+
+        it('should successfully log in', function() {
+            ucsTool.clientRequest.onCall(0).resolves({"body":"cookie"});
+            return ucsJob.logIn()
+                .then(function(data) {
+                    expect(data).to.deep.equal("cookie");
+                });
```
yyscamper: Actually, line 140-143 can be short for:
```
return ucsJob.logIn().to.become('cookie');
```

Your code is still OK, just for your awareness if you like it.

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,197 @@
...
+        };
+        var node = {
+            type: type,
+            name: data.name,
+            associatedServer: data.associatedServer,
+            identifiers: id,
+            relations : []
+        };
+        // Update existing node  or create a new one
+        return waterline.nodes.needOne({identifiers:data.path})
```
yyscamper: I am thinking of a corner case, since the identifiers is `[config.ucs, data.path]`, here you only find node document by `data.path`, but how about if the ucs server address is actually changed for the target node, then it means this node's obm service need be updated as well.

spec/lib/jobs/ucs-service-profile-discovery-spec.js
```diff
@@ -0,0 +1,146 @@
...
+        });
+    });
+});
```
yyscamper: I don't think the test coverage is good enough, I see a lot logic is actually not tested, for example:
(1) If ucs request error, then the job should fail.
(2) If input any invalid option, the job should fail.
(3) If ucs request returns multiple servers, then multiple nodes should be updated or created.
(4) The obm service should be created (or updated) when the node is created (or updated)

spec/lib/jobs/ucs-service-profile-discovery-spec.js
```diff
@@ -0,0 +1,146 @@
...
+        rootData,
+        serverData,
+        waterline = {},
+        Error;
+
+
+    var obm = {
+        config: {
+            uri: "http://10.240.19.220:7080/sys",
+            host: "10.240.19.220",
```
anhou: please use a faked IP

lib/jobs/ucs-service-profile-discovery.js
uppalk1: @yyscamper that is correct.

lib/task-data/base-tasks/ucs-service-profile-discovery.js
uppalk1: @yyscamper This is a place holder for follow on stories to create a catalog for UCS logical server (if we deem its necessary)

lib/task-data/base-tasks/ucs-service-profile-discovery.js
uppalk1: @yyscamper this will be refactored as part of https://rackhd.atlassian.net/browse/RAC-4389

lib/task-data/base-tasks/ucs-service-profile-discovery.js
uppalk1: @yyscamper can you point me to one that I could use as a reference ?

lib/jobs/ucs-service-profile-discovery.js
uppalk1: @yyscamper I don't understand your confusion. I'm creating an instance and assigning settings to it.

lib/jobs/ucs-service-profile-discovery.js
uppalk1: @yyscamper I created a new relation called "associatedTo" instead of creating a new property

lib/jobs/ucs-service-profile-discovery.js
uppalk1: @yyscamper Yes, the relation will be rebuilt with the discovery.

lib/jobs/ucs-service-profile-discovery.js
yyscamper: With the same variable name `ucs`, but when it is in `this.settings` it means an IP address, however it is a UCS tool instance if it is in `this`

lib/task-data/base-tasks/ucs-service-profile-discovery.js
yyscamper: This is pretty simple, just put a JSON schema in the property `optionsSchema`, this is very alike with the JSON schema in swagger ui.

This is an example: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/base-tasks/provide-catalog-data.js

lib/jobs/ucs-service-profile-discovery.js
yyscamper: This is good.

lib/jobs/ucs-service-profile-discovery.js
yyscamper: OK, thanks for explanation, could you please add some comment here

lib/task-data/base-tasks/ucs-service-profile-discovery.js
yyscamper: Sound good.

geoff-reid:
It look like the coveralls decreased coverage is due to a change in the way node-coveralls calculates the coverage percentage.  See https://github.com/nickmerwin/node-coveralls/issues/158

The actual coverage as reported by istanbul has increased.  

from master:  https://travis-ci.org/RackHD/on-tasks/jobs/206278092
```
=============================== Coverage summary ===============================
Statements   : 89.1% ( 6352/7129 )
Branches     : 73.79% ( 1681/2278 )
Functions    : 86.14% ( 1578/1832 )
Lines        : 89.15% ( 6343/7115 )
================================================================================
```
from this PR:  https://travis-ci.org/RackHD/on-tasks/jobs/209488312
```
=============================== Coverage summary ===============================
Statements   : 89.19% ( 6422/7200 )
Branches     : 73.82% ( 1686/2284 )
Functions    : 86.29% ( 1598/1852 )
Lines        : 89.24% ( 6413/7186 )
================================================================================
```

lib/task-data/base-tasks/ucs-service-profile-discovery.js
yyscamper: @uppalk1 : I am OK with your plan.

amymullins: @yyscamper if ok with the plan, can you please help remove the -1 that currently blocks the PR from being merged?

geoff-reid:
test this please

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -0,0 +1,199 @@
...
+            root: parse.pathname,
+            port: parse.port,
+            protocol: protocol,
+            username: this.options.username,
+            password: this.options.password,
+            ucs: this.options.ucs,
+            verifySSL: this.options.verifySSL || false
+        };
+        this.ucs = new UcsTool();
+        this.ucs.settings = this.settings;
```
yyscamper: It's even more confusing between `this.settings.ucs` and `this.ucs.settings`

spec/lib/jobs/ucs-service-profile-discovery-spec.js
```diff
@@ -0,0 +1,202 @@
...
+        waterline.nodes.updateOne.reset();
+    });
+    
+    describe('usc service profile discovery', function() {
+        it('should successfully run the job', function() {
+            ucsTool.clientRequest.onCall(0).resolves(rootData);
+            ucsTool.clientRequest.onCall(1).resolves(rootData);
+            return ucsJob._run()
+                .then(function() {
+                     expect(waterline.nodes.updateOne).to.be.called.twice;
```
yyscamper: There is no `called.twice` or `called.once`, you should use `calledTwice` and `calledOnce`.

The `twice` or `once` actually has no assertion effect, actually you could append any arbitrary words after `called`. For example:
`expect(xxxx).to.be.called.notexisting` or `expect(xxx).to.be.called.foobar` will always make the mocha happy. so the mocha success doesn't reflect such error.

PR #427, opened 2017-03-16
Optionally update an existing catalog

dalebremner:


lib/jobs/linux-catalog.js
```diff
@@ -46,6 +46,7 @@ function catalogJobFactory(
         this.commandUtil = new CommandUtil(this.nodeId);
         this.commands = this.commandUtil.buildCommands(options.commands);
         assert.arrayOfObject(this.commands);
```
yyscamper: there is linux-command job schema in lib/task-data/schemas, please add the new option there

lib/utils/job-utils/command-util.js
```diff
@@ -96,11 +96,22 @@ function commandUtilFactory(
...
             } else if (result.store) {
-                return waterline.catalogs.create({
+                var dataToSave = {
                     node: self.nodeId,
                     source: result.source || 'unknown',
                     data: result.data
-                });
+                };
+                if (self.updateExisting) {
+                    return waterline.catalogs.update(
```
yyscamper: I'm not sure, could you please confirm if the catalog doesn't exit, the update will become “create”

yyscamper: Or whether this is your design purpose: if the `updateExisting` is true but the catalog is not existing, then do nothing?

yyscamper: I'm not sure on your purpose of this design. If target on eliminating the out-of-date catalog, then fall back to create new catalog if the update fail will make more sense.

dalebremner: @yyscamper Good idea. I'll change the code to create a new catalog if it doesn't exist...

lib/jobs/linux-catalog.js
dalebremner: Thanks @yyscamper I added the property to the schema.

lib/utils/job-utils/command-util.js
```diff
@@ -96,11 +96,27 @@ function commandUtilFactory(
...
+                };
+                if (self.updateExistingCatalog) {
+                    var query = {
+                        node: self.nodeId,
+                        source: result.source
+                    };
+                    return waterline.catalogs.count(query)
+                        .then(function(count) {
+                            if (count) {
+                                return waterline.catalogs.update(query, dataToSave);
```
yyscamper: For our current design, there may be multiple catalog data from the same source because we don't remove the out-of-date data. For your choosed `update` command, all the out-of-date data (with same source) with be updated to  the exact same, if so, seems you should remove all duplicated ones, because it's useless to keep those identical data.

PR #429, opened 2017-03-20
WIP: Add UCS obm service:Fix Unit Tests

tannoa2:


tannoa2:
test this please

tannoa2:
test this please

tannoa2:
test this please

lib/services/ucs-obm-service.js
```diff
@@ -0,0 +1,113 @@
+// Copyright 2016, EMC, Inc.
```
yyscamper: Incorrect copyright date

lib/services/ucs-obm-service.js
```diff
@@ -0,0 +1,113 @@
...
+    util = require('util');
+
+module.exports = UcsObmServiceFactory;
+
+di.annotate(UcsObmServiceFactory, 
+    new di.Provide('ucs-obm-service')
+);
+di.annotate(UcsObmServiceFactory, 
+    new di.Inject(
+    'OBM.base',
```
yyscamper: Add indent

lib/jobs/ucs-discovery.js
```diff
@@ -210,6 +210,7 @@ function UcsDiscoveryJobFactory(
             id = [config.ucs, data.path];
         }
 
+        config.dn = data.path;
```
yyscamper: Please change a better variable name, `dn` is too short and never tell anything.

lib/services/ucs-obm-service.js
```diff
@@ -0,0 +1,113 @@
...
+    UcsObmService.prototype._runInternal = function (action) {
+        return this.run({
+            action:action
+        });
+    };
+    
+    UcsObmService.prototype.run = function (options) {       
+        var self = this;
+        assert.object(options);
+        assert.string(options.action);
```
yyscamper: The function `run` is designed to return a Promise either success or fail, but these assertion logic break the design. when assertion fails, it throws error rather than return a rejected Promise.

Usually, the good practice in our code is wrapping the assertion in promise.try:

```
return Promise.try(function() {
    assertion.object(option);
    assertion.string(options.action);
})
.then(function() {
    return waterline.nodes.findOne(....)
});
```

yyscamper: You may copy this code from our existing code, but I want to say the existing code isn't always a good example

lib/services/ucs-obm-service.js
```diff
@@ -0,0 +1,113 @@
...
+        var action = options.action;
+        return waterline.nodes.findOne(this.options.nodeId)
```
yyscamper: Since you need `nodeId` here, then why no assertion for it?

lib/services/ucs-obm-service.js
```diff
@@ -0,0 +1,113 @@
...
+
+'use strict';
+
+var di = require('di'),
+    util = require('util');
+
+module.exports = UcsObmServiceFactory;
+
+di.annotate(UcsObmServiceFactory, 
+    new di.Provide('ucs-obm-service')
```
yyscamper: Add the new obm service into schema https://github.com/RackHD/on-tasks/blob/master/lib/task-data/base-tasks/obm.js#L30, otherwise this obm will not pass the basic validation.

tannoa2:
test this please

lib/services/ucs-obm-service.js
```diff
@@ -0,0 +1,116 @@
...
+        });
+    };
+    
+    UcsObmService.prototype.run = function (options) {       
+        var self = this;
+        return Promise.try(function(){
+            assert.object(options);
+            assert.string(options.action);
+            var action = options.action;
+            var nodeId = this.options.nodeId;
```
yyscamper: Use `self` rather than `this`. The anonymous function will has its own `this` context, it is not the same with `self`

PR #430, opened 2017-03-21
Ssh options

VulpesArtificem:
related to https://github.com/RackHD/on-taskgraph/pull/222

@RackHD/corecommitters  @pscharla  

We needed to make the options for the underlying ssh.exec method accessible specifically to ensure that the chef install workflow can function when a pseudo-tty is required

### DIFF
--- a/lib/jobs/ssh-job.js
+++ b/lib/jobs/ssh-job.js
@@ -37,6 +37,7 @@ function sshJobFactory(
         this.nodeId = this.context.target;
         this.commandUtil = new CommandUtil(this.nodeId);
         this.commands = this.commandUtil.buildCommands(options.commands);
+        this.sshExecOptions = options.sshExecOptions;
         assert.arrayOfObject(this.commands);
     }
     util.inherits(SshJob, BaseJob);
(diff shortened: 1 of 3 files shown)


lib/jobs/ssh-job.js
```diff
@@ -37,6 +37,7 @@ function sshJobFactory(
         this.nodeId = this.context.target;
         this.commandUtil = new CommandUtil(this.nodeId);
         this.commands = this.commandUtil.buildCommands(options.commands);
+        this.sshExecOptions = options.sshExecOptions;
```
yyscamper: @brianparry 's team has added a "ibms" (InBand Management, just like obms) document, the ssh settings will be stored as a kind of ibm. So the ssh setting should be fetched from node's ibms settings.

Add @brianparry @BillyAbildgaard @dalebremner for comment.

yyscamper: Sorry, I misunderstand it, this is not ssh setting, but some arguments for ssh execution. please ignore my above comment.

spec/lib/utils/job-utils/command-util-spec.js
```diff
@@ -348,6 +348,40 @@ describe('Command Util', function() {
             return expect(commandUtil.sshExec(testCmd, sshSettings, mockClient)).to
                 .be.rejected;
         });
+
+        it('should pass sshExex options to the underlying method', function() {
```
yyscamper: typo: sshExex

lacarb:
+1

PR #431, opened 2017-03-23
fix: TaskCancellationError not caught

lanchongyizu:
FIX https://rackhd.atlassian.net/browse/RAC-4391
The TaskCancellationError should not be thrown, which will break the cancelTaskStream.

@RackHD/corecommitters  @iceiilin @pengz1 @sunnyqianzhang

### DIFF
--- a/lib/task.js
+++ b/lib/task.js
@@ -384,22 +384,30 @@ function factory(
     };
 
     Task.prototype.cancel = function(error) {
+        var self = this;
         error = error || new Errors.TaskCancellationError();
-        this.error = error;
+        self.error = error;
 
-        if (_.contains(Constants.Task.FinishedStates, this.state)) {
-            return;
-        } else if (this.state === TaskStates.Pending) {
-            // This block handles cancellation before run() has
-            // been called (i.e. the task has been created, but not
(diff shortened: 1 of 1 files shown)


lib/task.js
```diff
@@ -384,22 +384,34 @@ function factory(
...
+        if (self.state === TaskStates.Running) {
             // Task cancellation is passed down to the job first
             // and then bubbled up into the promise chain
             // created in this.run()
-            return this.job.cancel(error);
+            self.state = TaskStates.Cancelled;
+            return self.job
+                .cancel(error)
+                .catch(function(err) {
+                    logger.error('Swallow the Error when task cancelled', {
```
yyscamper: This error is expected, since user cancel the job purposely, so use `logger.error` doesn't make sense.

yyscamper:
Please also refine the PR title.

lib/task.js
```diff
@@ -384,22 +384,30 @@ function factory(
     };
 
     Task.prototype.cancel = function(error) {
+        var self = this;
         error = error || new Errors.TaskCancellationError();
-        this.error = error;
+        self.error = error;
```
pengz1: self is actually not necessary, but it is fine here.

PR #434, opened 2017-03-28
fixed issue with duplicate ucs node identifiers

geoff-reid:
Change node identifiers to be unique between nodes

@tannoa2 @uppalk1 @keedya @RackHD/corecommitters

### DIFF
--- a/lib/jobs/ucs-service-profile-discovery.js
+++ b/lib/jobs/ucs-service-profile-discovery.js
@@ -155,8 +155,8 @@ function UcsServiceProfileDiscoveryJobFactory(
 
         assert.string(data.name);
         assert.string(data.path);
-        var ids;
-        ids = [config.ucs, data.path];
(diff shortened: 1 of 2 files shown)


uppalk1:
@geoff-reid are you planning to update the unit tests too, if it implies ?

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -155,20 +155,20 @@ function UcsServiceProfileDiscoveryJobFactory(
...
+
         var obm = {
             config: config,
             service: 'ucs-obm-service'
         };
         var node = {
             type: type,
             name: data.name,
-            identifiers: ids,
+            identifiers: id,
```
yyscamper: The `identifiers` is an array, not string (see https://github.com/RackHD/on-core/blob/master/lib/models/node.js#L34), please change it back.

PR #438, opened 2017-04-20
remove "$ref" in build-in task schema to avoid conflict with mongodb

iceiilin:
Fix https://rackhd.atlassian.net/browse/RAC-4643 about conflict of "$ref" in task schema and mongodb keyword. The solution is to remove this keyword from build in task schema, which will be written to mongodb. Further complete solution is still under consideration.

### DIFF
--- a/lib/task-data/base-tasks/analyze-os-repo.js
+++ b/lib/task-data/base-tasks/analyze-os-repo.js
@@ -9,10 +9,14 @@ module.exports = {
     optionsSchema: {
         properties: {
             version: {
-                '$ref': 'types-installos.json#/definitions/Version'
+                "description": "The version of target OS",
+                "type": "string",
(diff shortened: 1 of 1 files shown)


yyscamper:
jenkins: test this please

iceiilin:
jenkins: test this please

iceiilin:
@nortonluo

yyscamper:
Add some comment for reviewer:

This is just a hot fix for the mongodb issue, to temporary avoid to insert keys that begins with `$` into mongo.

But to support the `$ref` actually will bring a lot convenience for future schema design, we have some ideas on this, but they are not perfect, either there is performance lost, or hard to implement.

Lyne and me will still work on this. But since the mongo issue is critical right now, so I am acceptable to merge this temporary fix, and leave us more time to find a better solution.

PR #442, opened 2017-05-01
update ucs discover jobs

geoff-reid:
Modify ucs discovery josbs to allow discovery and cataloging of ucs
nodes in the same graph

@uppalk1 @keedya @tannoa2 @RackHD/corecommitters

### DIFF
--- a/lib/jobs/ucs-catalog.js
+++ b/lib/jobs/ucs-catalog.js
@@ -44,7 +44,7 @@ function UcsCatalogJobFactory(
             options,
             context,
             taskId);
-        this.nodes = (this.context.physicalNodeList || []).concat(this.context.logicalNodeList);
+        this.nodes = _.union( this.context.physicalNodeList, this.context.logicalNodeList);
         if(_.isEmpty(this.nodes)) {
             this.nodes = [ this.context.target ];
         }
--- a/lib/jobs/ucs-catalog.js
+++ b/lib/jobs/ucs-catalog.js
@@ -44,7 +44,10 @@ function UcsCatalogJobFactory(
             options,
             context,
             taskId);
-        this.node = this.context.target;
+        this.nodes = (this.context.physicalNodeList || []).concat(this.context.logicalNodeList);
+        if(_.isEmpty(this.nodes)) {
+            this.nodes = [ this.context.target ];
+        }
         this.ucs = new UcsTool();
     }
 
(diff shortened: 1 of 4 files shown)


geoff-reid:
test this please

lib/jobs/ucs-catalog.js
```diff
@@ -44,7 +44,10 @@ function UcsCatalogJobFactory(
             options,
             context,
             taskId);
-        this.node = this.context.target;
+        this.nodes = (this.context.physicalNodeList || []).concat(this.context.logicalNodeList);
```
yyscamper: From your implementation, some nodes are extracted from node relations, so there may be some duplicated nodes in `this.nodes`. You need to remove those duplicated nodes.

lib/jobs/ucs-discovery.js
```diff
@@ -79,9 +79,19 @@ function UcsDiscoveryJobFactory(
...
-            .spread(function() {
+            .spread(function(rackmounts, servers) {
+                self.context.physicalNodeList = (self.context.physicalNodeList || []).concat(
+                    _.map(rackmounts, function(it) {
+                        return _.get(it, 'id');
+                    })
+                );
+                self.context.physicalNodeList = _.flattenDeep(self.context.physicalNodeList.concat(
+                    _.map(servers, function(it) {
+                        return [_.get(it, 'id'), _.get(it.relations[0], 'targets') ];
```
yyscamper: There are some hidden rules in your design:
- There is at least 1 relation
- The relation type happens to be the "enclosure" relation
- The newly created "enclosure" relation is happen to be the first relation.

If the node relation is empty, then you will get a "empty" node in final `self.context.physicalNodeList`, but you forgot to remove the empty node.

If the node relation is not "enclosure" relation or the "enclosure" relation is not at the first position, that means you are trying to handle wrong node, this will make  the ucs catalog fail.

These kinds of hidden rules make the code error-prone, especially somebody tries to refactor the code that generating the ucs server node but he isn't aware such kind of hidden rules. So my suggestion is:
- Filter the relation target by relation type, not use the fixed position.
- Always check whether the relation is empty or not.

lib/jobs/ucs-service-profile-discovery.js
```diff
@@ -80,9 +80,14 @@ function UcsServiceProfileDiscoveryJobFactory(
         
         return self.getRoot()
             .then(function(root) {
-                return [ self.createServiceProfile(root) ];
+                return self.createServiceProfile(root);
             })
-            .spread(function() {
+            .then(function(logicalServers) {
+                self.context.logicalNodeList = (self.context.logicalNodeList || []).concat(
+                    _.map(logicalServers, function(it) {
```
brianparry: @geoff-reid You can use the `_.property` iteratee shorthand for this: `_.map(logicalServers, 'id')`.

PR #443, opened 2017-05-02
Add option to allow console redirect in ESXi bootstrap

nortonluo:
In many rack mount servers or customized servers, there is no monitor connected, UART port are usually used as console. VMware ESXi use VGA card as console. User need pass VM kernel To support this use, I add three options: "gdbPort","logPort","debugLogToSerial". 
[VMware ESXi reference](https://pubs.vmware.com/vsphere-50/index.jsp?topic=%2Fcom.vmware.vsphere.install.doc_50%2FGUID-C65306C0-DA37-4F45-8A50-31F8D109BB1D.html])

To support this , 'on-http/data/templates/esx-boot-cfg', 'on-http/data/templates/esx-ks'(This PR) and 'on-tasks/lib/task-data/tasks/install-esx.js'(in another PR for on-tasks) are changed.
The default value of those options keep console in VGA as it does before. User can customize the console redirect setting in the OS bootstrapping payload.
Jenkins depends on https://github.com/RackHD/on-http/pull/636

### DIFF
--- a/lib/task-data/schemas/install-esxi.json
+++ b/lib/task-data/schemas/install-esxi.json
@@ -57,7 +57,7 @@
             ]
         },
         "gdbPort":{
-            "description":"GDB debug serial port mode",
+            "description":"Serial port used for GDB debug",
             "enum":[
                  "com1",
                  "com2",
--- a/lib/task-data/schemas/install-esxi.json
+++ b/lib/task-data/schemas/install-esxi.json
@@ -55,6 +55,32 @@
                 "0x3e8",
                 "0x2e8"
             ]
+        },
+        "gdbPort":{
+            "description":"GDB debug serial port mode",
+            "enum":[
+                 "com1",
+                 "com2",
+                 "none",
+                 "default"        
+            ]
+        },  
+        "logPort":{
+             "description":"Serial port for log output in ESXi",
+             "enum": [ 
+                   "com1",
+                   "com2",
+                   "none",
+                   "default"
+             ]  
+         },
+         "debugLogToSerial":{
+             "description":"0 = Serial debug off; 1 = Serial debug on;2 = defer to config option file",
+             "enum": [
+                     "0",
+                     "1",
+                     "2"
+             ]  
         }
     },
     "properties": {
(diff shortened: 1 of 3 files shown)


nortonluo:
@pengz1 @iceiilin @anhou @yyscamper

lib/task-data/tasks/install-esx.js
```diff
@@ -17,6 +17,9 @@ module.exports = {
         esxBootConfigTemplateUri: '{{ api.templates }}/{{ options.esxBootConfigTemplate }}?nodeId={{ task.nodeId }}', //jshint ignore: line
         comport: 'com1',
         comportaddress: '0x3f8', //com1=0x3f8, com2=0x2f8, com3=0x3e8, com4=0x2e8
+        gdbPort: 'default',
+        logPort: 'com1',
+        debugLogToSerial: '1',
```
yyscamper: This should be changed to `boolean`

yyscamper:
@nortonluo : Please also update the schema file: https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-esxi.json and its unit-test: https://github.com/RackHD/on-tasks/blob/master/spec/lib/task-data/schemas/install-esxi-spec.js

iceiilin:
Will these default settings misbehave in other HW platforms when installing ESXi?

lib/task-data/tasks/install-esx.js
```diff
@@ -17,6 +17,9 @@ module.exports = {
         esxBootConfigTemplateUri: '{{ api.templates }}/{{ options.esxBootConfigTemplate }}?nodeId={{ task.nodeId }}', //jshint ignore: line
         comport: 'com1',
         comportaddress: '0x3f8', //com1=0x3f8, com2=0x2f8, com3=0x3e8, com4=0x2e8
+        gdbPort: 'default',
+        logPort: 'com1',
```
nortonluo: comport is the device name of UART in ESXi. comportaddress is the physical address of UART. logport is the port that ESXi throw out debug log during boot/running. There are different. I think they couldn't get merged

lib/task-data/tasks/install-esx.js
nortonluo: This variable is a VMkernel value. It has value 0,1,2 so far.

nortonluo:
@iceiilin Those parameters are kept in ESXi default value by default. That is the same as the value before this PR. I Just test this (change parameters or not) on Rinjin, it is OK.

iceiilin:
Please also update schema file here (not only schema unit test) https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/install-esxi.json

lib/task-data/schemas/install-esxi.json
```diff
@@ -55,6 +55,32 @@
                 "0x3e8",
                 "0x2e8"
             ]
+        },
+        "gdbPort":{
+            "description":"GDB debug serial port mode",
```
pengz1: "mode" is unecessary

pengz1:
Please make sure to update document in:
http://rackhd.readthedocs.io/en/latest/rackhd/install_os.html

PR #449, opened 2017-05-10
on-tasks code

longhornteam35:
on-tasks code additions. Includes tasks and jobs for create volume,
delete volume, and add hotspare.

### DIFF
--- a/lib/jobs/add-volume.js
+++ b/lib/jobs/add-volume.js
@@ -1,0 +1,123 @@
+'use strict'
+
+var di = require('di');
+module.exports = AddVolumeJobFactory;
+di.annotate(AddVolumeJobFactory, new di.Provide('Job.Add.Volume'));
+di.annotate(AddVolumeJobFactory, new di.Inject(
+    'Job.Base',
+    'Logger',
+    'Promise',
+    'Assert',
+    'Util',
+    'Services.Waterline',
+    'Services.Lookup',
+    'Services.Configuration',
+    '_',
+    'HttpTool',
+    'Errors',
+    'JobUtils.WorkflowTool',
+    'Protocol.Events',
+    'validator',
+    'JobUtils.RedfishTool',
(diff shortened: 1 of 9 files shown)


lib/jobs/add-volume.js
```diff
@@ -0,0 +1,123 @@
...
+                //logger.info(results);
+                logger.info("create vd job is done");
+                self._done();
+            })
+            .catch(function(err) {
+                self._done(err);
+            }); 
+
+     	self._done();
+     }
```
rahmanmuhamad: Please have these houndCI comments fixed . some are important

lib/jobs/delete-volume.js
```diff
@@ -0,0 +1,109 @@
+'use strict'
```
rahmanmuhamad: we need to fix this.

rahmanmuhamad:
Plz have these comments and suggestion fixed,

iceiilin:
As the three jobs looks quite similar, is there possible to use only one job and distinguish their logic in JobUtils.RacadmTool? Or maybe we could leverage an existing racadm contorl base task (https://github.com/RackHD/on-tasks/blob/master/lib/task-data/base-tasks/racadm-control-base.js), and only implement corresponding tasks, just like https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/dell-racadm-update-firmware.js? Their actual logic can be added here, just like https://github.com/RackHD/on-tasks/blob/master/lib/utils/job-utils/racadm-tool.js#L222

dalebremner:
It would be good to have unit tests

rahmanmuhamad:
Don't have access to the repo to make these changes, i will do it after merge

michaelsteven:
Agree with the plan.  We do not have write access to the Longhorn 35 organization and the semester is over.  It is unknown if or when we would gain access.

PR #450, opened 2017-05-11
 Decouple secure erase with Megaraid operation

pengz1:
To decouple secure erase with Megaraid operation when possible, below changes are made:
 * Non-RAID disks won't extend drive catalogs 
 * getDriveIdCatalogExt is updated with flag as if we should extend drive  catalog for JBOD disks
 * secure erase job is update to get drive protocol from exiting driveId catalogs

### DIFF
--- a/lib/utils/job-utils/catalog-searcher.js
+++ b/lib/utils/job-utils/catalog-searcher.js
@@ -53,16 +53,14 @@ function searchCatalogDataFactory(
     }
 
     /**
-     * Get extended driveId catalog data
+     * Get driveId catalog data
      * @param {String} nodeId - node identifier
      * @param {Object} filter - [optional] The filter which contains the driveId catalogs identifier
-     *                          Or devName. For example: {'sda': 1, 'sdb': 1, '3': 1}.
+     *                          Or devName. For example: ['sda', 'sdb', '1'].
      *                          driveId catalogs in filter will return, otherwise skip.
-     * @return {Promise} driveId catalogs extended
+     * @return {Promise} driveId catalogs
      */
-    function getDriveIdCatalogExt(nodeId, filter) {
-        var virtualDiskData, controllerVendors = [];
-
+    function getDriveIdCatalog(nodeId, filter) {
         return waterline.catalogs.findMostRecent({
             node: nodeId,
             source: 'driveId'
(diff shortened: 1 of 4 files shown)


pengz1:
@iceiilin @anhou @lanchongyizu @sunnyqianzhang

lib/utils/job-utils/catalog-searcher.js
```diff
@@ -71,70 +69,138 @@ function searchCatalogDataFactory(
...
-                var controllerDataList = _.get(controllerCatalog, 'data.Controllers');
-                _.forEach(controllerDataList, function(data){
-                    vendor = _.get(data, '[Response Data][Scheduled Tasks].OEMID');
-                    controllerId = _.get(data, '[Command Status][Controller]');
-                    controllerVendors[controllerId] = vendor.toLowerCase();
+
+            return Promise.all([
+                foundVdHasValue ? getVirtualDiskCatalog(nodeId) : Promise.resolve(),
+                extendJbod ? getPhysicalDiskCatalog(nodeId) : Promise.resolve(),
+                getRaidControllerVendor(nodeId)
```
iceiilin: you could add "_" in the beginning of the above three function names to signify they are used internally.

lib/utils/job-utils/catalog-searcher.js
```diff
@@ -71,70 +69,138 @@ function searchCatalogDataFactory(
...
+    }
+
+    /**
+     * Get extended driveId catalog
+     * @param {String} nodeId - node identifier
+     * @param {Object} driveIds - driveId list retrieved from driveId catalogs
+     * @param {Boolean} extendJbod - flag for if we should extend JBOD physical information
+     * @return {Promise} Drive catalogs extended with Megaraid information
+     */
+    function getDriveIdCatalogExt(nodeId, filter, extendJbod) {
```
iceiilin: It seems you add a parameter "extendJbod" here, but secure erase job calls this function without the newly-added one. Is it left for further extension?

pengz1: You are right, I am struggling if we should extend disks for JBODs. For secure erase actually we don't need it. But we can leave it as it is in case we will use it in future.

iceiilin:
test this please

iceiilin:
test this please

PR #452, opened 2017-05-22
Documented the options for changing the iDrac network setting

tannoa2:


tannoa2:
@RackHD/veyron

lib/task-data/schemas/dell-racadm-control.json
```diff
@@ -18,6 +18,18 @@
             "description": "Specify if reboot after racadm control tasks are done",
             "type": "boolean"
         },
+        "ip": {
```
srinia6: @tannoa2  You might want to reference these from properties.

srinia6: NVM, got it!

pengz1: Same question: why those schema is not added to properties? And support you should only need one.

lib/task-data/schemas/dell-racadm-control.json
```diff
@@ -18,6 +18,18 @@
...
+            "description": "The ip address of the new iDrac  network setting",
+            "type": "string"
```
brianparry: @tannoa2 You should add `format ipv4`.

lib/task-data/schemas/dell-racadm-control.json
pengz1: Schema in a file is usually defined in "definitions" and then used in "properties". If items are not specified in "properties" they won't take effect.
I saw you defined those items for IP, netmask, gateway, however I didn't see you use those definitions.

pengz1: You should add some unit test cases to verify it.

anhou:
@tannoa2  is this PR still valid?

PR #453, opened 2017-05-25
modify search item for chassis serial number in generate enclosure job

iceiilin:
Previously the items searched in node's catalogs to determine the chassis serial number turned out to be inconsistent in all HW platforms. From ipmi fru spec, "chassis serial" in built-in FRU is the recommended field, while at least Dell C6320 doesn't follow this rule. Learning from BIOS engineer that chassis info in dmi is another place that might contain this info if manufacture flushes it to BIOS flash or board resume. 

This PR modifies these two items which will be consistent. If any of them have valid value, RackHD will use it as the chassis serial number.

### DIFF
--- a/lib/jobs/generate-enclosure.js
+++ b/lib/jobs/generate-enclosure.js
@@ -47,11 +47,11 @@ function generateEnclosureJobFactory(
             snPath: [
                 {
                     src: 'ipmi-fru',
-                    entry: 'Builtin FRU Device (ID 0).Product Serial'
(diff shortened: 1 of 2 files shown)


iceiilin:
test this please

pengz1:
A little confused by your description, so what is the status of C6320 "chassis serial"?

iceiilin:
There is no chassis serial in ipmi built in fru in C6320

yyscamper:
😄

PR #455, opened 2017-06-02
keep date consistency when generate enclosure

changev:
jenkins: depend on https://github.com/RackHD/on-core/pull/280

Modified ````generate-enclosure.js``` to use new DAL method to keep date consistency .

```_addRelation``` is duplicated with the same function in ```on-http/lib/services/node-api-service.js```
So the unit test is omitted here which causes ```Coverage decreased (-0.3%) to 72.391%```

### DIFF
--- a/lib/jobs/generate-enclosure.js
+++ b/lib/jobs/generate-enclosure.js
@@ -68,90 +68,41 @@ function generateEnclosureJobFactory(
         var self = this;
         var enclNode;
 
-        Promise.all([
-            Promise.any([
-                self._findSerialNumber(self.enclConst.snPath[0]),
-                self._findSerialNumber(self.enclConst.snPath[1])
-            ]),
-            waterline.nodes.find({ type: self.enclConst.type })
+        Promise.any([
+            self._findSerialNumber(self.enclConst.snPath[0]),
+            self._findSerialNumber(self.enclConst.snPath[1])
         ])
-        .spread(self._matchEnclosure.bind(self))
-        .then(function (matchInfo) {
-            // Create an enclosure node if there isn't matched one
-            enclNode = matchInfo.encl;
-
-            if (_.isEmpty(enclNode)) {
-
-                var enclData = {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by changev** [comment]
test this please

File: lib/jobs/generate-enclosure.js [hunk 1]
```diff
@@ -191,62 +142,75 @@ function generateEnclosureJobFactory(
...
-    GenerateEnclosureJob.prototype._popEnclTarget = function (node) {
-        var self = this;
-        var enclRelation = _.find(node.relations, function(entry) {
-            return entry.relationType === self.enclConst.relationEncl ||
-                entry.relationType === self.enclConst.relationEnclBy;
+            return targetsToBeAdded;
+        })
+        .then(function(targetsToBeAdded){
+            return waterline.nodes.addListItemsIfNotExistByIdentifier(
+                 node.id, targetsToBeAdded
```
iceiilin: without signs, is there any possibility that enclosure node will "encloses" one compute node more than once, if running this job multiple times?

lib/jobs/generate-enclosure.js
```diff
@@ -191,62 +142,75 @@ function generateEnclosureJobFactory(
...
+            var targetsToBeAdded = {};
+            targetsToBeAdded[field] = _.uniq(targetsItems);
 
-            if (enclSn === '') {
-                // No serial number in enclosure name, jump to the next entry
-                return true;
+            // Can not make sure prevent every exception in high concurrency.
+            if (type === 'containedBy' && 
+                modifiedNode.relations[index].targets.length + targets.length > 1) {
+                throw new Error("Node "+node.id+" can only be contained by one node");
```
changev: Two same functions may be more easily to maintain?

lib/jobs/generate-enclosure.js
changev: Mongo list update operator won't add two same items of value type to one list.

changev:
test this please

changev:
test this please

changev:
test this please

PR #456, opened 2017-06-06
Increase ssh handshake timeout to 1 minute

keedya:
- Default is set to 20 seconds
- This timeout can be added in the future to the workflow


keedya:
test this please

keedya:
test this please

keedya:
@RackHD/veyron

PR #458, opened 2017-06-08
 Add Wait.Sel.Events task.

pengz1:
Add Wait.Sel.Events task related jobs, base-task, schema and tasks and unit tests.
Jenkins depends on: https://github.com/RackHD/on-core/pull/283

### DIFF
--- a/lib/jobs/wait-sel-events-job.js
+++ b/lib/jobs/wait-sel-events-job.js
@@ -34,36 +34,35 @@ function waitSelEventJobFactory(
      * @param {String} taskId
      * @constructor
      */
-    function waitSelEventJob(options, context, taskId) {
-        waitSelEventJob.super_.call(this, logger, options, context, taskId);
-        assert.arrayOfObject(options.alertFilters, 'alertFilters should an array of object');
-        this.options = options;
+    function WaitSelEventJob(options, context, taskId) {
+        WaitSelEventJob.super_.call(this, logger, options, context, taskId);
         this.nodeId = context.target;
         this.pollerId = undefined;
         this.alertFilters = options.alertFilters;
         this.pollInterval = options.pollInterval || 60000;
         this.legacyInterval = undefined;
         this.freshAlerts = [];
+        this.severity =  options.severity || 'information';
     }
-    util.inherits(waitSelEventJob, BaseJob);
+    util.inherits(WaitSelEventJob, BaseJob);
 
     /**
-     * @memberOf waitSelEventJob
+     * @memberOf WaitSelEventJob
      */
-    waitSelEventJob.prototype._run = function run() {
+    WaitSelEventJob.prototype._run = function run() {
         var self = this;
         return waterline.workitems.findPollers({
             node: self.nodeId,
             'config.command': 'selEntries'
         })
         .tap(function(pollers){
-            self._retrievePollerInfo(pollers);
+            return self._retrievePollerInfo(pollers);
         })
         .then(function(pollers){
             return self.setSelAlertConfig(pollers[0]);
         })
         .then(function(){
-            return self._subscribeSelEvent(self.pollerId, self.nodeId, self._callback.bind(self));
+            return self._subscribeSelEvent(self.severity, self.pollerId, self._callback.bind(self));
         })
         .catch(function(err){
             logger.error('Fail to run wait sel events', {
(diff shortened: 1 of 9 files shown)


lib/task-data/schemas/wait-sel-events.json
```diff
@@ -0,0 +1,37 @@
+{
+    "copyright": "Copyright 2017, Dell EMC, Inc.",
+    "definitions": {
+        "alertFilter": {
+            "type": "object",
+            "properties": {
+                "count": {
```
yyscamper: From the unit-test, there are still other properties which are not described in schema, such as `sensorNumber`, `action`, .etc

lib/task-data/base-tasks/wait-sel-events.js
```diff
@@ -0,0 +1,12 @@
+// Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Wait for SEL events',
+    injectableName: 'Task.Base.Wait.Sel.Events',
+    runJob: 'Job.Wait.Sel.Events',
+    requiredOptions: ['alertFilters'],
```
yyscamper: the schema file should be specified in the baseTask using key `optionsSchema`

yyscamper: Meanwhile, if the optionsSchema is specified, the `requiredOptions` could be omitted.

lib/jobs/wait-sel-events-job.js
```diff
@@ -0,0 +1,213 @@
...
+        var query = {_id: waterline.workitems.mongo.objectId(this.pollerId)};
+        return waterline.workitems.updateMongo(query, update, {});
+    };
+
+    /**
+     * @memberOf waitSelEventJob
+     */
+    waitSelEventJob.prototype.unsetSelAlertConfig = function() {
+        var self = this;
+        var update = {
```
yyscamper: Previous there is a convention, the job should avoid to send raw database command, these commands should be moved into the on-core's waterline model.

lib/jobs/wait-sel-events-job.js
```diff
@@ -0,0 +1,213 @@
...
+     * @memberOf waitSelEventJob
+     */
+    waitSelEventJob.prototype._run = function run() {
+        var self = this;
+        return waterline.workitems.findPollers({
+            node: self.nodeId,
+            'config.command': 'selEntries'
+        })
+        .tap(function(pollers){
+            self._retrievePollerInfo(pollers);
```
yyscamper: this promise isn't returned! Firstly the timing is a problem, then if this promise is rejected, the `catch` cannot catch the error and process will crash.

lib/task-data/schemas/wait-sel-events.json
pengz1: I can't enumerate all, only count is what I defined.

lib/task-data/base-tasks/wait-sel-events.js
pengz1: Good catch

lib/jobs/wait-sel-events-job.js
pengz1: Make sense.

lib/task-data/schemas/wait-sel-events.json
iceiilin: I think other items in alert config can be added here to prevent user's error input as soon as user posting the workflow. Otherwise, the FW update (or other case users would like to use this task) will be executed for hours before the error format is detected.

lib/jobs/wait-sel-events-job.js
```diff
@@ -0,0 +1,209 @@
...
+        }
+        return Promise.try(function(){
+            poller = pollers[0];
+            self.pollerId = poller.id;
+            self.legacyInterval = poller.pollInterval;
+            existingAlerts = poller.config.alerts || [];
+            _.forEach(self.alertFilters, function(alertFilter){
+                var filter = _.omit(alertFilter, 'count');
+                if (!filter.action || filter.action !== 'critical'){
+                    filter.action = 'critical'; //filter.action will always be critical
```
iceiilin: Can this action be "information", and this job only listening to "severity: information" in the sel alert queue? I can't think of a better solution except for changing user's alert config in skupack, but reducing the severity to "information" might reduce the impact to the users. 

Meanwhile, the interface "subscribeSelEvent" in https://github.com/RackHD/on-core/pull/281/files can be more flexible to let the upper layer decide which severity level it would like to pay attention to.

lib/task-data/schemas/wait-sel-events.json
pengz1: I prefer @nortonluo 's idea to set a timeout for this task instead.
I can't enumerate what kind of alert it is required, it might be platform/generation dependent...

pengz1:
test this please

iceiilin:
travis is expected to fail due to dependency of PR in on-core

PR #459, opened 2017-06-09
 Fix bugs in RACADM getBiosConfig 

pengz1:
Two changes:
1. Fix bug for getSoftwareList which is now a standard catalog data structure, use getSoftwareList.data as data source instead of getSoftwareList for getBiosConfig method.
2. Add support for NFS share folder.


pengz1:
@iceiilin @anhou @nortonluo @sunnyqianzhang @lanchongyizu

PR #460, opened 2017-06-12
 Add force PXE boot task and job

pengz1:
Some platforms have different reset type and behaviors, we need enhanced PXE boot to force system into to pxe boot without 60 seconds timeout and without clearing reset flags with some kinds of reset.
force-pxe-boot tasks will implement those two parameters when doing pxe boot.

### DIFF
--- a/lib/services/ipmi-obm-service.js
+++ b/lib/services/ipmi-obm-service.js
@@ -60,6 +60,15 @@ function ipmiObmServiceFactory(BaseObmService, Promise) {
     IpmiObmService.prototype.setBootPxe = function() {
         return this._runInternal(['chassis', 'bootdev', 'pxe']);
     };
+    
+    /*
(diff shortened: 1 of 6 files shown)


pengz1:
@iceiilin @anhou @nortonluo @sunnyqianzhang @lanchongyizu

lib/services/ipmi-obm-service.js
```diff
@@ -61,6 +61,12 @@ function ipmiObmServiceFactory(BaseObmService, Promise) {
         return this._runInternal(['chassis', 'bootdev', 'pxe']);
     };
 
+    IpmiObmService.prototype.forceBootPxe = function() {
```
anhou: @pengz1  I'm not sure what's the better function name here, but could you help add some annotations about the difference with 'setBootPxe' so that the developers/users wouldn't be confused when reviewing this parts of codes in future

PR #461, opened 2017-06-15
WSMan Poller support for wsman discovered nodes

phelpdh:
- Implement wsman pollers
- Fix wsman-job where pollers were all polling the same target.
- Clean up formatting of wsman-job.js
- Fix wsman obm service functionality


lib/task-data/base-tasks/create-default-pollers.js
```diff
@@ -7,8 +7,7 @@ module.exports = {
     injectableName: 'Task.Base.Pollers.CreateDefault',
     runJob: 'Job.Pollers.CreateDefault',
     requiredOptions: [
-        'pollers',
-        'nodeId'
```
michaelsteven: Add a comment explaining why this is removed for this poller.

phelpdh: When calling wsman-post-discovery taskgraph via workflow-tool, the node id is passed in as the first arg to the runGraph() function.  For some reason, the passed in node id is not found by the create-default-pollers task.  I've also tried passing node id in an options object, which also fails.

lib/jobs/dell-wsman-discovery.js
```diff
@@ -660,20 +660,20 @@ function DellWsmanDiscoveryJobFactory(
...
                                                 self.addCatalog(node_, device.summary, 'DeviceSummary'),
                                                 self.addCatalog(node_, dmiCatalog, 'dmi'),
                                                 self.doPublishDiscovered(node_)])
                             .then(function(){
                                 self.createEnclosure(node_, device.ipAddress)
                                 .then(function(encNode){
-                                    return self.createRedfishObm(encNode, device.ipAddress, 1)
-                                    .then(function(){
+                                    //return self.createRedfishObm(encNode, device.ipAddress, 1)
+                                    //.then(function(){
```
MichaelRegert: Can we just delete these lines rather than checking in commented-out code?

lib/jobs/wsman-job.js
```diff
@@ -51,169 +51,156 @@ function wsmanJobFactory(
...
-    
-    
-                  
-     return WsmanJob;
+     * Client Request API
+     * 
+     * 
+     */
+
+    WsmanJob.prototype.clientRequest = function (host, path, method, data) {
```
rahmanmuhamad: How about removing this function and remove it with WsmanTool ? I know it was my bad to copy the function here

lib/jobs/dell-wsman-discovery.js
```diff
@@ -726,9 +718,6 @@ function DellWsmanDiscoveryJobFactory(
                         case "SWITCH":
                             self.processSwitch(data);
                         break;
-                        case "VM":
```
phelpdh: Removed this case since it is not implemented.

lib/jobs/wsman-job.js
```diff
@@ -2,24 +2,19 @@
...
-    'Assert',
     'Promise',
-    '_',
     'Services.Waterline',
-    'JobUtils.WsmanTool',       
-    'Constants',
+    'JobUtils.WsmanTool',
     'Services.Configuration',
     'Services.Encryption',
-    'HttpTool',
```
phelpdh: Removed unused modules.

PR #468, opened 2017-06-29
Fix obm service name to match the entry in Mongo DB (wsman-obm-service)

uppalk1:
and Fix obm username setting


geoff-reid:
Can you fix the hound error?

PR #469, opened 2017-07-03
Promise object not defined when create obm

lyanchih:
When created obm by sku, exception will been raised with promise is not defined at Promise.Try.

Because of the module do not inject Promise object. We just need to add Promise object into injection list.

### DIFF
--- a/lib/jobs/create-obm-settings.js
+++ b/lib/jobs/create-obm-settings.js
@@ -11,6 +11,7 @@ di.annotate(createObmSettingsJobFactory, new di.Inject(
     'Services.Waterline',
     'Task.Services.OBM',
     'Logger',
+    'Promise',
     'Assert',
     'Util',
     di.Injector
@@ -20,6 +21,7 @@ function createObmSettingsJobFactory(
     waterline,
     ObmService,
     Logger,
+    Promise,
     assert,
     util,
     injector


pengz1:
+ 1

yyscamper:
@lyanchih : if not use the injected Promise, then it falls back to use the ES6 native promise, ES6 native promise has no method "try" defined, that's the reason why you see the error. RackHD injected promise is "Bluebird Promise", it defines a lot helpful functions including `try`.

@pengz1 @iceiilin @anhou : I see two problems from this bug:
(1) Ideally the unittest should catch such simple error. But the reason why the unit-test cannot catch is because while doing unit-test, the global `Promise` is replaced by the BlueBird one (https://github.com/RackHD/on-core/blob/master/spec/helper.js#L16). So to keep the testing env and product env as same as possible will help to identify bug earlier.

(2) The Bluebird is injected with the same name of ES6 native Promise, this is "keyword confliction", thus often bring confusion. We had talking about renaming the Bluebird Promise, but this is definitely a huge work, as so many files need be modified

I don't find a perfect solution right now, hope you can have a good one.

iceiilin:
@yyscamper good comments. will keep in mind and think if any method applicable to eliminate this problem.

PR #470, opened 2017-07-06
 Get node IP from notification AMQP events

pengz1:
1. Why we need to get node IP:
RackHD needs to access http service on a compute node in  diag related tasks with node IP. The normal way to get IP/MAC from lookups can't work because lookups may contain more than one IPs that are legacy and incorrect. 

2. How this feature is implemented:
Retried node IP in on-http and add IP into notification AMQP messages => Install OS job retrieve IP from AMQP message and add it to graph context => following jobs to retrieve IP from graph context.

Jenkins depends on:
https://github.com/RackHD/on-http/pull/685

### DIFF
--- a/lib/jobs/install-os.js
+++ b/lib/jobs/install-os.js
@@ -48,7 +48,6 @@ function installOsJobFactory(
         self.nodeId = self.context.target;
         self.profile = self.options.profile;
         self.taskId = taskId;
-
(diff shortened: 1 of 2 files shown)


pengz1:
@iceiilin @nortonluo

pengz1:
@iceiilin @nortonluo

iceiilin:
test this please

PR #472, opened 2017-07-07
[RAC-5535] Implement clear sel by usage feature

pengz1:
**Background**
For emc platforms, we monitored sel to get firmware update status. However in some cases sel might not have enough space to store firmware update logs. Thus we need to clear sel before firmware updating to make sure we have enough space to store monitor firmware update process.

**Details**
1. SEL will only be clear by RackHD when maximum usage is reached and user explicitly permit clear sel in payload
2. User can use maxSelUsage parameter to control at which point RackHD will consider clearing SEL
3. User can use selForceClear parameter to permit RackHD clearing SEL when maxSelUsage is reached.
4. If maxSelUsage is not reached, SEL will not be cleared.

Paired with @nortonluo 

@iceiilin @anhou @nortonluo

### DIFF
--- a/lib/jobs/clear-sel-by-usage.js
+++ b/lib/jobs/clear-sel-by-usage.js
@@ -41,7 +41,7 @@ function ClearSelByUsageJobFactory(
     function ClearSelByUsageJob(options, context, taskId) {
         ClearSelByUsageJob.super_.call(this, logger, options, context, taskId);
         this.nodeId = context.target;
-        this.selForceClear = options.selForceClear || false;
+        this.selClear = options.selClear || false;
         this.maxSelUsage = options.maxSelUsage || 60;
     }
     util.inherits(ClearSelByUsageJob, BaseJob);
@@ -66,12 +66,12 @@ function ClearSelByUsageJobFactory(
             if (_.isNaN(usePercent)){
                 throw new Error("Can not get node SEL usage status");
             }
-            if (usePercent > self.maxSelUsage){
-                if (self.selForceClear) {
+            if (usePercent >= self.maxSelUsage){
+                if (self.selClear) {
                     return ipmitool.clearSel(obmConfig.host, obmConfig.user, obmConfig.password);
                 } else {
                     var errMsg = "Sel usage is %s\%, exceeds max sel usage %s\%".format(
-                        usePercent, 
+                        usePercent,
                         self.maxSelUsage
                     );
                     throw new Error(errMsg);
--- a/lib/jobs/clear-sel-by-usage.js
+++ b/lib/jobs/clear-sel-by-usage.js
@@ -1,0 +1,94 @@
+// Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
+
+'use strict';
+
+var di = require('di');
+
+module.exports = ClearSelByUsageJobFactory;
+di.annotate(ClearSelByUsageJobFactory, new di.Provide('Job.Clear.Sel.By.Usage'));
+di.annotate(ClearSelByUsageJobFactory, new di.Inject(
+    'Job.Base',
+    'JobUtils.Ipmitool',
+    'JobUtils.IpmiCommandParser',
+    'Logger',
+    'Util',
+    'Assert',
+    'Promise',
+    '_',
+    'Services.Waterline'
+));
+
+function ClearSelByUsageJobFactory(
+    BaseJob,
+    ipmitool,
+    parser,
+    Logger,
+    util,
+    assert,
+    Promise,
+    _,
+    waterline
+) {
+    var logger = Logger.initialize(ClearSelByUsageJobFactory);
+
+    /**
+     *
+     * @param {Object} options
+     * @param {Object} context
+     * @param {String} taskId
+     * @constructor
+     */
+    function ClearSelByUsageJob(options, context, taskId) {
+        ClearSelByUsageJob.super_.call(this, logger, options, context, taskId);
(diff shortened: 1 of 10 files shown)


spec/lib/jobs/clear-sel-by-usage-spec.js
```diff
@@ -0,0 +1,128 @@
...
+    });
+
+    it("should throw sel is almost full error", function() {
+        this.sandbox.stub(ipmitool, 'selInformation').resolves(mockData);
+        var _job = new ClearSelJob({maxSelUsage: 50}, context, taskId);
+        this.sandbox.stub(_job, '_done');
+        return _job._run()
+        .then(function(){
+            var err = new Error("Sel usage is 57\%, exceeds max sel usage 50\%");
+            expect(_job._done).to.callOnce;
```
yyscamper: As I remember, there is no `callOnce` in sinon, the correct usage is "to.be.calledOnce". 

The chain function style (xxx.xxx.xxx) cannot detect the typo error, so unit test will not catch this error.

lib/jobs/clear-sel-by-usage.js
```diff
@@ -0,0 +1,94 @@
...
+            return parser.parseSelInformationData(selRawData);
+        })
+        .then(function(selData){
+            var percent =  _.trimRight(selData["Percent Used"], '%');
+            var usePercent = parseInt(percent);
+            if (_.isNaN(usePercent)){
+                throw new Error("Can not get node SEL usage status");
+            }
+            if (usePercent > self.maxSelUsage){
+                if (self.selForceClear) {
```
yyscamper: A little naming suggestion. In my understanding, the `selForceClear` means always clearing the SEL no matter what's the current usage. I suggest changing another name to reflect this is to automatically clear SEL when it reach threshold.

lib/task-data/schemas/clear-sel-by-usage.json
```diff
@@ -0,0 +1,23 @@
+{
+    "copyright": "Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.",
+    "definitions": {
+        "selForceClearFlag": {
```
yyscamper: I think this is a typo, as I see your code uses `selForceClear`

lib/task-data/schemas/clear-sel-by-usage.json
```diff
@@ -0,0 +1,23 @@
...
+            "type": "boolean",
+            "description": "User specifies whether force to clear sel when sel entry is almost full"
+        },
+        "maxSelUsagePercent": {
```
yyscamper: Also this is a typo, the code use variable `maxSelUsage` rather than `maxSelUsagePercent`

lib/task-data/schemas/clear-sel-by-usage.json
yyscamper: Ignore this comment, I found you change the name in `properties`

lib/task-data/schemas/clear-sel-by-usage.json
yyscamper: ignore this comment.

spec/lib/jobs/clear-sel-by-usage-spec.js
```diff
@@ -0,0 +1,128 @@
...
+    });
+
+    it("should throw can not get node sel usage status error", function() {
+        var _mockData = "sel \n Percent Used 25% : Anything";
+        this.sandbox.stub(ipmitool, 'selInformation').resolves(_mockData);
+        var _job = new ClearSelJob(options, context, taskId);
+        this.sandbox.stub(_job, '_done');
+        return _job._run()
+        .then(function(){
+            expect(_job._done).to.callOnce;
```
yyscamper: `to.be.calledOnce`

lib/jobs/clear-sel-by-usage.js
```diff
@@ -0,0 +1,94 @@
...
+        .then(function(selRawData){
+            return parser.parseSelInformationData(selRawData);
+        })
+        .then(function(selData){
+            var percent =  _.trimRight(selData["Percent Used"], '%');
+            var usePercent = parseInt(percent);
+            if (_.isNaN(usePercent)){
+                throw new Error("Can not get node SEL usage status");
+            }
+            if (usePercent > self.maxSelUsage){
```
yyscamper: use `>=` is more reasonable, in my personal opinion.

spec/lib/jobs/clear-sel-by-usage-spec.js
pengz1: Good catch

spec/lib/jobs/clear-sel-by-usage-spec.js
pengz1: Good catch.

lib/jobs/clear-sel-by-usage.js
iceiilin: agree. Maybe something like "selClearWhenExceed"? not quite satisfying though

lib/jobs/clear-sel-by-usage.js
pengz1: Nop, it is not just only judging isPercent but have to compare the value.

nortonluo: I think >= is better. In a corner case, our definition maximum value in schema is 100. If user se 100 as a parameter, which is valid here, the code will not do clear if use ">"

PR #478, opened 2017-07-14
RAC-4123 can't create ipmi obm settings from rmm catalog

manfrednde:
I changed the operator so that only the variable value is compared instead of its value and type. 

@johren @jimturnquist


manfrednde:
Closing this PR as user will need to pass ipmichannel as designed ( string) instead of  using as string or integer.

PR #480, opened 2017-07-18
RAC-5414 - LogServices/sel/Entries "Cannot read property '3' of null"

larry-dean:
* If Row[1] contains "Pre-Init' force row[1] and row[2] to ""
* Add test for "Pre-Init" parsing
* Correct spelling


larry-dean:
@jimturnquist @BrentHiggins @linlynn 

This must me merged after: https://github.com/RackHD/on-http/pull/701

PR #486, opened 2017-07-28
Modify friendlyName of force pxe boot task

iceiilin:
**Background**
When using on-web-ui, it is found that there are two "reboot node" tasks with "forcePxeBoot" as its content.

**Details**
forcePxeBoot task uses "Reboot Node" as its friendly name wrongly. Modify it to "set node force pxeboot", the same structure as pxe boot task.

**Test Done**
On-web-ui can find "reboot node" and "force pxe boot" tasks correctly.

**Reviewer**
@changev @pengz1 @anhou

### DIFF
--- a/lib/task-data/tasks/force-pxe-boot.js
+++ b/lib/task-data/tasks/force-pxe-boot.js
@@ -3,7 +3,7 @@
 'use strict';
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


pengz1:
+1

PR #488, opened 2017-07-31
 Fix RAC 5789: Native promise leads to failure in catalog-searcher.js

pengz1:
Background
@wangzhizheng found we are using native promise while bluebird is required.

Details
Bluebird is introduced in catalog-searcher.js

@iceiilin @anhou


PengTian0:
test this please

PR #491, opened 2017-08-09
add manager catalog to wsman inventory collection

phelpdh:
manager catalog will provide idrac details.
This is a fix for RAC-5348 - Ethernet catalog information missing from Dell-based node catalogs

### DIFF
--- a/lib/jobs/dell-wsman-inventory.js
+++ b/lib/jobs/dell-wsman-inventory.js
@@ -43,8 +43,6 @@ function DellWsmanInventoryJobFactory(BaseJob, Logger, Promise, util, waterline,
         this.inventories = undefined;
         this.retrys = undefined;
         this.dellConfigs = undefined;
-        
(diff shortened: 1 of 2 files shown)


lib/jobs/dell-wsman-inventory.js
```diff
@@ -158,11 +158,13 @@ function DellWsmanInventoryJobFactory(BaseJob, Logger, Promise, util, waterline,
       	var requestUri = '';
       	if(self.nodeType === 'compute'){
       		requestUri = self.dellConfigs.services.inventory.serverCallback;
+            if(type === 'manager'){  // special case for manager endpoint - use 2.0 api
```
phelpdh: Special case to use new 2.0 api in the inventory service.  This can be removed if we shift all wsman inventory collection to the 2.0 api.  The 2.0 api will provide all data returned by iDRAC without translation to model.

PR #493, opened 2017-08-11
Disable timer prior to setting pxe bootdev

keedya:
- Some HW takes longer then one minute to come up. Default timer based on ipmmi spec is set to 60 
   seconds. 
- The raw command 0x00 0x08 0x03 0x08 Disable the timer


keedya:
@geoff-reid @johren @hohene

stuart-stanley:
This is past merge time, but in the interest of following the code-base I have a question:

Why was this needed exactly? (The comment at least indicates what the raw command is doing, but no reason -why-). Thanks, --stuart

PR #500, opened 2017-08-24
Add clear ipmi watch dog task and job

nortonluo:
## Background
Some server have enabled ipmi watchdog when boot to their OS. if the watch dog is not feed, the system will be judge as boot failure and force to reboot. Our micokernel and emc diag don't build in watchdog clear naively. When use those servers, it need run watchdog clear as part of the whole workflow, right after boot kernel success.
## Details
Add a task to do clearWatchdog. Add clearWatchd dog entry in the OBM bask task. Add "mc watchdog off" as the clear command in the obm service. Add related unit test.
## Test Result
Validated on EMC platform-O .

### DIFF
--- a/lib/services/obm-service.js
+++ b/lib/services/obm-service.js
@@ -457,6 +457,20 @@ function obmServiceFactory(
         return this.retryObmCommand('forceBootPxe');
     };
 
+     /**
+     * Clear the boot up watchdog to make those servers not reboot after entering OS.
+     * Not all OBM services will implement this function.
+     *
+     * @memberOf ObmService
+     * @function
+     *
+     * @param {String} nodeId
+     * @returns {Promise}
+     */
+    ObmService.prototype.clearWatchDog = function() {
+        return this.retryObmCommand('clearWatchDog');
+    };
+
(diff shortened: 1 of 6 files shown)


lib/services/obm-service.js
```diff
@@ -457,6 +457,22 @@ function obmServiceFactory(
...
     };
 
+     /**
+     * Clear the boot up watchdog to make those servers not reboot after entering OS.
+     * Not all OBM services will implement this function.
+     *
+     * @memberOf ObmService
+     * @function
+     *
+     * @param {String} nodeId
```
pengz1: This line is not necessary

lib/services/obm-service.js
```diff
@@ -457,6 +457,22 @@ function obmServiceFactory(
...
+     * @returns {Promise}
+     */
+    ObmService.prototype.clearWatchDog = function() {
+        // Delay less here since some OBM services
+        // just won't have this call.
```
iceiilin: Could you help to explain the comment here? Or remove it if unnecessary.

PR #507, opened 2017-09-06
Update the regex for the CoreOs install Schema

tannoa2:
Update the regex of the VaultToken to match the initially intended regex in the following commit: 
https://github.com/RackHD/on-tasks/commit/cc246fed12a229172af1a4e9b83f655857d5b1bf 

It looks like the [ajv ](https://www.npmjs.com/package/ajv )Json validation library doesn't support  [POSIX Bracket Expressions](http://www.regular-expressions.info/posixbrackets.html). Used ASCII regexp instead. 

Opened  bug is [here  ](https://rackhd.atlassian.net/browse/RAC-6065).

### DIFF
--- a/lib/task-data/schemas/types-installos.json
+++ b/lib/task-data/schemas/types-installos.json
@@ -313,7 +313,7 @@
         "VaultToken": {
             "type": "string",
             "description": "The token used for unwrapping a wrapped Vault response",
-            "minLength": 1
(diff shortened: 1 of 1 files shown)


tannoa2:
@RackHD/corecommitters @RackHD/veyron @tgelter

tgelter:
👍
Thanks @tannoa2!

PR #510, opened 2017-09-12
Refine firmware file upload method for EMC diag firmware update job

nortonluo:
## Background
The original EMC diag firmware update job need user upload the firmware image to the RackHD server himself.  User need have direct storage access in RackHD. That is not a safe way.  And this is also not align with the way used in other firmware update method in on-skupack.

## Details
We now pack the image file into the skupack. Thus, we can get the file from on-http. To make on-http can map the static folder in skupack, it is needed to use customized HTTP header 'X-Real-IP' . 
We change to use the "/api/upload/file" api in emc diag instead of "/api/upload/folder", user now don't have to pack the firmware file as zip file.
We also refined the name of some parameter. The firmwareName is now changed to firmwareType, type bios is changed to spi. This will avoid some conlusion.

## Test Result
Tested pass on EMC platform-O.


lib/jobs/emc-diag-update-firmware.js
```diff
@@ -72,10 +72,11 @@ function diagUpdateFirmwareFactory(
             return self.diagTool.getAllDevices();
         })
         .then(function(){
-            return self.diagTool.uploadImageFile(self.localImagePath);
+            return self.diagTool.uploadImageFile(
+        self.imageUrl, self.imageName,'/api/upload/file');
```
pengz1: indent problem.

lib/task-data/schemas/emc-diag-update-firmware.json
```diff
@@ -1,10 +1,10 @@
 {
     "copyright": "Copyright 2017, Dell EMC, Inc.",
     "definitions": {
-        "localImagePath": {
-            "description": "Image zip file path",
+        "imageUrl": {
+            "description": "Image URL in on-http",
             "type": "string",
-            "pattern": "^/.*\\.zip$"
+            "pattern": ".*\\.bin$"
```
pengz1: If it is a sku pack image, you could provide more detailed schema here.

nortonluo:
@linl7 @pengz1 @zhuanj3

PR #511, opened 2017-09-13
RAC-6100 Add tasks for ucs poller and their unit tests

mcgG:
Detail
----

Add tasks for ucs poller as well as their unit test

### DIFF
--- a/lib/task-data/base-tasks/ucs.js
+++ b/lib/task-data/base-tasks/ucs.js
@@ -1,0 +1,12 @@
+// Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Base Ucs Task',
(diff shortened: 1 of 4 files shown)


lib/task-data/base-tasks/ucs.js
```diff
@@ -0,0 +1,12 @@
+// Copyright 2017, EMC, Inc.
```
pengz1: EMC => Dell EMC

lib/task-data/tasks/ucs-poller.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2016, EMC, Inc.
```
pengz1: // Copyright 2016, EMC, Inc. => // Copyright 2017, Dell EMC, Inc.

PR #512, opened 2017-09-13
 Add create UCS default pollers

pengz1:
Background
This is part of UCS poller design, to create default UCS pollers.

Details
1. Refined create default poller method for UCS/WSMAN/Redfish
2. Add UCS create default feature.

Paired with @bbcyyb 
@anhou @iceiilin  @mcgG

### DIFF
--- a/lib/jobs/create-default-pollers.js
+++ b/lib/jobs/create-default-pollers.js
@@ -52,7 +52,6 @@ function createDefaultPollersJobFactory(
      * @param {String} obmName: OBM service name required to create pollers
      */
     CreateDefaultPollersJob.prototype._createWorkitem = function(nodeId, poller, obmName) {
-        assert.isMongoId(nodeId, 'nodeId');
         return waterline.obms.findByNode(nodeId, obmName, true)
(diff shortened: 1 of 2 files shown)


lib/jobs/create-default-pollers.js
```diff
@@ -44,41 +44,63 @@ function createDefaultPollersJobFactory(
...
-    
+
+    /**
+     * @memberOf CreateDefaultPollersJob
+     * @param {String} nodeId: node id string
+     * @param {Object} poller: poller to be created
+     * @param {String} obmName: OBM service name required to create pollers
+     */
+    CreateDefaultPollersJob.prototype._createWorkitem = function(nodeId, poller, obmName) {
+        assert.isMongoId(nodeId, 'nodeId');
```
yyscamper: It's better to remove this check, as nodeId may not always be MongoID since RackHD supports multiple kind of database.

pengz1: @yyscamper  Good catch

bbcyyb:
+1

PR #514, opened 2017-09-13
On-Task changes to have ConfigureBios Task and Job

muqeethkowkab:
Made the changes and ran the Hound checks.

### DIFF
--- a/lib/jobs/dell-wsman-configure-bios.js
+++ b/lib/jobs/dell-wsman-configure-bios.js
@@ -87,14 +87,14 @@ function DellWsmanConfigureBiosFactory(
 
     DellWsmanConfigureBiosJob.prototype.configureBios = function(obm) {
 
-        if (!validator.isIP(this.options.serverIP)) {
+        if (!validator.isIP(obm.config.host)) {
             throw new Error('Invalid ServerIP');
         }
 
         var self = this;
         var data = {
             "serverRequest": {
-                "serverIP": this.options.serverIP,
+                "serverIP": obm.config.host,
                 "serverUsername": obm.config.user,
                 "serverPassword": encryption.decrypt(obm.config.password),
             },
--- a/lib/jobs/dell-wsman-configure-bios.js
+++ b/lib/jobs/dell-wsman-configure-bios.js
@@ -5,9 +5,9 @@
 var di = require('di'),
     urlParse = require('url-parse');
 
-module.exports = DellWsmanExportSCPJobFactory;
-di.annotate(DellWsmanExportSCPJobFactory, new di.Provide('Job.Dell.Wsman.Export.SCP'));
-di.annotate(DellWsmanExportSCPJobFactory, new di.Inject(
+module.exports = DellWsmanConfigureBiosFactory;
+di.annotate(DellWsmanConfigureBiosFactory, new di.Provide('Job.Dell.Wsman.ConfigureBios'));
+di.annotate(DellWsmanConfigureBiosFactory, new di.Inject(
     'Job.Base',
     'Logger',
     'Promise',
@@ -22,7 +22,7 @@ di.annotate(DellWsmanExportSCPJobFactory, new di.Inject(
     'validator'
 ));
 
-function DellWsmanExportSCPJobFactory(
+function DellWsmanConfigureBiosFactory(
     BaseJob,
     Logger,
     Promise,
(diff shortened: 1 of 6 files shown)


lib/jobs/dell-wsman-configure-bios.js
```diff
@@ -0,0 +1,151 @@
...
+    DellWsmanConfigureBiosJob.prototype.configureBios = function(obm) {
+
+        if (!validator.isIP(this.options.serverIP)) {
+            throw new Error('Invalid ServerIP');
+        }
+
+        var self = this;
+        var data = {
+            "serverRequest": {
+                "serverIP": this.options.serverIP,
```
phelpdh: Should be able to get this from obm settings (obm.config.host)

lib/jobs/dell-wsman-configure-bios.js
```diff
@@ -0,0 +1,151 @@
...
+            });
+    };
+
+    DellWsmanConfigureBiosJob.prototype.checkOBM = function() {
+        var self = this;
+        logger.info('checkOBM: Self.nodeID: ' + self.nodeId);
+        return waterline.nodes.findByIdentifier(self.nodeId)
+            .then(function(result) {
+                self.nodeType = result.type;
+                if (self.nodeType !== 'compute') {
```
yyscamper: Please change to use this constant: https://github.com/RackHD/on-core/blob/master/lib/common/constants.js#L324

lib/jobs/dell-wsman-configure-bios.js
```diff
@@ -0,0 +1,151 @@
...
+                    logger.info('SCP ConfigureBios is not applicable to node type ' + self.nodeType);
+                    self.cancel();
```
yyscamper: `self.cancel` is designed to be called in baseTask or Task level, the job don't need to cancel itself. You just need to threw an error, the task engine will cancel the task.

lib/jobs/dell-wsman-configure-bios.js
```diff
@@ -0,0 +1,151 @@
...
+        this.nodeId = this.context.target;
+    }
+
+    util.inherits(DellWsmanConfigureBiosJob, BaseJob);
+
+    DellWsmanConfigureBiosJob.prototype._run = function () {
+        var self = this;
+        self.dell = configuration.get('dell');
+
+        if (!self.dell || !self.dell.services || !self.dell.services.configuration) {
```
yyscamper: A trivial style suggestion, if you design a function to return promise, then it's better to wrap all error into Promise, so the caller can use promise only to catch all errors, rather than use both promise catch and `try/catch`

lib/jobs/dell-wsman-configure-bios.js
```diff
@@ -0,0 +1,151 @@
...
+        if (!validator.isIP(obm.config.host)) {
+            throw new Error('Invalid ServerIP');
+        }
+
+        var self = this;
+        var data = {
+            "serverRequest": {
+                "serverIP": obm.config.host,
+                "serverUsername": obm.config.user,
+                "serverPassword": encryption.decrypt(obm.config.password),
```
yyscamper: Since you have reveal the password in line 79, you don't need to decrypt it again.

lib/jobs/dell-wsman-import-scp.js
```diff
@@ -123,13 +123,19 @@ function DellWsmanImportSCPJobFactory(
                 logger.info('Response from SCP Microservice for Import System Configuration: ' +  response.body);
                 var json = JSON.parse(response.body);
                 logger.info('Status from SCP Microservice for Import System Configuration:  ' + json["status"]);
-                if(json["status"]=="OK") {
-                    return response;
+
+                if (json["status"]==="OK" && json.hasOwnProperty("xmlConfig") && null !== json["xmlConfig"]) {
```
yyscamper: I see this code block is duplicated across multiple files, suggest creating a function for it.

lib/task-data/tasks/dell-wsman-configure-bios.js
```diff
@@ -0,0 +1,20 @@
...
+    implementsTask: "Task.Base.Dell.Wsman.ConfigureBios",
+    options: {
+        attributes: null,
+        biosBootSequenceOrder: null,
+        hddSequenceOrder: null,
+        enableBootDevices: null,
+        disableBootDevices: null,
+        rebootJobType: null,
+        scheduledStartTime: null,
+        untilTime: null
```
yyscamper: Create a schema for these options, then RackHD can use it to do upfront parameter validation and automatically generate a document for this task, this is a link about Options Schemas: http://rackhd.readthedocs.io/en/latest/rackhd/tasks.html?highlight=Task%20Schema#options-schema

yyscamper:
Every new job, new baseTask, new task should have a corresponding unit-test file, you need to complete the unit-test

PR #517, opened 2017-09-26
Bugfix - Failed to create poller in parallel

bbcyyb:
### Background
When trying to create default poller in on-tasks/lib/job/create-default-pollers.js, I found I can't create poller in parallel.
The case is that I have a nodeId lists (like nodeIds=A,B,C) and want to create 6 pollers (like pollers= ["1","2","3","4","5","6"]) for each node. I use below mechanism:
Promise.map(pollers, function(poller){
...
Promise.map(nodeIds, function(nodeId)
{ ... return waterline.workitems.findOrCreate }
)
})
Totally 18 pollers will be created. however, all pollers are in a mess. Node A may have 3 poller "1", node B may have 3 poller "2" and node C have12 pollers without poller "1" and "2".
There should be some problem with waterline while create pollers in parallel. I simply solved this problem by change the inner Promise.map to Promise.each. Promise.each actually executes in serial.
### Details
The reason of this issue is we used the SAME poller config obj to build data entity before writing into database in multiple parallel functions.
Promise.map has a option argument named concurrency, default is 3. The concurrency limit applies to Promises returned by the mapper function and it basically limits the number of Promises created. For example, if concurrency is 3 and the mapper callback has been called enough so that there are three returned Promises currently pending, no further callbacks are called until one of the pending Promises resolves. So the mapper function will be called three times and it will be called again only after at least one of the Promises resolves.
Once we use Promise.map, the data building has been process three times before the first perform a write to database. So that poller config obj may be written many times by different parallel function, will end up either with corrupted information or missing information.
#### How to:
Create a deep clone for poller config object before performing database operation, ensure each database processing has a single poller config for itself.
Add deep clone logic on create-default-pollers.js line 64
#### Scope:
- Create UCS default pollers
- Create Wsman default pollers
- Create Redfish default pollers
### UnitTest
Add new logic to check parameters nodeId and poller command name which will be passed into poller database operation is expected or not.

@anhou @iceiilin @pengz1 @lanchongyizu @mcgG

### DIFF
--- a/lib/jobs/create-default-pollers.js
+++ b/lib/jobs/create-default-pollers.js
@@ -52,6 +52,7 @@ function createDefaultPollersJobFactory(
      * @param {String} obmName: OBM service name required to create pollers
      */
     CreateDefaultPollersJob.prototype._createWorkitem = function(nodeId, poller, obmName) {
+        var clonedPoller = _.cloneDeep(poller);
         return waterline.obms.findByNode(nodeId, obmName, true)
         .then(function(obmSetting){
             if(_.isEmpty(obmSetting)){
@@ -60,13 +61,13 @@ function createDefaultPollersJobFactory(
             }
         })
         .then(function(){
-            poller.node = nodeId;
+            clonedPoller.node = nodeId;
             return waterline.workitems.findOrCreate(
                 {
                     node: nodeId,
-                    "config.command": poller.config.command
+                    "config.command": clonedPoller.config.command
                 },
-                poller
+                clonedPoller
             )
             .then(function(workitem){
                 var message = _.dropRight(obmName.split('-'), 2).join(" ");
@@ -84,8 +85,7 @@ function createDefaultPollersJobFactory(
     CreateDefaultPollersJob.prototype.createUcsPoller = function(nodeIds, poller){
         var self = this;
         return Promise.map(nodeIds, function(nodeId) {
-            var clonedPoller = _.cloneDeep(poller);
-            return self._createWorkitem(nodeId, clonedPoller, "ucs-obm-service");
+            return self._createWorkitem(nodeId, poller, "ucs-obm-service");
         });
     };
 
--- a/lib/jobs/create-default-pollers.js
+++ b/lib/jobs/create-default-pollers.js
@@ -83,8 +83,9 @@ function createDefaultPollersJobFactory(
      */
     CreateDefaultPollersJob.prototype.createUcsPoller = function(nodeIds, poller){
         var self = this;
-        return Promise.each(nodeIds, function(nodeId) {
-            return self._createWorkitem(nodeId, poller, "ucs-obm-service");
+        return Promise.map(nodeIds, function(nodeId) {
+            var clonedPoller = _.cloneDeep(poller);
+            return self._createWorkitem(nodeId, clonedPoller, "ucs-obm-service");
         });
     };
 
--- a/lib/jobs/create-default-pollers.js
+++ b/lib/jobs/create-default-pollers.js
@@ -48,11 +48,11 @@ function createDefaultPollersJobFactory(
     /**
      * @memberOf CreateDefaultPollersJob
      * @param {String} nodeId: node id string
-     * @param {Object} poller: poller to be created
+     * @param {Object} pollerTemplate: poller template to be created
      * @param {String} obmName: OBM service name required to create pollers
      */
-    CreateDefaultPollersJob.prototype._createWorkitem = function(nodeId, poller, obmName) {
-        var clonedPoller = _.cloneDeep(poller);
+    CreateDefaultPollersJob.prototype._createWorkitem = function(nodeId, pollerTemplate, obmName) {
+        var poller = _.cloneDeep(pollerTemplate);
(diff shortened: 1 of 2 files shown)


lib/jobs/create-default-pollers.js
```diff
@@ -83,8 +83,9 @@ function createDefaultPollersJobFactory(
      */
     CreateDefaultPollersJob.prototype.createUcsPoller = function(nodeIds, poller){
         var self = this;
-        return Promise.each(nodeIds, function(nodeId) {
-            return self._createWorkitem(nodeId, poller, "ucs-obm-service");
+        return Promise.map(nodeIds, function(nodeId) {
+            var clonedPoller = _.cloneDeep(poller);
```
pengz1: This change is also required for CreateDefaultPollersJob.prototype.createWsmanPoller and CreateDefaultPollersJob.prototype.createRedfishPoller, suggest you to move it to _createWorkitem.

spec/lib/jobs/create-default-pollers-spec.js
```diff
@@ -235,16 +235,19 @@ describe("Job.Pollers.CreateDefault", function () {
...
+            expect(waterline.obms.findByNode).to.have.been.calledWith(
+                nodeIds[1], 'ucs-obm-service', true
+            );
+
+            var arrayLength = pollers.length * nodeIds.length;
+            _.each(Array.from({length: arrayLength}, function(v, i){ return i; }),
+                function(i) {
+                    expect(waterline.workitems.findOrCreate.getCall(i).args[0])
+                        .to.have.been.deep.equal({node: nodeIds[i % pollers.length], 
+                            'config.command': pollers[parseInt(i/pollers.length)].config.command});
```
pengz1: You should have another line to test waterline.workitems.findOrCreate.getCall(i).args[1].

lib/jobs/create-default-pollers.js
bbcyyb: I have considered add the same logic into Wsman and Redfish, but I am not sure if this bigger scope change will cause unexpected issues for these two pollers.
OK, I will move new logic into _createWorkitem method and test it.

spec/lib/jobs/create-default-pollers-spec.js
```diff
@@ -235,16 +235,19 @@ describe("Job.Pollers.CreateDefault", function () {
...
+            );
+            expect(waterline.obms.findByNode).to.have.been.calledWith(
+                nodeIds[1], 'ucs-obm-service', true
+            );
+
+            var arrayLength = pollers.length * nodeIds.length;
+            _.each(Array.from({length: arrayLength}, function(v, i){ return i; }),
+                function(i) {
+                    expect(waterline.workitems.findOrCreate.getCall(i).args[0])
+                        .to.have.been.deep.equal({node: nodeIds[i % pollers.length],
```
pengz1: have.been is not necessary.

spec/lib/jobs/create-default-pollers-spec.js
```diff
@@ -235,16 +235,19 @@ describe("Job.Pollers.CreateDefault", function () {
...
-                );
+            expect(waterline.obms.findByNode).to.have.been.calledWith(
+                nodeIds[0], 'ucs-obm-service', true
+            );
+            expect(waterline.obms.findByNode).to.have.been.calledWith(
+                nodeIds[1], 'ucs-obm-service', true
+            );
+
+            var arrayLength = pollers.length * nodeIds.length;
+            _.each(Array.from({length: arrayLength}, function(v, i){ return i; }),
```
pengz1: _.forEach has the same functionality and is more popular than _.each.

spec/lib/jobs/create-default-pollers-spec.js
bbcyyb: fixed.

spec/lib/jobs/create-default-pollers-spec.js
bbcyyb: fixed.

spec/lib/jobs/create-default-pollers-spec.js
bbcyyb: fixed.

lib/jobs/create-default-pollers.js
bbcyyb: fixed.

lib/jobs/create-default-pollers.js
```diff
@@ -52,6 +52,7 @@ function createDefaultPollersJobFactory(
      * @param {String} obmName: OBM service name required to create pollers
      */
     CreateDefaultPollersJob.prototype._createWorkitem = function(nodeId, poller, obmName) {
+        var clonedPoller = _.cloneDeep(poller);
```
anhou: It looks like the 'poller' here is just a 'pollerTemplate' , it could apply to different nodes, create the same type pollers in multiple nodes.

panpan0000:
Nice Description . +0

lib/jobs/create-default-pollers.js
bbcyyb: fixed.
Use 'pollerTempete' as variable to instead of poller in create-default-poller.js

PR #523, opened 2017-10-11
Fix ipmi-user-list catalog data

gavin-scott:
Code that parsed `sudo ipmitool -c user list N` assumed there
would be a header row printed and built the user hash keys from
the first line. However `ipmitool user list` only prints the
headers when invoked without `-c` which puts it into CSV output
mode. When invoked with `-c` there is no header line and the
first user was getting eaten up and treated as the header.

Example ipmitool output:

    monorail@monorail-micro:~$ sudo ipmitool user list 1
    ID  Name             Callin  Link Auth  IPMI Msg   Channel Priv Limit
    2   root             true    true       true       ADMINISTRATOR
    monorail@monorail-micro:~$ sudo ipmitool user -c list 1
    2,root,true,true,true,ADMINISTRATOR

This fix hard-codes the header values as the header that is
printed when invoked without `-c` looks nigh-impossible to
parse. The previous code was also eating the first user because
that line was taken as the header line.

### DIFF
--- a/lib/utils/job-utils/command-parser.js
+++ b/lib/utils/job-utils/command-parser.js
@@ -114,7 +114,7 @@ function commandParserFactory(Logger, Promise, _) {
                     }
                 });
                 var parsed = {},
-                    header = lines.shift().split(','),
+                    header = ['ID', 'Name', 'Callin', 'Link Auth', 'IPMI Msg', 'Channel Priv Limit'],
                     columns = _.map(lines, function (line) {
                         return line.split(',');
                     });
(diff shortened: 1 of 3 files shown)


RackHDCI:
The Copyright Check has failed!
The files listed below do not have a valid copyright statement:
- spec/lib/utils/job-utils/stdout-helper.js

gavin-scott:
@geoff-reid , @keedya any changes needed here? The copyright check failed on a a pre-existing file, but I can update that if you want.

lanchongyizu:
@gavin-scott could you fix the Copyright failure?

RackHDCI:
The Copyright Check has failed!
The files listed below do not have a valid copyright statement:
- spec/lib/utils/job-utils/stdout-helper.js

gavin-scott:
@lanchongyizu sorry for delay, I added the copyright message to `stdout-helper.js` now. Looks like concourse CI failed though. I don't think I have access to that to check the issue.

gavin-scott:
Sure @lanchongyizu -- just did.

PR #532, opened 2017-10-24
Add ESXi installer option for clearing disk partitions

gavin-scott:
Add `clearDisk` option to Graph.InstallESXi to allow better control
over the `clearpart` command which is used to clear disk partition
tables during OS install. Valid `clearDisk` values are:

- `null` or unspecified: only clear the partition table on the disk
  the OS is being installed to. This is the same behavior as prior to
  this change.

- `alldrives`: clear the partition table on all drives.

- or a disk path that the OS can recognize.

### DIFF
--- a/lib/task-data/schemas/install-esxi.json
+++ b/lib/task-data/schemas/install-esxi.json
@@ -1,6 +1,20 @@
 {
     "copyright": "Copyright 2016, EMC, Inc.",
     "definitions": {
+        "ClearDisk": {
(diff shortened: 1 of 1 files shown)


gavin-scott:
@geoff-reid any changes required here from my side?

keedya:
@gavin-scott can u commit an empty commit to re-run the tests ? this is ready to be merged

gavin-scott:
@keedya -- done

PR #533, opened 2017-10-25
support IPDU(severtech-CWG-16H2C454) power control(power on/off)

kellylu2sym:
Background: 
test and implement  IPDU power control of rackhd , model is  CWG-16H2C454.

Details: 
code change in :lib/services/servertech-obm-service.js  .

test :  Do manual test of power on and power off with rackhd api

code review : @anhou @bbcyyb @pengz1 @panpan0000

### DIFF
--- a/lib/services/servertech-obm-service.js
+++ b/lib/services/servertech-obm-service.js
@@ -23,26 +23,30 @@ function serverTechObmServiceFactory(Promise, BaseObmService, _) {
     util.inherits(ServerTechObmService, BaseObmService);
 
     ServerTechObmService.prototype.reboot = function() {
-        return this._runInternal(this.mibPowerCommand + '.' + this.options.config.port, 'i', '3');
+        return this._runInternal([this.mibPowerCommand + '.' + this.options.config.port, 'i', '3']);
     };
 
     ServerTechObmService.prototype.powerOn = function() {
-        return this._runInternal(this.mibPowerCommand + '.' + this.options.config.port, 'i', '1');
+        return this._runInternal([this.mibPowerCommand + '.' + this.options.config.port, 'i', '1']);
     };
 
     ServerTechObmService.prototype.powerOff = function() {
-        return this._runInternal(this.mibPowerCommand + '.' + this.options.config.port, 'i', '2');
+        return this._runInternal([this.mibPowerCommand + '.' + this.options.config.port, 'i', '2']);
     };
 
     ServerTechObmService.prototype.powerStatus = function() {
         return this._runInternal(this.mibStatusCommand + '.' + this.options.config.port, 'snmpwalk')
         .then(function (result) {
-            if (_.contains(result.stdout.contains, 'INTEGER: on') ||
+            if (_.contains(result.stdout, 'INTEGER: on') ||
+                _.contains(result.stdout, 'INTEGER: idleOn') ||
+                _.contains(result.stdout.contains, 'INTEGER: on') ||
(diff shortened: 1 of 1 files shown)


lib/services/servertech-obm-service.js
```diff
@@ -23,27 +23,27 @@ function serverTechObmServiceFactory(Promise, BaseObmService, _) {
...
+        return this._runInternal([this.mibPowerCommand + '.' + this.options.config.port, 'i', '2']);
     };
 
     ServerTechObmService.prototype.powerStatus = function() {
         return this._runInternal(this.mibStatusCommand + '.' + this.options.config.port, 'snmpwalk')
         .then(function (result) {
-            if (_.contains(result.stdout.contains, 'INTEGER: on') ||
-                _.contains(result.stdout.contains, 'INTEGER: idleOn')) {
+            if (_.contains(result.stdout, 'INTEGER: on') ||
+                _.contains(result.stdout, 'INTEGER: idleOn')) {
```
pengz1: Are you sure result.stdout.contains is incorrect data path? If you are not sure about this, you should find the old devices and try on it. Otherwise, below lines are safer.

_.contains(result.stdout, 'INTEGER: on') ||
_.contains(result.stdout, 'INTEGER: idleOn' ||
_.contains(result.stdout.contains, 'INTEGER: on' ||
_.contains(result.stdout.contains, 'INTEGER: idleOn'

lib/services/servertech-obm-service.js
```diff
@@ -23,27 +23,27 @@ function serverTechObmServiceFactory(Promise, BaseObmService, _) {
...
                 return Promise.resolve(true);
             }
 
-            if (_.contains(result.stdout.contains, 'INTEGER: off') ||
-                _.contains(result.stdout.contains, 'INTEGER: idleOff')) {
+            if (_.contains(result.stdout, 'INTEGER: off') ||
+                _.contains(result.stdout, 'INTEGER: idleOff')) {
```
pengz1: The same as above comment.

lib/services/servertech-obm-service.js
kellylu2sym: ok，I will a old one to do some test

lib/services/servertech-obm-service.js
```diff
@@ -23,27 +23,27 @@ function serverTechObmServiceFactory(Promise, BaseObmService, _) {
     util.inherits(ServerTechObmService, BaseObmService);
 
     ServerTechObmService.prototype.reboot = function() {
-        return this._runInternal(this.mibPowerCommand + '.' + this.options.config.port, 'i', '3');
+        return this._runInternal([this.mibPowerCommand + '.' + this.options.config.port, 'i', '3']);
```
anhou: What's the background of changing it to array, it does't work in old way?

anhou:
Confused by 'PowerControllByRackhd', a more meaningful or specific title is better

lib/services/servertech-obm-service.js
kellylu2sym: ServerTechObmService.prototype._runInternal this function  allow to accept 2 arguments : command and file (file can be null). 
however , "return this._runInternal(this.mibPowerCommand + '.' + this.options.config.port, 'i', '3')" without the brackets actually send 3 arguments to the function ,and it will consider the string "i " as file value, it is not a available snmp command  and will run failed . so we need the brackets to make all the arguments value as the first arguments "command".

PR #537, opened 2017-10-27
Add additional data to node discovered event

gavin-scott:
Include the serial number, manufacturer and product from the
dmi catalog.

### DIFF
--- a/lib/jobs/node-discovered-alert-job.js
+++ b/lib/jobs/node-discovered-alert-job.js
@@ -59,7 +59,17 @@ function nodeAlertJobFactory(
                         self.context.data
                     );
                 }
-                return eventsProtocol.publishNodeEvent(node, 'discovered', nodeInfo);
+
+                return self._getDmiSystemInformation()
+                .then(function(systemInformation) {
+                    if (systemInformation) {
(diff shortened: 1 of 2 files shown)


lib/jobs/node-discovered-alert-job.js
```diff
@@ -71,5 +80,19 @@ function nodeAlertJobFactory(
...
 
+    NodeAlertJob.prototype._getHwInfoCatalog = function () {
+        var self = this;
+        return waterline.catalogs.findMostRecent({
+            node: self.nodeId,
+            source: 'dmi'
+        }).then(function (catalog) {
+            if (!catalog) {
+                return Promise.reject(
+                    new Error('Could not find dmi catalog data.'));
```
gavin-scott: @keedya mentioned in slack we probably don't want to fail here if the data isn't available. I can clean this up tomorrow to make sure missing catalog data doesn't fail this job.

gavin-scott:
@keedya updated this with better error handling and a test.

keedya:
@gavin-scott looks good. Will run our system testing and run it by the team.
It is a  +1 for me

keedya:
Adding more reviewere:
@geoff-reid @anhou  @johren

PR #540, opened 2017-11-02
Implement ucsCallback job

bbcyyb:
### Background
UCSM http response is slower than RackHD in some cases and will lead to RackHD fails to get data from UCSM via HTTP. Thus in ucs-service we implemented asynchronous API to avoid this and data will be post back to RackHD via a callback API.

### Details
Add new base job for ucs related job.
Unit test is included.

Depends on https://github.com/RackHD/on-http/pull/802, https://github.com/RackHD/ucs-service/pull/28

Paired with: @pengz1
@anhou @pengz1 @mcgG

### DIFF
--- a/lib/jobs/ucs-base-job.js
+++ b/lib/jobs/ucs-base-job.js
@@ -1,4 +1,4 @@
-//Copyright 2017, Dell EMC, Inc.
+// Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
 
 'use strict';
 
@@ -14,7 +14,7 @@ di.annotate(ucsBaseJobFactory,
         'Promise',
         'JobUtils.UcsTool',
         '_',
-        'Errors'));
+        'Assert'));
 
 function ucsBaseJobFactory(
     BaseJob,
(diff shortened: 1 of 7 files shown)


lib/jobs/ucs-base-job.js
```diff
@@ -0,0 +1,76 @@
...
+            .then(function(res) {
+                if (res && res.body && res.body.toUpperCase() !== "ACCEPTED") {
+                    throw new Error(
+                        "Request was not ACCEPTED. Please check input parameters.");
+                }
+                return self._subscribeHttpResponseUuidByPromisify(taskId);
+            });
+    };
+
+    UcsBaseJob.prototype._subscribeHttpResponseUuidByPromisify = function(id) {
```
anhou: Good job!

lanchongyizu: `_subscribeHttpResponseUuid `is already a `promise`(https://github.com/RackHD/on-tasks/blob/master/lib/jobs/base-job.js#L230), we don't need to promisify it again.

lanchongyizu: OK, since you want to Repromisify `_subscribeHttpResponseUuid`, errors should be processed by calling `callback`.

bbcyyb: Yes, current code will eat all error scenes. I'm going to add error handler for asynchronous processing, to make sure they can be found and handled.

bbcyyb:
Change parameter from taskId to a uuid string named callbackId which is used to define a amqp channel.

PR #541, opened 2017-11-02
A fix for on-tasks to be able to render workflow

mtannous:
This is a fix for on-tasks to be able to render objects and array of objects...

Signed-off-by: Bala Balakumaran <bala.balakumaran@dell.com>


keedya:
@geoff-reid

PR #545, opened 2017-11-03
Various fixes for UpdateSystemConfig after wsman refactor

gavin-scott:
- Increase receive timeout as the underlying SMI service is a very
  long-running synchronous REST service.

- Need to return value from _handleSyncRequest()

- JSON is already parsed by WsmanTool now


uppalk1:
@RackHD/corecommitters please take a look

PR #549, opened 2017-11-08
fix UI defect RAC-4416

kellylu2sym:
BackGround : 
when we create a workflow  without task definition , rackhd will add api path to the taskName ,when you change the workflow and need save again , it will failed when validation of the task with taskName like "api/2.0/workflows/tasks****" 
Solution : 
code change in task-graph.js . when validate ,we use actually task name  instead of the whole name witch contains api path 
Test : 
manual test 
Jira: 
https://rackhd.atlassian.net/browse/RAC-4416
Reviewer : 
@anhou @panpan0000 @bbcyyb

### DIFF
--- a/lib/task-graph.js
+++ b/lib/task-graph.js
@@ -165,6 +165,23 @@ function taskGraphFactory(
         });
     };
 
+   /*
+    *This function is help to get the task name for the task graph with
+    * without API path
+    */ 
+    var getTaskName = function(taskNameInput){
+        // check if there is "/" in the input task name ,if yes 
+        // it may have the api path before task name
+        var taskName = taskNameInput;
+        var slashOffset = taskNameInput.indexOf("\/");
+        if(slashOffset !== -1){
+            var nameArray = taskNameInput.split("\/");
+            // get the really taskName
+            taskName = nameArray[nameArray.length - 1];
+        }
+        return taskName;    
+    };
+
     /*
      * Take the tasks definitions in this.definition.tasks, generate instanceIds
      * to use for each task, and then create new Task objects that reference
@@ -208,10 +225,11 @@ function taskGraphFactory(
                 waitingOn: taskData.waitOn,
                 ignoreFailure: taskData.ignoreFailure
             };
-
             if (taskData.taskName) {
+                var taskNameInput = taskData.taskName;
+                var taskName = getTaskName(taskNameInput);
                 return self.constructTaskObject(
-                    taskData.taskName,
+                    taskName,
                     taskOverrides,
                     taskData.optionOverrides,
                     taskData.label
@@ -229,7 +247,7 @@ function taskGraphFactory(
             return self;
         });
     };
-
+ 
     TaskGraph.prototype.constructInlineTaskObject = function(_definition, taskOverrides,
            optionOverrides, label) {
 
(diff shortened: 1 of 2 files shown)


lib/task-graph.js
```diff
@@ -208,10 +208,15 @@ function taskGraphFactory(
                 waitingOn: taskData.waitOn,
                 ignoreFailure: taskData.ignoreFailure
             };
-
-            if (taskData.taskName) {
+            if (taskData.taskName && typeof(taskData.taskName) !== "undefined") {
```
anhou: the condition never goto 'typeof(taskData.taskName) !== "undefined"' which after '&&' , right?

lib/task-graph.js
```diff
@@ -208,10 +208,15 @@ function taskGraphFactory(
...
+                var taskName = taskData.taskName;
+                var contains = taskName.indexOf("\/");
+                if(contains !== -1){
```
anhou: please add annotations for string operations

lib/task-graph.js
```diff
@@ -360,7 +365,15 @@ function taskGraphFactory(
             assert.arrayOfObject(self.definition.tasks, 'Graph.tasks');
             return Promise.map(self.definition.tasks, function(taskData) {
                 if (!_.has(taskData, 'taskDefinition')) {
-                    return store.getTaskDefinition(taskData.taskName)
+                    var taskName = taskData.taskName;
```
anhou: duplicated code with above, please extract a function if possible

lib/task-graph.js
```diff
@@ -208,10 +208,15 @@ function taskGraphFactory(
                 waitingOn: taskData.waitOn,
                 ignoreFailure: taskData.ignoreFailure
             };
-
-            if (taskData.taskName) {
+            if (taskData.taskName && typeof(taskData.taskName) !== "undefined") {
+                var taskName = taskData.taskName;
+                var contains = taskName.indexOf("\/");
```
anhou: contains what, a descriptive name is better

spec/lib/task-graph-spec.js
```diff
@@ -130,6 +130,13 @@ describe('Task Graph', function () {
                 expect(graph).to.be.an.instanceof(TaskGraph);
             });
         });
+        it('should return this', function() {
```
panpan0000: nice

panpan0000:
a *silly* question:
is it possible that  there's different path but with same task name ?
ex:   
`api/workflows/tasks/wsman/Task.test`
`api/workflows/tasks/ucs/Task.test`

lib/task-graph.js
kellylu2sym: done,  however this line is in the new function.

lib/task-graph.js
```diff
@@ -165,6 +165,22 @@ function taskGraphFactory(
...
     };
 
+   /*
+    *This function is help to get the task name for the task graph with
+    * without API path
+    */ 
+    var getTaskName = function(taskName){
+        // check if there is "/" in the task name ,if yes 
+        // it may have the api path before task name
+        var contains = taskName.indexOf("\/");
```
anhou: temporary name could be isSlashContained. I think this is your purpose, it's more straightford than a verb 'contain

lib/task-graph.js
```diff
@@ -360,7 +377,11 @@ function taskGraphFactory(
             assert.arrayOfObject(self.definition.tasks, 'Graph.tasks');
             return Promise.map(self.definition.tasks, function(taskData) {
                 if (!_.has(taskData, 'taskDefinition')) {
-                    return store.getTaskDefinition(taskData.taskName)
+                    var taskName = taskData.taskName;
+                    if(taskName){
+                        taskName = getTaskName(taskName);
```
anhou: taskName in this function have two meanings.  former taskName is taskName input, latter is true taskName, it's not so good for clean code. so you could naming former one as taskNameInput, the latter is taskName. Otherwise, it's a little werid that get 'taskName' from 'taskName'

panpan0000:
can you `git commit --amend` then  `git push` again, to trigger the Concourse?

PR #550, opened 2017-11-09
 Fix unittest can't complete issue for on-tasks.

pengz1:
**Background**
During upgrading node from 4 to 6/8, mocha is required to be updated. With latest mocha, unittest won't finish if there is pending tasks.

**Details**
Manually clean setInterval or setTimeout in our test cases if unittest triggered them.

@mcgG @PengTian0 @anhou

### DIFF
--- a/spec/lib/jobs/clean-work-items-spec.js
+++ b/spec/lib/jobs/clean-work-items-spec.js
@@ -37,7 +37,10 @@ describe("Job.Catalog.CleanWorkItems", function () {
         }];
 
         var job = new CleanWorkItems({}, { graphId: uuid.v4() }, uuid.v4());
-
+        job.isPending = sinon.stub();
+        job.isPending.onCall(0).returns(true);
+        job.isPending.onCall(1).returns(false);
+        job.options.queuePollInterval = 0;
         waterline.workitems.findExpired.resolves(workItems);
 
         job.run();
@@ -51,5 +54,6 @@ describe("Job.Catalog.CleanWorkItems", function () {
                 done(e);
             }
         });
+
     });
 });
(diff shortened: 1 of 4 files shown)


mcgG:
👍 +1

yyscamper:
I‘d suggest you doing these kind of cleanup work in the job code rather than the unit-test, some reason below:
（1) The product code still need to do the same clean up work
  (2) Hide the design detail and make the exported interface easy-to-use. Various jobs depends on various resources, the job itself should safely close these resource when it finishes.

Actually, now  the BaseJob has implemented a "_cleanup" hook for every job (see https://github.com/RackHD/on-tasks/blob/master/lib/jobs/base-job.js#L68）， just like the `_run` in every job, actually every job can implement a `_cleanup`, the `_cleanup` hook will be called whenever the job is succeed or failed. So I suggest you putting these cleanup code into the `_cleanup` hook. 

RackHD leverages the bluebird `disposor/using` for auto cleanup resource that used by each job, see http://bluebirdjs.com/docs/api/disposer.html for usage. That means, you should call `run` (not `_run`!) in unit-test, these kind of cleanup hook will take effect.

pengz1:
@yyscamper 
Not sure if I got you, there is cleanup for each job:
https://github.com/RackHD/on-tasks/blob/master/lib/jobs/run-graph.js#L104
https://github.com/RackHD/on-tasks/blob/master/lib/task.js#L358

 It is not that our product code has problem but our unittest has problem, 

As for run and _run, I think the difference lies in the test methodology. It is more easier to use _run as most case it is a promise. What we need do is to make sure cleanup is done correctly.

PR #552, opened 2017-11-10
update mocha

mcgG:
Update mocha from 2.1.0 to 4.0.1

**Background**
----
Current mocha 2.1.0 cannot return failure in PR Gate, to fix this, update is needed

@iceiilin @anhou @pengz1 @bbcyyb @PengTian0


anhou:
travisci tests are all pass, the test results details are green, it looks it's a travisci own issue that the light is yellow.

PR #557, opened 2017-11-15
add dell-wsman-reset-components task/base-task/job

lanchongyizu:
**Background**
Based on the previous investigation result, develop to port Graph.Dell.Racadm.ResetComponents and /Systems/{identifier}/Bios.ResetBios to use WSMAN API.
https://rackhd.atlassian.net/browse/RAC-6481

**Done**
add dell-wsman-reset-components task/base-task/job

@RackHD/corecommitters @nortonluo

### DIFF
--- a/lib/jobs/dell-wsman-discovery.js
+++ b/lib/jobs/dell-wsman-discovery.js
@@ -8,7 +8,7 @@ var di = require('di');
 module.exports = DellWsmanDiscoveryJobFactory;
 di.annotate(DellWsmanDiscoveryJobFactory, new di.Provide('Job.Dell.Wsman.Discovery'));
 di.annotate(DellWsmanDiscoveryJobFactory, new di.Inject(
-    'Job.Base',
+    'Job.Dell.Wsman.Base',
     'JobUtils.WsmanTool',
     'Logger',
     'Promise',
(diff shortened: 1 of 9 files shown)


lib/jobs/dell-wsman-discovery.js
```diff
@@ -8,7 +8,7 @@ var di = require('di');
 module.exports = DellWsmanDiscoveryJobFactory;
```
nortonluo: I think it's better to put wsman-discovery Refine in another PR(Including the unit test). Although this bios reset PR depends on wsman discover function to find a valid node. But from unit test perspective, there is no tight coupling.

lib/task-data/schemas/dell-wsman-reset-components.json
```diff
@@ -0,0 +1,20 @@
+{
+    "copyright": "Copyright 2017, Dell EMC, Inc.",
+    "definitions": {
+        "components": {
+            "type": "array",
+            "description": "A list of components that will be reset",
+            "items": {
+                "type": "string",
```
nortonluo: Use emum type could be better than string.
The parameter could only be bios, diag, drvpack, idrac and lcdata

lanchongyizu: The component name is case-insensitive, it could be BiOs.

PR #558, opened 2017-11-16
RAC-6214 : add dell wsman delete volume

yaolingling:
**Background**

Replace RACADM based RAID operation feature implementation with WSMAN API. This pr is used for deleting volume operation. 

**Details**

To solve this problem, a new graph is created which contains four tasks. The first task is used to getComponent xml file. The second task is used to update the xml file for deleting volume. The third task is used to delete volume. The last task is used to update mongodb. Details in https://rackhd.atlassian.net/browse/RAC-6214

**Reviewers**
@nortonluo @lanchongyizu @AlaricChan


lib/task-data/schemas/dell-wsman-getInventory.json
```diff
@@ -0,0 +1,29 @@
+{
+    "copyright": "Copyright © 2017 Dell Inc. or its subsidiaries. All Rights Reserved.",
+    "definitions": {
+        "verifySSL": {
+            "description": "Specify if SSL certificate is enabled",
+            "type": "boolean"
+        },
+        "_taskTimeout": {
```
yyscamper: This is an options for all tasks and it has been covered in common-task-options (https://github.com/RackHD/on-tasks/blob/master/lib/task-data/schemas/common-task-options.json#L22), so you don't need to describe it again.

lib/task-data/schemas/dell-wsman-delete-volume.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "properties": {
+        "volumeId": {
+            "$ref": "#/definitions/volumeId"
+        },
+        "shutdownType": {
+            "$ref": "#/definitions/shutdownType"
+        }
+    },
+    "required": ["volumeId"]
```
yyscamper: if shutdownType is optional, you'd better tell its default value in its description.

yyscamper: And there is difference between "setting a default value in task definition" and "describe it as optional in task schema".  If the job needs this option for execution, then the option has to be `required` no matter you set a default value in task definition or not.

lib/task-data/tasks/dell-wsman-delete-volume-getComponent.js
```diff
@@ -0,0 +1,15 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: "Dell Wsman GetComponent for delete volume",
+    injectableName: "Task.Dell.Wsman.Delete.Volume.getComponent",
+    implementsTask: "Task.Base.Dell.Wsman.getComponent",
+    optionsSchema: 'dell-wsman-delete-volume.json',
+    options: {
+        volumeId: "",
```
yyscamper: Could you please confirm, is the empty `volumeId` valid?

lib/jobs/dell-wsman-RAID.js
```diff
@@ -0,0 +1,99 @@
...
+    'Logger',
+    'Assert',
+    'Util',
+    'Services.Configuration',
+    'Errors',
+    'Services.Encryption',
+    'validator'
+));
+
+function DellWsmanRAIDFactory(
```
yyscamper: why no `_run` callback?

lib/utils/job-utils/smb2-client.js
```diff
@@ -0,0 +1,66 @@
...
+            domain: 'DOMAIN',
+            username: userName,
+            password: password
+        });
+        XMLSerializer = XmlDOM.XMLSerializer;
+        serializer = new XMLSerializer();
+    }
+
+    Smb2Client.prototype.readFile = function(fileName){
+        logger.info("Smb2Client readFile function, fileName:" + fileName);
```
yyscamper: use `logger.debug` seems to be more reasonable

spec/lib/jobs/dell-wsman-RAID-spec.js
```diff
@@ -0,0 +1,118 @@
...
+    it('Should send RAID request succesfully', function(){
+        validator.isIP.returns(true);
+        job.dell = configFile;
+        var response = {
+            "body": {
+                "response": "Request Submitted"
+            }
+        };
+        WsmanTool.prototype.clientRequest.resolves(response);
+        expect(job.doRAIDoperation(obms)).to.be.fulfilled;
```
yyscamper: this is a promise it has to be returned, otherwise this test case is always success no matter the promise is fufilled or rejected.

spec/lib/jobs/dell-wsman-RAID-spec.js
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    it('Should throw an error: Invalid ServerIP', function(){
+        validator.isIP.returns(false);
+        expect(function(){
+            job.doRAIDoperation(obms);
+        }).to.throw('Invalid ServerIP');
```
yyscamper: `throw` only works for sync function. For async function, use something like `return expect(...).to.be.rejected;'

spec/lib/jobs/dell-wsman-delete-volume-xml-spec.js
```diff
@@ -0,0 +1,91 @@
...
+    afterEach(function(){
+        sandbox.restore();
+    });
+
+
+    it('Should init wsman wsmanDeleteVolumeXml job succesfully', function(){
+        configuration.get.returns(configFile);
+        expect(function(){
+            job._run();
+        }).to.not.throw('Dell SCP  web service is not defined in smiConfig.json.');
```
yyscamper: `throw` cannot be used for async function, it has been explained above.

lib/jobs/dell-wsman-getComponent.js
```diff
@@ -0,0 +1,107 @@
...
+        if (json["status"] === "OK") {
+            return response;
+        } else {
+            throw new Error("Failed to getComponent from smi service");
+        }
+    };
+
+    WsmanGetComponentJob.prototype.getComponent = function(obm) {
+        if (!validator.isIP(obm.config.host)) {
+            throw new Error('Invalid ServerIP');
```
yyscamper: if you design the function to return a promise, then you'd better also wrap the error in promise, you will find it will become easier to write unit-test.

lib/jobs/dell-wsman-getComponent.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+        var self = this;
+        var componentNames = "";
+        if(self.options.volumeId !== undefined){
+            componentNames = self.options.volumeId.split(':')[1];
+        }else if(self.options.drives !== undefined){
```
yyscamper: The task schema misses the description for the option `drives`.
Meanwhile,  look like this is a typo `drives` ==> 'drivers'

lib/task-data/schemas/dell-wsman-getInventory.json
yaolingling: I will remove "_taskTimeout"

lib/task-data/schemas/dell-wsman-delete-volume.json
yaolingling: I will add description for default value.

lib/task-data/tasks/dell-wsman-delete-volume-getComponent.js
yaolingling: The volumeId check will be added.

lib/jobs/dell-wsman-RAID.js
yaolingling: _run callback is defined in it's parent class

lib/utils/job-utils/smb2-client.js
yaolingling: ok

lib/jobs/dell-wsman-getComponent.js
yaolingling: Spelling error.  It should be drivers. The option isn't used in delete volume. It will be defined in other  feature, such as "add volume"

lib/task-data/base-tasks/dell-wsman-getComponent.js
```diff
@@ -0,0 +1,15 @@
+// Copyright © 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Dell Wsman GetComponent Base Task',
+    injectableName: 'Task.Base.Dell.Wsman.GetXml',
```
lanchongyizu: Duplicated file.

yaolingling: it's deleted

PR #559, opened 2017-11-17
New Mongo doesn't store properties with dot notation

tannoa2:


keedya:
@geoff-reid this is fixed now

PR #564, opened 2017-11-30
RAC-6480 Refine get system config catalogs task & graph

nortonluo:
## Background
The original get system config components task is not working. it used share folder to handle .xml exchange file but it doesn't accept share folder username &password.
It only do catalog create. When there is already catalogs in db, it will get problem.
Also latest data format update in SMI service response make the result data incorrect.

## Action
1. Add share folder setting parameters. It should be still compatible with original call method.
2. Change catalog create method. Only create if there is no such record in db. If found similar on, it only do update.
3. Fixed the json format transfer bug.
4. Add task schema
5. Add unit test for all the jobs.

@lanchongyizu @yaolingling @AlaricChan @anhou  @leoyjchang


lib/jobs/dell-wsman-get-systemcomponents.js
```diff
@@ -62,25 +70,40 @@ function DellWsmanGetSystemConfigComponentsFactory(
...
 
     DellWsmanGetSystemConfigComponentsJob.prototype.getComponents = function(obm) {
-
-        if (!validator.isIP(this.options.serverIP) || !validator.isIP(this.options.shareAddress)) {
-            throw new Error('Invalid ServerIP/ShareAddress');
-        }
         var self = this;
-        this.options.serverUsername = obm.config.user;
-        this.options.serverPassword = encryption.decrypt(obm.config.password);
+        if (!self.options.serverIP || !self.options.serverUsername || ! self.options.serverPassword) {
```
AlaricChan: Confusion: if the logic shall be that if server information is inputted via payload, then there is no need to check OBM. Besides, such check point shall be placed in _initJob method?

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+            helper.require('/lib/jobs/base-job.js'),
+            helper.require('/lib/jobs/dell-wsman-base-job.js'),
+            helper.require('/lib/jobs/dell-wsman-get-systemcomponents.js'),
+            helper.require('/lib/utils/job-utils/wsman-tool.js')
+        ]);
+        waterline = helper.injector.get('Services.Waterline');
+        WsmanJob = helper.injector.get('Job.Dell.Wsman.Get.SystemConfigurationComponents');
+        uuid = helper.injector.get('uuid');
+        configuration = helper.injector.get('Services.Configuration');
+        //BaseJob = helper.injector.get('Job.Dell.Wsman.Base');
```
AlaricChan: remove line

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+
+    afterEach(function(){
+        sandbox.restore();
+    });
+
+    describe('wsman bios init job', function() {
+        it('Should init attributes for wsmanBios job successfully', function() {
+            sandbox.stub(configuration,'get').returns(configFile);
+            job._initJob();
+            expect(job.options.shareAddress).to.equal("20.20.20.20");
```
AlaricChan: return promise

lib/jobs/dell-wsman-get-systemcomponents.js
```diff
@@ -62,25 +70,40 @@ function DellWsmanGetSystemConfigComponentsFactory(
...
+            if (!obm) {
+                throw new Error('Could not get valid ServerIP');
+            }
+            else {
+                self.options.serverIP = obm.config.host;
+                self.options.serverUsername = obm.config.user;
+                self.options.serverPassword = encryption.decrypt(obm.config.password);
```
yyscamper: the obm model will reveal the password, I don't think this is a good place to to decryption here.

lib/task-data/schemas/dell-wsman-get-config-catalog.json
```diff
@@ -0,0 +1,81 @@
...
+        "sharePassword": {
+            "description": "NFS/CIFS server password",
+            "type": "string"
+        },
+        "shareName": {
+            "description": "Share path of the firmware repository folder on share server",
+            "type": "string"
+        },
+        "shareType": {
+            "description": "Specify whehter use NFS or CIFS as share server to hold firmware repository",
```
yyscamper: typo for whehter

lib/task-data/schemas/dell-wsman-get-config-catalog.json
```diff
@@ -0,0 +1,81 @@
...
+            "type": "integer",
+            "enum": [0, 2]
```
yyscamper: I think you should explanation what the number `0` and `2` means. For to be better, you could use a string  type, user directly inputs `nfs` or `cifs`, so that user doesn't have to remember the mapping.

lib/task-data/schemas/dell-wsman-get-config-catalog.json
```diff
@@ -0,0 +1,81 @@
...
+        },
+        "shutdownType": {
+            "description": "Specify the type of system shutdown: Graceful(0), Forced(1), default value:0",
```
yyscamper: The description tells the default `shutdownType` is zero, but I don't see the default value in task definition: https://github.com/RackHD/on-tasks/pull/564/files#diff-a4208a0c727e4c4e91866bbc80e01457R20

lib/task-data/schemas/dell-wsman-get-config-catalog.json
```diff
@@ -0,0 +1,81 @@
...
+            "type": "integer",
+            "enum": [0, 1]
+        },
+        "componentNames": {
+            "description": "A list of components to get config catalogs",
+            "type": "array",
+            "uniqueItems": true,
+            "minItems": 0
```
yyscamper: are you sure empty componentNames is valid?

lib/task-data/schemas/dell-wsman-get-config-catalog.json
```diff
@@ -0,0 +1,81 @@
+{
+    "copyright": "Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.",
+    "definitions": {
+        "serverIP": {
+            "description": "Dell server iDARC IP address",
+            "type": "string"
+        },
+        "serverUsername": {
+            "description": "Dell server iDARC username",
```
yyscamper: typo: `iDARC' ==> `iDRAC`

lib/task-data/schemas/dell-wsman-get-config-catalog.json
```diff
@@ -0,0 +1,81 @@
...
+            "type": "string"
+        },
+        "serverPassword": {
+            "description": "Dell server iDARC password",
```
yyscamper: typo: `iDARC' ==> `iDRAC`

lib/task-data/schemas/dell-wsman-get-config-catalog.json
```diff
@@ -0,0 +1,81 @@
+{
+    "copyright": "Copyright 2017 Dell Inc. or its subsidiaries. All Rights Reserved.",
+    "definitions": {
+        "serverIP": {
+            "description": "Dell server iDARC IP address",
+            "type": "string"
```
yyscamper: the serverIP should not be empty, so add more constraint (such as `minLength` or even `pattern`) for it will help user to identify problem earlier. The same for others.

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+        waterline.nodes = {
+            findByIdentifier: sandbox.stub()
+        };
+
+        waterline.obms = {
+            findByNode: sandbox.stub()
+        };
+
+        waterline.catalogs = {
+            findLatestCatalogOfSource: sandbox.stub(),
```
yyscamper: This function is originally a promise, so mock it also a promise will be better. The same for other waterline mocks.

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+
+    describe('handle cases for _handlesyncRequest function', function() {
+        it('Should return reponse successfully with iDARC parameters in options', function() {
+            job.options.serverAddress = "2.2.2.2";
+            job.options.serverUsername = "abc";
+            job.options.serverPassword = "123";
+            job.checkOBM.resolves(obms);
+            sandbox.stub(configuration,'get').returns(configFile);
+            sandbox.stub(job, 'getComponents');
+            job._run().then(function(){
```
yyscamper: return this promise!

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+        });
+
+        it('Should return reponse successfully with empty iDARC parameter', function() {
+            job.options.serverAddress = null;
+            job.options.serverUsername = null;
+            job.options.serverPassword = null;
+            job.checkOBM.resolves("");
+            sandbox.stub(configuration, 'get').returns(configFile);
+            sandbox.stub(job, 'getComponents');
+            job._run().then(function(){
```
yyscamper: return this promise

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+            job.options.serverPassword = "123";
+            job.checkOBM.resolves(obms);
+            sandbox.stub(configuration,'get').returns(configFile);
+            sandbox.stub(job, 'getComponents');
+            job._run().then(function(){
+                expect(job.getComponents).to.be.called;
+            });
+        });
+
+        it('Should return reponse successfully with empty iDARC parameter', function() {
```
yyscamper: typo: `IDARC` => `IDRAC`

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+            expect(job.getComponents(obms)).to.be.rejectedWith('Failed to Get the Requested Components');
+        });
+
+        it('Should work with obms when serverIP in task option is null', function(){
+            job.options.serverIP = null;
+            job.options.serverUsername = null;
+            job.options.serverPassword = null;
+            job.dell = configFile;
+            WsmanTool.prototype.clientRequest.resolves(fullResponse);
+            expect(job.getComponents(obms)).to.be.fullfilled;
```
yyscamper: return this promise!

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+        it('Should throw error: response has no expected field', function(){
+            job.options.serverIP = "2.2.2.2";
+            job.options.serverUsername = "abc";
+            job.options.serverPassword = "123";
+            job.dell = configFile;
+            var response = {
+                "body": '{"status":"OK"}'
+            };
+            WsmanTool.prototype.clientRequest.resolves(response);
+            expect(job.getComponents(obms)).to.be.rejectedWith('Failed to Get the Requested Components');
```
yyscamper: return this promise!

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+            var result = {
+                data: fullResponse.boby,
+                source: "idrac-wsman",
+                store: true,
+            };
+            var catalog = {};
+            waterline.catalogs.findLatestCatalogOfSource.resolves(catalog);
+            return job._handleSyncResponse(result)
+            .then(function(){
+                expect(waterline.catalogs.create).to.be.called.once;
```
yyscamper: called.once ==> "calledOnce"

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+                              "Model": "PowerEdge R730",
+                              "ServiceTag": "xxxxxx",
+                              "TimeStamp": "2017"}
+                }
+            };
+            waterline.catalogs.findLatestCatalogOfSource.resolves(catalog);
+            return job._handleSyncResponse(result)
+            .then(function(){
+                expect(waterline.catalogs.create).not.to.be.called;
+                expect(waterline.catalogs.updateByIdentifier).to.be.called.once;
```
yyscamper: called.once ==> "calledOnce"

spec/lib/jobs/dell-wsman-get-systemcomponents-spec.js
```diff
@@ -0,0 +1,233 @@
...
+            });
+        });
+
+        it('Should get error: data is invalid.  No catalog created.', function(){
```
yyscamper: Seems this description doesn't align with the code

lib/task-data/schemas/dell-wsman-get-config-catalog.json
nortonluo: I thought about change to nfs cifs as my previous firmware update PR. But this will bring in API change in this Graph. The original get component task is using 0/2 as input, and there are some other teams already using this API. So I have to stay the same

lib/task-data/schemas/dell-wsman-get-config-catalog.json
nortonluo: Yes, empty component Name for SMI service mean dump all config data field. And my default action set to dump all field. Maybe I could add this explain to schema.

lib/task-data/schemas/dell-wsman-get-config-catalog.json
nortonluo: Good idea!

lib/task-data/schemas/dell-wsman-get-config-catalog.json
yyscamper: Actually, you could support both, and mark the number format as `deprecated`.  So you don't break the compatibility but also encourage user to use the better one.

PR #565, opened 2017-12-01
add dell-wsman-configure-idrac job

lanchongyizu:
**Background**

Based on the previous investigation result, develop to port Graph.Dell.Configure.Idrac to use WSMAN API.
https://rackhd.atlassian.net/browse/RAC-6479

**Done**

add dell-wsman-configure-idrac job/schema/unit test

**Reviewers**

@RackHD/corecommitters @nortonluo @AlaricChan @yaolingling

### DIFF
--- a/spec/lib/jobs/dell-wsman-configure-idrac-spec.js
+++ b/spec/lib/jobs/dell-wsman-configure-idrac-spec.js
@@ -1,0 +1,71 @@
+// Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+describe('Dell Wsman Configure Idrac Job', function() {
+    var WsmanJob;
+    var uuid;
+    var job;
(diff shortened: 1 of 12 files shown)


spec/lib/jobs/dell-wsman-configure-idrac-spec.js
```diff
@@ -0,0 +1,71 @@
...
+        waterline.catalogs.findLatestCatalogOfSource.resolves({
+            "data": {
+                "DCIM_iDRACCardString": [
+                    {
+                        "FQDD": "testFQDD"
+                    }
+                ]
+            }
+        });
+        job._run().then(function() {
```
yyscamper: Promise need be returned, otherwise this test case will always success no matter the promise is fulfilled or rejected/

PR #568, opened 2017-12-05
RHEL 7.2 installation support

sushilrai:
-  Added support for additional packages
-  Cleanup of non-installation disks

Supporting PR for https://github.com/RackHD/on-taskgraph/pull/343

### DIFF
--- a/lib/task-data/schemas/install-centos.json
+++ b/lib/task-data/schemas/install-centos.json
@@ -50,6 +50,9 @@
         "networkDevices": {
             "$ref": "types-installos.json#/definitions/NetworkDeviceArray"
         },
+        "bonds": {
+            "$ref": "types-installos.json#/definitions/BondsArray"
+        },
         "kvm": {
             "$ref": "types-installos.json#/definitions/Kvm"
         },
@@ -61,6 +64,15 @@
         },
         "progressMilestones": {
             "$ref": "types-installos.json#/definitions/ProgressMilestones"
+        },
+        "packages": {
+            "$ref": "types-installos.json#/definitions/PackagesArray"
+        },
+        "disableServices": {
+            "$ref": "types-installos.json#/definitions/DisableServices"
+        },
+        "enableServices": {
+            "$ref": "types-installos.json#/definitions/EnableServices"
         }
     },
     "required": ["osType", "version", "repo", "profile", "installScript",
--- a/lib/task-data/schemas/types-installos.json
+++ b/lib/task-data/schemas/types-installos.json
@@ -107,9 +107,15 @@
                         "$ref": "#/definitions/VlanId"
                     },
                     "uniqueItems": true
+                },
+                "mtu": {
+                    "description": "The ipv4 MTU",
+                    "type": "integer",
+                    "minimum": 1500,
+                    "maximum": 9026
                 }
             },
-            "required": ["ipAddr", "netmask", "gateway"],
+            "required": ["ipAddr", "netmask"],
             "additionalProperties": false
         },
         "Ipv6Configuration": {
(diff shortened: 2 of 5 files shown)


anhou:
I didn't see 'clearDisk' related codes in https://github.com/RackHD/on-taskgraph/pull/343.

lib/task-data/schemas/types-installos.json
```diff
@@ -319,6 +319,22 @@
...
+            "items": {
+                "type": "string"
+            },
+            "uniqueItems": true
+        },
+        "ClearDisk": {
+            "description": "a flag to indicate whether partitions of non-installation should be cleared",
+            "type": "boolean"
+        },
+        "Kdump": {
```
anhou: add a field 'Kdump' is not a general way, if others want to add other items want to disable/enable by systemctl, more fields will be added, it's not so good.  so either controlling it by puppet/ansible/.. after os installation, or adding a more general way for things which could be controlled by 'systemctl'

lib/task-data/schemas/types-installos.json
```diff
@@ -319,6 +319,22 @@
...
+        },
+        "PackagesArray": {
+            "description": "specify additional packages to install",
+            "type": "array",
+            "items": {
+                "type": "string"
+            },
+            "uniqueItems": true
+        },
+        "ClearDisk": {
```
anhou: I didn't see 'clearDisk' related codes in https://github.com/RackHD/on-taskgraph/pull/343.

lib/task-data/schemas/types-installos.json
sushilrai: Thanks for the suggestion. Removed kdump from the PR. Will add support for providing the list of services that needs to be enabled / disabled during the post-installation.

lib/task-data/schemas/types-installos.json
sushilrai: Removed 'clearDisk' from the PR. Observed some failures after rebooting the server due to 'dd' of non-installation disks. Investigating other options. Will push a separate PR for cleanup of non-OS installation disks

lib/task-data/schemas/install-centos.json
```diff
@@ -50,6 +50,9 @@
         "networkDevices": {
             "$ref": "types-installos.json#/definitions/NetworkDeviceArray"
         },
+        "bonds": {
```
anhou: @sushilrai  I re-think about 'bonds' field, maybe 'bonds' is also one type of networkDevices interfaces, right?  so 'bond' could be at the same level with 'eth0' 'eth1'  under  networkDevices http://rackhd.readthedocs.io/en/latest/rackhd/install_os.html#networkdevices, right?
If I missunderstand something, please point out

sushilrai: Bonds networks are defined separately for simplicity. These logical interfaces requires different configuration parameters as compared to normal physical interfaces.
Also we need to configure the bond interfaces before the normal interfaces to avoid any conflict.

@anhou please confirm if we need to include the bond interfaces in `networkDevices` then I will make necessary changes

lib/task-data/schemas/types-installos.json
```diff
@@ -165,6 +171,38 @@
...
+        "BondConfig": {
+            "description": "Bond Interface configuration",
+            "type": "object",
+            "properties": {
+                "name": {
+                    "description": "The bond interface name.",
+                    "type": "string"
+                },
+                "nics": {
+                    "description": "the ipv4 configuration for this interface",
```
anhou: I guess it's copied from 'ipv4', it could be refined

lib/task-data/schemas/install-centos.json
anhou: @sushilrai  make sense for me that it's logical device that has different configurations. +1

anhou:
Could you also help update the document?  the docs repo is https://github.com/RackHD/docs. I think at least these two parts could be included. 1. include RHEL 7.2 support in the OS support table at http://rackhd.readthedocs.io/en/latest/rackhd/install_os.html#supported-os-installation-workflows, and update the payloads table at http://rackhd.readthedocs.io/en/latest/rackhd/install_os.html#non-windows-os-installation-workflow-payload, because some fields are only avalable at RHEL/CentOS, so please also help note (RHEL/CentOS only) in the docs. If you have any questions, let me know

sushilrai:
PR https://github.com/RackHD/docs/pull/419 is submitted with documentation change

anhou:
Thanks @sushilrai !
by the way, I left a  comment about 'nics' 's description last time. I believe it's the only left things. I'll merge this PR after it's fixed  :)

sushilrai:
@anhou I have added change for OS installation and have added relevant tests to cover new changes. As per test coverage, we are having lesser coverage for base installation. Please let me know if we are good to merge this PR or I need to add some tests to `lib/jobs/base-job.js`

anhou:
@sushilrai  we're looking at why concourse-ci fail, that's blocking the  merge.

PR #570, opened 2017-12-09
Adding new tasks and jobs for deploying hypervisor on target bare met…

rahmanmuhamad:
…al server via out of band controller using WSMAN protocol

### DIFF
--- a/lib/jobs/dell-wsman-os-create-job.js
+++ b/lib/jobs/dell-wsman-os-create-job.js
@@ -1,9 +1,7 @@
 //Copyright 2017, EMC, Inc.
 
 'use strict';
-
 var di = require('di');
-
 module.exports = wsmanOsDeploymnetCreateJobFactory;
 di.annotate(wsmanOsDeploymnetCreateJobFactory, new di.Provide('Job.Dell.Wsman.Os.Create'));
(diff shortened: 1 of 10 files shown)


lanchongyizu:
@rahmanmuhamad Could you please help to add schema for task, and unit test for job/task/base-task?

lib/jobs/dell-wsman-os-create-job.js
```diff
@@ -0,0 +1,127 @@
...
+
+
+     /**
+     *  Initialize basic configuration for the job
+     *
+     */
+    WsmanOsCreateJob.prototype._initJob = function () {
+        var self = this;
+        self.dellConfigs = configuration.get('dell');
+      if (!self.dellConfigs ||
```
nortonluo: Indent should be 4 space

nortonluo:
@rahmanmuhamad the unit test does not require real hardware and real OS deployment. Only function test need do that thing. In unit test, we often mock/stub those thing. For example:  create a sandbox and reply a smi-service http REST request with designated JSON payloads {status:success} instead of do the real thing. This trick cheats the program to make the code under test "feel as run in real word".

lanchongyizu:
depends on: https://github.com/RackHD/on-taskgraph/pull/345

PR #573, opened 2017-12-13
RAC-6553 Port add & delete volume redfish api

yaolingling:
**Background**
To port add & delete volume redfish api, the parameters of the two graphs need to be added. Three parameters are  ipAddress, username and password, which is used for obm setting. Details in https://rackhd.atlassian.net/browse/RAC-6553.

**Reviewers**
@nortonluo @lanchongyizu @AlaricChan

### DIFF
--- a/lib/jobs/dell-wsman-RAID.js
+++ b/lib/jobs/dell-wsman-RAID.js
@@ -56,10 +56,22 @@ function DellWsmanRAIDFactory(
 
     DellWsmanRAIDJob.prototype._handleSyncRequest = function() {
         var self = this;
-        return self.checkOBM('SCP RAID operation')
-        .then(function(obm){
+        if(self.options.ipAddress === '' || self.options.username === '' ||
+            self.options.password === ''){
+                return self.checkOBM('SCP RAID operation')
+                    .then(function(obm){
+                        return self.doRAIDoperation(obm);
+                    });
+        }else{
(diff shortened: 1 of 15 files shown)


lib/jobs/dell-wsman-RAID.js
```diff
@@ -56,10 +56,22 @@ function DellWsmanRAIDFactory(
 
     DellWsmanRAIDJob.prototype._handleSyncRequest = function() {
         var self = this;
-        return self.checkOBM('SCP RAID operation')
-        .then(function(obm){
+        if(self.options.ipAddress === '' || self.options.username === '' ||
+            self.options.password === ''){
```
pengz1: Your schema doesn't restrict type of those ipAddress/username/password to be string, you should consider cases that those parameters to be null and undefined as well. 
Expression like (Job.Dell.Wsman.RAID || self.options.username ||  self.options.password) should be better for you.

yaolingling: The type of ipAddress/username/password is string, which defined in schema.

pengz1: If you are mentioning dell-wsman-RAID.json, I didn't see that.

pengz1: OK, I see you update schema in this PR as well.

PR #574, opened 2017-12-14
[RAC-6550] Add unit tests for Job.Dell.Wsman.Config job

AlaricChan:
**background**
Add unit tests for WSMAN related jobs and graphs.
https://rackhd.atlassian.net/browse/RAC-6550

**Details**
Add unit tests for WSMAN config service job, related task and base task.

**Reviewer**
@yaolingling @nortonluo @lanchongyizu

### DIFF
--- a/lib/jobs/dell-wsman-config.js
+++ b/lib/jobs/dell-wsman-config.js
@@ -74,7 +74,7 @@ function DellWsmanConfigJobFactory(
         return Promise.resolve(result)
         .then(function() {
             if(!result || _.isEmpty(result)){
-                   throw new Error('Response for wsman microservice configuration is invalid.'); //jshint ignore:line
+                throw new Error('Response for wsman microservice configuration is invalid.'); //jshint ignore:line
             }
             configuration.set('dell:credentials', result.credentials);
             configuration.set('dell:gateway', result.gateway);
@@ -89,12 +89,15 @@ function DellWsmanConfigJobFactory(
             var buffer = {
                 "dell": configuration.get('dell')
             };
-            fs.writeFile(Constants.Configuration.Files.Dell, JSON.stringify(buffer, null, 4) + '\n', function(err){
-                if(err) {
-                    throw new Error('Could not write wsman microservice configs to file');
-                } else {
-                    logger.info('Wsman microservice config updated');
-                }
+            return new Promise(function(resolve, reject) {
+                fs.writeFile(Constants.Configuration.Files.Dell, JSON.stringify(buffer, null, 4) + '\n', function(err) {
+                    if(err) {
+                        reject('Could not write wsman microservice configs to file');
+                    } else {
+                        resolve();
+                        logger.info('Wsman microservice config updated');
+                    }
+                });
             });
         });
     };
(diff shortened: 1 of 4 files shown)


spec/lib/jobs/dell-wsman-config-spec.js
```diff
@@ -0,0 +1,124 @@
...
+                                name: 'updater',
+                                url: '/api/1.0/server/firmware/updater'
+                            }
+                        ]
+                    }
+                ]
+            };
+            fs.writeFile.yields(new Error('test'));
+            return expect(wsmanConfigJob._handleSyncResponse(response)).to.
+                be.rejectedWith('Could not write wsman microservice configs to file');
```
pengz1: I am not sure how this works, you put a callback into promise, suppose it is asynchronous.

AlaricChan: I guess yes. The "return" keyword is used to publish UT result asserted asynchronously, otherwise such case would always succeed.
http://chaijs.com/plugins/chai-as-promised/

pengz1: I got you, perhaps you should try stub.yieldsAsync().  stub.yields should be executed synchronously, that means you turned a callback that should be async into sync.

pengz1: And there is a obviously bug in the job that put async callback into a promise, it should be corrected.

spec/lib/jobs/dell-wsman-config-spec.js
```diff
@@ -0,0 +1,124 @@
...
+        });
+    });
+
+    describe('_handleSyncResponse', function() {
+        it('should reject if response is invalid', function() {
+            return expect(wsmanConfigJob._handleSyncResponse({})).to.
+                be.rejectedWith('Response for wsman microservice configuration is invalid.');
+        });
+
+        it('should update micro service config successfully', function() {
```
yaolingling: micro service? it should be microservice

spec/lib/jobs/dell-wsman-config-spec.js
```diff
@@ -0,0 +1,124 @@
...
+                        endpoint: [
+                            {
+                                name: 'updater',
+                                url: '/api/1.0/server/firmware/updater'
+                            }
+                        ]
+                    }
+                ]
+            };
+            fs.writeFile.yields();
```
yaolingling: sandbox.stub(fs, writeFile) maybe better

lib/jobs/dell-wsman-config.js
```diff
@@ -89,12 +89,15 @@ function DellWsmanConfigJobFactory(
             var buffer = {
                 "dell": configuration.get('dell')
             };
-            fs.writeFile(Constants.Configuration.Files.Dell, JSON.stringify(buffer, null, 4) + '\n', function(err){
-                if(err) {
-                    throw new Error('Could not write wsman microservice configs to file');
-                } else {
-                    logger.info('Wsman microservice config updated');
-                }
+            return new Promise(function(resolve, reject) {
```
pengz1: Looks good to me, just a tip: you can move line 92 to replace line 74 and 75

PR #580, opened 2018-01-24
Separate ssh-job into ssh-command and ssh-catalog

lanchongyizu:
**Background**

This is to fix "No parser" issue:
```
No parser exists for command sudo ./megaraid-config.sh
No parser exists for command sudo python set_interfaces.py 
No parser exists for command sudo ipmitool chassis bootdev pxe
```

https://rackhd.atlassian.net/browse/RAC-6579

**Done**

* Separate ssh-job into ssh-command and ssh-catalog

@RackHD/corecommitters @pengz1 @nortonluo


lib/task-data/tasks/ssh-exec.js
```diff
@@ -5,7 +5,7 @@
 module.exports = {
     friendlyName: 'Ssh Exec',
     injectableName: 'Task.Ssh.Exec',
-    implementsTask: 'Task.Base.Ssh',
+    implementsTask: 'Task.Base.Ssh.Command',
     options: {
         commands: null
```
nortonluo: If there is a schema in base task. There should be problem to put a null in default value here.

lib/utils/job-utils/command-util.js
```diff
@@ -33,34 +33,25 @@ function commandUtilFactory(
         this.nodeId = target;
     }
 
-    CommandUtil.prototype.parseResponse = function(tasks) {
+    CommandUtil.prototype.parseUnknownTasks = function(tasks) {
```
pengz1: Should the method name be something like "parseNonCatalogTasks"?

lib/utils/job-utils/command-util.js
```diff
@@ -33,34 +33,25 @@ function commandUtilFactory(
...
         });
 
-        var catalogTasks = [],
-            newTasks = [],
-            unknownTasks = [];
+         var unknownTasks = [];
 
          _.forEach(tasks, function(task) {
             logger.debug('task result', {data:task});
             if (task.catalog) {
```
pengz1: This should not be necessary under your case.

PR #583, opened 2018-02-09
RAC-6647 Fix for redfish ip range discovery

BrentHiggins:
When running discovery we found that the concurrency caused requests to fail 
This has been fixed

@jimturnquist


BrentHiggins:
Coverage decrease is due to removal of task failure for missing one catalog.
This makes it so the discovery will find everything it can regardless of errors

jimturnquist:
I believe the hound-ci bot is still reporting its last error incorrectly.  It looks like it was addressed but not updated.

PR #586, opened 2018-03-09
RAC-6706 Redfish discovery can not work with HTTPS

BrentHiggins:
Also lowers the time between requests to 3 seconds
@larry-dean @jimturnquist

### DIFF
--- a/lib/jobs/general-redfish-catalog.js
+++ b/lib/jobs/general-redfish-catalog.js
@@ -94,6 +94,7 @@ function GeneralRedfishCatalogJobFactory(
                     'user': settings.username || '',
                     'pass': settings.password || '',
                 },
+                strictSSL: false,
                 json: true
             });
         })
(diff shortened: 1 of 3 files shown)


lib/jobs/general-redfish-catalog.js
```diff
@@ -94,6 +94,7 @@ function GeneralRedfishCatalogJobFactory(
                     'user': settings.username || '',
                     'pass': settings.password || '',
                 },
+                rejectUnauthorized: false,
```
lanchongyizu: Could we use `strictSSL` instead of `rejectUnauthorized` since `rejectUnauthorized` is not documented in `request`?

lanchongyizu: Could we use `strictSSL` instead of `rejectUnauthorized` since `rejectUnauthorized` is not documented in `request`?

BrentHiggins:
Thank you for the comment I had missed the obvious tag when looking for the solution fixed now

PR #587, opened 2018-03-12
[RAC-6682] Add lookup update in post WSMAN device discovery

nortonluo:
## Background
The node discovered by WSMAN discovery workflow have no mac lookup entry associated with it. This is because the traditional micro kernel discovery way have a step to update lookup table but WSMAN do not.

## Solution 
Add a task to use the MAC address in node catalog to update lookup table. This task will be execute in the post WSMAN discovery workflow after get WSMAN inventory.

@lanchongyizu @pengz1 @iceiilin @markbeierl

### DIFF
--- a/lib/task-data/base-tasks/wsman-update-lookups-base.js
+++ b/lib/task-data/base-tasks/wsman-update-lookups-base.js
@@ -1,0 +1,12 @@
+// Copyright © 2018 Dell Inc. or its subsidiaries. All Rights Reserved. 
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Snmp Update Lookups',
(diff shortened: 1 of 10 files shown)


lib/task-data/base-tasks/wsman-update-lookups-base.js
```diff
@@ -0,0 +1,12 @@
+// Copyright © 2018 Dell Inc. or its subsidiaries. All Rights Reserved. 
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Snmp Update Lookups',
```
lanchongyizu: Snpm?

keedya:
👍

PR #592, opened 2018-04-02
[RAC-6712] Add failure handling for OS installation

iceiilin:
OS installation job uses powershell script to prepare for Windows Server installation. The script doesn't throw a error but notify RackHD by a POST command about its completion, which results in the success of the job while it actually fails.

This PR signals the job as fail when receiving notification with error messages. If the notification doesn't contain the key word "status", the job views it as success, which is compatible with previous OS tasks.

The modification of powershell script is in https://github.com/RackHD/on-taskgraph/pull/374

### DIFF
--- a/spec/lib/jobs/install-os-spec.js
+++ b/spec/lib/jobs/install-os-spec.js
@@ -387,7 +387,7 @@ describe('Install OS Job', function () {
         });
     });
 
-    it('should finish job if task notification received', function() {
+    it('should finish job if task notification received without error', function() {
         subscribeNodeNotification.restore();
         subscribeNodeNotification = sinon.stub(
             InstallOsJob.prototype, '_subscribeNodeNotification', function(_nodeId, callback) {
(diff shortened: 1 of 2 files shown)


spec/lib/jobs/install-os-spec.js
```diff
@@ -406,6 +406,61 @@ describe('Install OS Job', function () {
...
+                    nodeId: _nodeId,
+                    nodeIp: '1.1.1.1',
+                    status: 'fail'
+                });
+            });
+
+        waterline.graphobjects.findOne = sinon.stub().resolves(graph);
+        job.run()
+        .then(function(){
+            done(new Error("Expected job to fail"));
```
lanchongyizu: The error should have been rejected by `_subscribeNodeNotification`, so `"Expected job to fail"` error makes no sense though this kind of code exists everywhere in our repo.

lanchongyizu:
depends on: https://github.com/RackHD/on-taskgraph/pull/374

PR #594, opened 2018-04-30
Add underscore module; used but not defined

keedya:


keedya:
@geoff-reid @uppalk1 @dalebremner @king-jam

markbeierl:
I have tested this and can confirm it is now working.

